PR #102, opened 2025-09-08
fix: case-sensitive import path issues by renaming files

frhanjav:
### What this PR does?
This PR fixes incorrect import paths where filenames had mismatched casing. This resolves build/runtime issues on case-sensitive file systems (Linux).

### Before:
Import statements were failing due to case mismatches:

In files like `agent/src/lib/browser/BrowserPage.ts`:
```
import { profileAsync, profileSync } from '../utils/profiler';
```

gave this error: `Module not found: Error: Can't resolve '../utils/profiler'` as we had Profiler.ts and not profiler.ts

Similarly:
```
import xyz from '@/lib/tools/Tool.interface'
```

gave this error: `Cannot find module '@/lib/tools/Tool.interface' or its corresponding type declarations` as we had tool.interface.ts and not Tool.interface.ts

### After:
Renamed files to match existing import paths:
- Profiler.ts → profiler.ts 
- tool.interface.ts → Tool.interface.ts

Builds successfully on all file systems.


frhanjav:

PR #263, opened 2025-12-19
feat: windows icon fix

shadowfax92:
- **fix: windows dll to chrome.dll and chrome.exe**
- **fix: clean-up old windows icon fix**


shadowfax92:
image

verified it works

PR #448, opened 2026-03-15
fix: filter empty messages from conversation history

shivammittal274:
## Summary
- Adds `message-validation.ts` utility with `hasMessageContent()` / `filterValidMessages()` to detect empty UIMessages that would fail SDK or provider validation
- Filters messages before passing to `createAgentUIStreamResponse` and when persisting in `onFinish`
- Skips empty `content` entries during `previousConversation` injection

## Problem
Two failure modes caused crashes on multi-turn conversations:
1. **AI SDK validation**: Assistant messages with `parts: []` (e.g. from aborted streams) fail `.nonempty()` check — "Message must contain at least one part"
2. **Provider API**: Assistant messages with `parts: [{ type: 'text', text: '' }]` pass SDK validation but are rejected by providers like Gemini — "must not be empty"

## Test plan
- [x] Tested new conversation with empty assistant `content: ""` in `previousConversation` — previously crashed, now skips empty message and responds correctly
- [x] Tested valid `previousConversation` — history recalled correctly
- [x] Tested follow-up on same session — stored messages reused without errors
- [x] Typecheck passes

### DIFF
--- a/packages/browseros-agent/apps/server/src/agent/message-validation.ts
+++ b/packages/browseros-agent/apps/server/src/agent/message-validation.ts
@@ -1,0 +1,46 @@
+/**
+ * @license
+ * Copyright 2025 BrowserOS
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+import type { UIMessage } from 'ai'
+
+/**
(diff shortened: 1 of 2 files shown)


shivammittal274:

PR #550, opened 2026-03-24
feat: rebuild bdev as an XDG-backed Cobra CLI

shadowfax92:
## Summary
- rewrite `tools/bdev/` as a Go Cobra CLI for BrowserOS Chromium patch workflows
- add XDG-backed checkout/session state for multi-checkout apply, sync, rebase, export, and conflict repair flows
- restore standard CLI repo ergonomics in `tools/bdev` with `make build`, `make test`, `make install`, binary ignores, and local install docs

## Rework
After feedback, I kept the new bdev engine and reworked the packaging layer to match the existing CLI repos under `/Users/shadowfax/code/clis`. The follow-up restores the expected `make install` developer flow instead of leaving `tools/bdev` as a one-off package with no install target.

## Test plan
- `cd tools/bdev && make build`
- `cd tools/bdev && make test`
- `cd tools/bdev && make install`
- `cd tools/bdev && go run . --help`


shadowfax92:
Addressed the concrete review items in 831aefac:\n- fixed grouped help rendering\n- fixed single-pass reject apply behavior\n- skipped overlay replay when upstream sync for the same path fails\n- propagated reset head-rev errors\n- moved export/continue activity timing to wrap real work\n- added a help-output regression test\n\nI did not apply the suggestion to add successful overlay paths to \ because overlay paths are already a subset of successful upstream paths in this implementation, so that change would double-count them.

PR #586, opened 2026-03-26
feat: move execution history into app shell

shadowfax92:
## Summary
- rename the app-shell trust surface to `Agent Observability` and move it to the nav slot directly above `Settings`
- simplify the page into a calmer recent-runs feed by removing stat cards, search, and filter tabs
- add task-level local deletion and keep `/executions` as a redirect to `/observability`

## Rework
This follow-up addresses feedback that the first app-shell page still felt too dashboard-like. The surface now reads as a lighter observability feed instead of a metrics console while preserving the existing execution capture model.

## Test plan
- bun run --filter @browseros/agent typecheck
- bun run --filter @browseros/agent lint
- bun test apps/agent/lib/execution-history/normalize.test.ts
- bun run test


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #611, opened 2026-03-28
fix: rebuild chat sessions when provider config changes

shadowfax92:
## Summary
- rebuild live chat sessions when the material agent config changes so provider/model/auth updates take effect in the current thread
- prevent exhausted BrowserOS-backed sessions from continuing to serve requests after the user switches to another provider
- preserve sanitized conversation history across rebuilds and cover the invalidation path with focused server tests

## Design
The server now fingerprints the resolved agent configuration for each conversation and stores that fingerprint alongside the in-memory session metadata. On each `/chat` request, `ChatService` compares the current resolved config to the stored fingerprint and rebuilds the `AiSdkAgent` when the provider configuration changed, reusing the existing sanitized message history so the thread stays usable.

## Test plan
- `bun test apps/server/tests/api/services/chat-service.test.ts`
- `bunx biome check apps/server/src/api/services/chat-service.ts apps/server/src/agent/session-store.ts apps/server/tests/api/services/chat-service.test.ts .llm/0327-fix_provider_switch_credit_session/*.md`
- `bunx tsc -p apps/server/tsconfig.json --noEmit`
- `bun run --filter @browseros/server test:integration`

### DIFF
--- a/packages/browseros-agent/apps/server/src/api/services/chat-service.ts
+++ b/packages/browseros-agent/apps/server/src/api/services/chat-service.ts
@@ -308,42 +308,46 @@ export class ChatService {
     const contextChanges: string[] = []
     let currentSession = session
 
-    if (currentSession && currentSession.mcpServerKey !== mcpServerKey) {
+    const hasMcpChange =
+      currentSession !== undefined &&
+      currentSession.mcpServerKey !== mcpServerKey
+    const hasWorkingDirChange =
+      currentSession !== undefined &&
+      currentSession.workingDir !== request.userWorkingDir
+    const hasAgentConfigChange =
+      currentSession !== undefined &&
+      currentSession.agentConfigKey !== undefined &&
+      currentSession.agentConfigKey !== agentConfigKey
+
+    if (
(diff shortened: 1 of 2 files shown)


shadowfax92:
Addressed the review follow-up in `79bb8888`.

Changes made:
- added `evalMode` to the agent config fingerprint
- skipped agent-config rebuilds for legacy in-memory sessions that do not have a stored `agentConfigKey`
- changed `applySessionChanges` to compute change flags first and rebuild at most once per request
- expanded the focused server tests to cover `evalMode` fingerprinting and the legacy-session guard

I intentionally kept the other creation-time fields in `buildAgentConfigKey` (`chatMode`, `userSystemPrompt`, `declinedApps`, `origin`, `contextWindowSize`, `supportsImages`, reasoning settings, etc.). In this codebase those values are baked into `AiSdkAgent.create()` via toolset selection, prompt construction, middleware setup, and provider options, so excluding them would leave stale-session behavior in place for mid-conversation changes outside raw provider credentials.

PR #614, opened 2026-03-28
fix: 👍 Top features/issues by Votes (updated automatically every hour)

abhiavi:
Fix for 👍 Top features/issues by Votes (updated automatically every hour). Verified with local tests.


shadowfax92:
Sorry @abhiavi the PR title and fix is completely different! can you please elaborate what is being fixed here. Thanks!

PR #629, opened 2026-04-01
fix: isolate server release packaging to ci mode

shadowfax92:
## Summary
- restore the stricter production inline-env validation for normal server builds
- add an explicit `--ci` server build mode that packages local release zips without R2 or production secret requirements
- point `build:server:ci` at `--ci` so the release workflow keeps producing GitHub release assets without changing the shared production build contract

## Why
The previous follow-up fixed the release workflow by weakening shared production validation. That was the wrong boundary: normal production builds should still fail if required inline env values are missing. This rework keeps the release exception explicit and isolated to the CI packaging path.

## Test plan
- `bunx biome check packages/browseros-agent/scripts/build/server/cli.ts packages/browseros-agent/scripts/build/server/config.ts packages/browseros-agent/scripts/build/server/orchestrator.ts packages/browseros-agent/scripts/build/server/types.ts packages/browseros-agent/package.json packages/browseros-agent/apps/server/src/env.ts packages/browseros-agent/apps/server/tests/build.test.ts .github/workflows/release-server.yml`
- `bun test packages/browseros-agent/apps/server/tests/build.test.ts`
- `cd packages/browseros-agent && cp apps/server/.env.production.example apps/server/.env.production && bun run build:server:ci`

### DIFF
--- a/packages/browseros-agent/apps/server/src/env.ts
+++ b/packages/browseros-agent/apps/server/src/env.ts
@@ -23,5 +23,8 @@ export const INLINED_ENV = {
 } as const
 
 export const REQUIRED_FOR_PRODUCTION = [
+  'SENTRY_DSN',
+  'CODEGEN_SERVICE_URL',
+  'POSTHOG_API_KEY',
   'BROWSEROS_CONFIG_URL',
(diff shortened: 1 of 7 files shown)


shadowfax92:
@claude review

PR #645, opened 2026-04-03
fix: skip windows exe patching in ci mode to avoid wine dependency

shadowfax92:
## Summary

- The server release CI workflow (`release-server.yml`) fails on `ubuntu-latest` because `patch-windows-exe.ts` requires Wine to run `rcedit-x64.exe`, and Wine is not available on the CI runner
- Threads the existing `--ci` flag through `compileServerBinaries` → `compileTarget` so Windows PE metadata patching is skipped in CI mode with a warning log
- Consistent with the existing CI isolation pattern where `--ci` already skips R2 uploads and production secrets

## Changes

- `scripts/build/server/compile.ts` — Added `ci` parameter to `compileTarget` and optional `options` to `compileServerBinaries`. When `ci=true` and target is Windows, logs a warning and skips patching instead of failing.
- `scripts/build/server/orchestrator.ts` — Passes `{ ci: args.ci }` to `compileServerBinaries`.

## Test plan

- [ ] Re-run the `Release BrowserOS Server` workflow on main after merging — should pass the "Build release artifacts" step
- [ ] Verify local `bun run build:server:test` still works (non-CI mode, no behavior change)
- [ ] Verify CI log shows warning: "Skipping Windows exe metadata patching in CI mode"

### DIFF
--- a/packages/browseros-agent/scripts/build/server/compile.ts
+++ b/packages/browseros-agent/scripts/build/server/compile.ts
@@ -1,6 +1,7 @@
 import { mkdirSync, rmSync } from 'node:fs'
 import { join } from 'node:path'
 
+import { log } from '../log'
 import { wasmBinaryPlugin } from '../plugins/wasm-binary'
 import { runCommand } from './command'
 import type { BuildTarget, CompiledServerBinary } from './types'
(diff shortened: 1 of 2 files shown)


shadowfax92:
@claude review

Issue #641, opened 2026-04-03
Theme setting does not persist after closing and re-opening the browser


DenysMb:
### Issue Type

Browser Issue

### Operating System

Linux

### Description of the bug

Hey!

One thing I noticed some time ago is that the theme setting does not persist after closing and re-opening the browser.

### Steps to Reproduce

1. Under Appearance > Theme change it to "Use GTK".
2. Close the browser and re-open

### Screenshots / Videos

⚠️ **The glitches are from the recording, not a problem with the browser.**

What matters here is that when we change to GTK, close the browser and open again, the setting is reverted.

https://github.com/user-attachments/assets/a02607a6-d921-4fcc-bc57-28b4b439160f

### BrowserOS Version

0.44.0.1

shadowfax92:
Hey @DenysMb — thanks for the report and the video! Logged as a bug. We'll look into this soon.

shadowfax92:
Thanks @DenysMb — theme not persisting across restart is a legit persistence bug (same family as other "setting doesn't stick" reports). Adding to my list and assigning to me.

shadowfax92:
This is fixed in [BrowserOS 0.50.3](https://github.com/browseros-ai/BrowserOS/releases/tag/v0.50.3). Please update to 0.50.3; your theme selection should persist after restarting the browser. Thanks for the report and recording — closing as fixed.

DenysMb:
I am still able to reproduce this bug with the latest version. 😕

https://github.com/user-attachments/assets/54edd98a-a2b3-415f-aeaf-47b45dae3670

PR #647, opened 2026-04-04
fix(browser): persist theme setting by checking system theme before reset

vaidik-bajpai:
This PR fixes this [issue](https://github.com/browseros-ai/BrowserOS/issues/641) where the user's theme selection (specifically "Use GTK" on Linux) would not persist after restarting the browser.

The Problem: The SyncDefaultTheme function was executing on every startup and overwriting the theme with the "BrowserOS Blue" tonal spot if it found the kUserColor preference at its default value. When using GTK, kUserColor is indeed default, causing the reset.

The Fix: Added a guard clause to SyncDefaultTheme in browseros_prefs.cc that check if ::prefs::kUsesSystemTheme is enabled. If true, the function returns early, preserving the user's system theme selection.


shadowfax92:

PR #661, opened 2026-04-08
feat: show custom model ID as first option in model selector

shivammittal274:
## Summary
- When typing in the model dropdown, the user's exact input now appears as the first selectable row, followed by fuzzy search suggestions
- Pressing Enter always commits the typed text as a custom model ID
- Context window badge is hidden for custom entries (since it's unknown)
- Analytics correctly tracks `is_custom_model` by checking against the known model list

## Context
Previously, entering a custom model ID required typing something that returned zero fuzzy matches, then pressing Enter. Since fuzzy search (Fuse.js threshold 0.4) almost always returns results, users could never discover the custom model path.

## Test plan
- [ ] Open Settings → Add Provider (e.g. OpenRouter)
- [ ] Type a custom model ID (e.g. `my-org/my-model`) → verify it appears as the first row without a context window badge
- [ ] Click it or press Enter → verify it's selected as the model
- [ ] Type a known model prefix (e.g. `anthropic/`) → verify the exact text appears first, followed by matching suggestions with context window badges
- [ ] Type an exact model ID (e.g. `anthropic/claude-sonnet-4.5`) → verify no duplicate first row appears
- [ ] Open dropdown without typing → verify full model list shows as before

### DIFF
--- a/packages/browseros-agent/apps/agent/entrypoints/app/ai-settings/NewProviderDialog.tsx
+++ b/packages/browseros-agent/apps/agent/entrypoints/app/ai-settings/NewProviderDialog.tsx
@@ -305,9 +305,14 @@ export const NewProviderDialog: FC<NewProviderDialogProps> = ({
     [modelInfoList],
   )
 
-  const filteredModels = modelSearch
-    ? modelFuse.search(modelSearch).map((r) => r.item)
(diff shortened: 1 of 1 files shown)


shivammittal274:
@greptile-ai review

PR #665, opened 2026-04-09
Feat: Improved ACL robustness with semantic and fuzzy matching

neel04:
## Summary

  This adds an enhanced, python-based ACL with fuzzy and semantic matching. It's gated behind `BROWSEROS_ACL_PYTHON` for the time being,

We compute a weighted sum with the individual scores of exact, fuzzy and semantic matches to output an overall "confidence" in blocking some request. However, the coefficients need to be tuned further based on user feedback.

  ## Run

  From `packages/browseros-agent/python/acl_lab`:

  ```py
  uv sync
  uv run python -m acl_lab score fixtures/submit_button.json --pretty
  uv run pytest -s tests
```

  From packages/browseros-agent:

```c
  BROWSEROS_ACL_PYTHON=1 bun run dev:watch
```

  ## Notes

  - If `BROWSEROS_ACL_PYTHON` is unset, the server uses the existing TypeScript matcher.
  - If the Python worker is enabled but fails, the current behavior is fail-open.

## Future

Integrating a UI wherein if ACL blocks an action, it should just come up in the UI and the user should be able to override that. 

This feedback should be reported back to us, which we can use to setup a pipeline and automatically adjust coefficients to  balance False positive/True negatives


neel04:

PR #672, opened 2026-04-10
perf(agent): Optimize file reading in memory search using Promise.all

vincerevu:
**What:**
Optimized `loadMemoryEntries` in `packages/browseros-agent/apps/server/src/tools/memory/search.ts` to read multiple `.md` memory files concurrently rather than sequentially.

**Why:**
The previous implementation sequentially awaited `readFile` inside a `for...of` loop. This caused an N+1 Query-style issue, particularly noticeable as the number of memory files grew. By parsing files concurrently, we unblock the event loop and speed up I/O execution.

**Measured Improvement:**
I created a dummy benchmark script generating 100 small markdown files and running the sequential vs parallel methods 100 times.
- Baseline (Sequential): ~1340ms
- Optimized (Parallel): ~136ms
- Net improvement: ~10x speedup for this specific simulated workload.

### DIFF
--- a/benchmark.ts
+++ b/benchmark.ts
@@ -1,0 +1,116 @@
+import { readdir, readFile } from 'node:fs/promises'
+import { join } from 'node:path'
+
+interface MemoryEntry {
+  source: string
+  content: string
+}
+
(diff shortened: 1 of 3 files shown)


vincerevu:

PR #673, opened 2026-04-10
perf: optimize I/O in generateCatalog with Promise.all

vincerevu:
**What:** Replaced the sequential `for...of` loop in `generateCatalog` in `upload-skills-catalog.ts` with a `.map` mapping each directory entry to a Promise that stats the entry and reads `SKILL.md`, and then resolved them all concurrently using `Promise.all`.

**Why:** The previous implementation awaited `stat` and `readFile` sequentially for each entry, leading to an N+1 problem and slower performance as the number of skill directories grows. The optimization allows the file system to parallelize reads.

**Measured Improvement:** We established a performance benchmark measuring 100 iterations of reading the skills defaults directory. 
- Baseline: ~303-443ms (depending on machine load)
- Optimized: ~104-138ms 
- Result: roughly a 3.0x - 3.2x performance improvement.


vincerevu:

PR #674, opened 2026-04-10
perf: Optimize modifier key release and press loop performance

vincerevu:
**What:** The optimization uses `Promise.all` to concurrently dispatch `keyDown` and `keyUp` events for modifier keys.
**Why:** Previously, the `for...of` loops sequentialized these network requests via `await`, leading to an N+1 query issue for key modifier release/press sequences.
**Measured Improvement:** In a mocked benchmark for 4 modifier keys over 10 iterations, the average time dropped from ~26.9ms to ~5.6ms (~5x speedup), proving that concurrent CDP request dispatch vastly outperforms sequential await chaining without breaking CDP ordering requirements.


vincerevu:

PR #675, opened 2026-04-10
perf(eval): optimize screenshot loading in fara multimodal grader

vincerevu:
**What:**
Replaced the sequential `for...of` loop inside `FaraMultimodalGrader.loadScreenshots` with `Promise.all` and `.map`. This processes file reads in parallel and correctly retains error-skipping by catching rejections inside the map and returning `null`, then safely filtering them out. Because `Promise.all` returns elements in order, it natively preserves identical ordering logic as the initial sequence.

 **Why:**
The previous code awaited file reads (`readFile`) and base64 conversions sequentially within a loop. For agents outputting a high number of screenshots, this linear wait causes significant delays before evaluation even begins. Parallelizing I/O significantly optimizes evaluation processing time.

**Measured Improvement:**
Created a localized mock benchmark to measure reading 500 screenshots of 1MB each:
* **Baseline**: 7583.24ms
* **Optimized**: 5338.94ms
* **Change**: ~2244.30ms speed up (approx 30% performance improvement on local disk I/O test).

### DIFF
--- a/packages/browseros-agent/apps/eval/src/graders/fara/multimodal.ts
+++ b/packages/browseros-agent/apps/eval/src/graders/fara/multimodal.ts
@@ -191,17 +191,26 @@ export class FaraMultimodalGrader implements Grader {
       indices.sort((a, b) => a - b)
     }
 
-    for (const i of indices) {
-      try {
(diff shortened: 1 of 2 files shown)


vincerevu:

PR #676, opened 2026-04-10
perf: Optimize N+1 query issue in inspect-ui.ts modifier keys loop

vincerevu:
**What:** The sequential CDP event dispatches in `cmdPressKey` within `inspect-ui.ts` for modifier keys (both keyDown and keyUp) have been replaced with `Promise.all()` to execute concurrently.

**Why:** The N+1 query issue caused a performance bottleneck when simulating key events that included multiple modifier keys (e.g. `Control+Shift+Alt+A`). Sequentially awaiting the dispatch of each `Input.dispatchKeyEvent` through CDP introduces latency equal to the round-trip time per key. By utilizing `Promise.all()`, the CDP commands are executed concurrently, cutting down the total execution time dramatically.

**Measured Improvement:** 
Baseline (Mock CDP 10ms Network Delay, `Control+Shift+Alt+A`): ~66.6ms
Improved (Mock CDP 10ms Network Delay, `Control+Shift+Alt+A`): ~24.3ms
Improvement: ~2.7x speedup for combos with 3 modifiers.

### DIFF
--- a/packages/browseros-agent/bun.lock
+++ b/packages/browseros-agent/bun.lock
@@ -1,6 +1,6 @@
 {
   "lockfileVersion": 1,
-  "configVersion": 1,
+  "configVersion": 0,
   "workspaces": {
     "": {
       "name": "browseros-monorepo",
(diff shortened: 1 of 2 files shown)


vincerevu:

PR #677, opened 2026-04-10
perf: apply exponential backoff to polling loops

vincerevu:
**What:** The optimization replaced a fixed 500ms delay in several polling loops (`waitForServerHealth`, `waitForBrowserReady`, etc.) with an exponential backoff pattern. The delay now starts at 100ms and scales by 1.5x on each attempt, maxing out at 2000ms.
**Why:** The test lifecycle and evaluation scripts were suffering from excessive wait times on happy paths (as they couldn't detect a ready server earlier than 500ms bounds) and unnecessary requests on unhappy paths. This is effectively an N+1 query polling issue.
 **Measured Improvement:** In a simulated test mimicking `waitForServerHealth` polling that resolves immediately after 100ms:
* **Baseline**: Took ~500ms (1 successful polling attempt delayed by fixed 500ms).
* **Improvement**: Took ~128ms with exponential backoff.
* **Change**: Nearly an ~80% reduction in delay for fast-to-ready services, minimizing the wait time impact without bombarding the target server over an extended duration.

### DIFF
--- a/packages/browseros-agent/apps/eval/scripts/test-failure-scenarios.ts
+++ b/packages/browseros-agent/apps/eval/scripts/test-failure-scenarios.ts
@@ -107,6 +107,7 @@ async function waitForServerHealth(
   port: number,
   maxAttempts = 60,
 ): Promise<boolean> {
+  let delay = 100
   for (let i = 0; i < maxAttempts; i++) {
     try {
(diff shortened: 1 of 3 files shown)


vincerevu:

PR #708, opened 2026-04-15
refactor: remove redundant context-overflow middleware

shadowfax92:
## Summary

- Removes `context-overflow-middleware.ts` and its wiring in `ai-sdk-agent.ts`. The middleware caught provider overflow errors and retried with a naive `nonSystem.slice()` truncation that has no awareness of tool_use/tool_result pairing — a cut between an assistant `tool_use` and the matching `tool_result` produces an orphaned `tool_use` that Anthropic/Bedrock/Google reject with a different error, turning one failure into a harder-to-debug one.
- Compaction (`createCompactionPrepareStep`) already handles overflow safely: `findSafeSplitPoint` walks past `tool` messages to preserve pair integrity, and the proactive pipeline (strip binary → prune old tool calls → reduce tool outputs → LLM summarize → sliding-window fallback) catches overflow before the request ever leaves the agent.
- Also drops the matching `describe('createContextOverflowMiddleware')` block (8 tests) and 11 helpers it relied on in `compaction.test.ts`. Net: −426 lines.

## Design

Compaction runs in `prepareStep` *before* the request — it estimates token usage and rewrites the message list to fit. The middleware ran *after* the request failed at the provider, with a much cruder truncation strategy. Two systems for the same problem, where the second one is structurally less safe than the first. Removing the middleware leaves a single, well-tested code path. The `isV3Model` check is preserved (now guards only the dev-only `devToolsMiddleware` wrap).

## Test plan

- [x] `bun run typecheck` — passes across all packages
- [x] `bunx biome check` on changed files — clean
- [x] `bun test apps/server/tests/agent/compaction.test.ts` — 88 pass, 0 fail
- [ ] Smoke-test a long agent session with each provider (Anthropic, OpenAI, Google) to confirm compaction triggers cleanly without the safety net


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #723, opened 2026-04-16
fix: add null guard in getYahooIndiaSuggestions to prevent TypeError

### LINKED ISSUE #722: Missing null guard in getYahooIndiaSuggestions causes uncaught TypeError on bad API response
## Problem

In `packages/browseros-agent/apps/agent/entrypoints/newtab/index/lib/searchSuggestions/getSearchSuggestions.ts`, the `getYahooIndiaSuggestions` function accesses `data.gossip.results` without any null/undefined guard:

```ts
const getYahooIndiaSuggestions = async (query: string): Promise<string[]> => {
  const response = await fetch(
    `https://in.search.yahoo.com/sugg/gossip/gossip-in-loc/?command=${encodeURIComponent(query)}&output=json`,
  )
  const data = await response.json()
  return data.gossip.results.map((item: YahooSuggestionItem) => item.key) || []
}
```

If the Yahoo API returns an unexpected response shape (e.g. `data.gossip` is `null` or `undefined` — which can happen on network errors, API changes, or rate limiting), this will throw:
```
TypeError: Cannot read properties of null/undefined (reading 'results')
```

Note that every other search provider (`getGoogleSuggestions`, `getBingSuggestions`, `getDuckDuckGoSuggestions`, `getBraveSuggestions`) safely uses `data[1] || []`, but Yahoo lacks this protection.

Additionally, none of the fetch calls check `response.ok`, so HTTP errors (4xx/5xx) silently pass through to JSON parsing, which can throw a `SyntaxError` on non-JSON error bodies.

## Steps to Reproduce

1. Set the search engine to Yahoo in BrowserOS.
2. If the Yahoo suggestions API is unreachable or returns an error/unexpected payload, the `getYahooIndiaSuggestions` call will throw an uncaught `TypeError`.
3. This bubbles up and can break the search suggestions UI entirely.

## Expected

`getYahooIndiaSuggestions` should safely return `[]` (empty array) when the API response is missing or malformed, consistent with all other provider implementations.

## Fix

Add a null guard and optional chaining:
```ts
return data?.gossip?.results?.map((item: YahooSuggestionItem) => item.key) ?? []
```

Also check `response.ok` before parsing JSON for all providers.

kuishou68:
Closes #722

## Problem

In `getSearchSuggestions.ts`, the `getYahooIndiaSuggestions` function accesses `data.gossip.results` without any null/undefined guard:

```ts
return data.gossip.results.map((item: YahooSuggestionItem) => item.key) || []
```

If the Yahoo API returns an unexpected response shape (e.g. `data.gossip` is `null` or `undefined`), this throws an uncaught `TypeError` that can break the search suggestions UI entirely. Every other provider safely uses `data[1] || []`, but Yahoo lacks this protection.

## Fix

Use optional chaining and nullish coalescing, consistent with how other providers handle missing data:

```ts
return data?.gossip?.results?.map((item: YahooSuggestionItem) => item.key) ?? []
```

This ensures an empty array is safely returned when the API response is missing or malformed.

### DIFF
--- a/packages/browseros-agent/apps/agent/entrypoints/newtab/index/lib/searchSuggestions/getSearchSuggestions.ts
+++ b/packages/browseros-agent/apps/agent/entrypoints/newtab/index/lib/searchSuggestions/getSearchSuggestions.ts
@@ -25,7 +25,7 @@ const getYahooIndiaSuggestions = async (query: string): Promise<string[]> => {
     `https://in.search.yahoo.com/sugg/gossip/gossip-in-loc/?command=${encodeURIComponent(query)}&output=json`,
   )
   const data = await response.json()
-  return data.gossip.results.map((item: YahooSuggestionItem) => item.key) || []
+  return data?.gossip?.results?.map((item: YahooSuggestionItem) => item.key) ?? []
 }
 
 const getDuckDuckGoSuggestions = async (query: string): Promise<string[]> => {


kuishou68:

shadowfax92:

PR #726, opened 2026-04-16
feat(ota): bundle full server resources tree in Sparkle payload

shadowfax92:
## Summary

- OTA Sparkle zip now ships the full `resources/` tree the agent build produced (browseros_server + bun + rg + podman + gvproxy + vfkit + krunkit + podman-mac-helper + win-sshproxy), not just `browseros_server`. After an OTA update, `versions/<X>/resources/bin/third_party/` exists and matches a fresh Chromium-build install.
- Every binary inside the Sparkle zip is codesigned with the identifier + options + entitlements from a single shared table (`build/common/server_binaries.py`) that the Chromium-build sign path and the OTA sign path both consume — so adding a new third-party dep is a one-file edit that flows to both paths automatically.
- Unknown executables under `resources/bin/` are a hard error at release time, preventing silent unsigned shipment when the shared table hasn't been updated.

## Background

PR #719 added Podman to the Chromium-build path but did not touch the OTA pipeline at `packages/browseros/build/modules/ota/`. That pipeline was built around a single-binary abstraction: `find_server_binary()` returned one path, `create_server_zip()` staged exactly that one file, `sign_binary.py` signed one path at a time. Every third-party binary in the upstream artifact zip was silently discarded when the OTA payload was built.

Empirically verified: after OTA to v0.0.86, `~/Library/Application Support/BrowserOS/.browseros/versions/0.0.86/resources/bin/` contained only `browseros_server` — no `third_party/` dir. Podman integration was non-functional for any user updating via OTA, which is the primary distribution channel for existing installs. No appcast schema change, no client-side updater change, no agent-build change needed — just a refactor of the OTA bundling step.

## Design

- `build/common/server_binaries.py` (new) holds `MACOS_SERVER_BINARIES: Dict[str, SignSpec]` keyed by binary stem and `WINDOWS_SERVER_BINARIES: List[str]` of relative paths. `modules/sign/macos.py` and `modules/sign/windows.py` now import from here; OTA imports the same table.
- `modules/ota/common.py`: `create_server_zip(binary_path, ...)` → `create_server_bundle_zip(resources_dir, output_zip)` that zips the whole extracted tree; `find_server_binary` → `find_server_resources_dir` that returns the extracted `resources/` directory.
- `modules/ota/sign_binary.py`: adds `sign_server_bundle_macos` (walks `resources/bin/**`, fails fast on unknowns, codesigns each file using the shared table's identifier + options + entitlements) and `sign_server_bundle_windows` (signs the explicit Windows relative-path list). Also adds `notarize_macos_zip` so the outer Sparkle zip is submitted to `notarytool` directly — the existing `notarize_macos_binary` wraps its input via `ditto --keepParent`, which would double-zip the already-built bundle and prevent Apple's service from reaching the inner Mach-O binaries.
- `modules/ota/server.py::ServerOTAModule.execute()`: per-platform flow becomes copy-resources → codesign bundle → zip → notarize zip → Sparkle-sign. Sparkle zip layout is identical to what the Chromium build already bakes into `.app`, so BrowserOS's in-process updater needs zero changes.

## Test plan

- [ ] `cd packages/browseros && python -m build.common.server_binaries_test` (7 tests, validates table shape + entitlements plists exist)
- [ ] `cd packages/browseros && python -m build.modules.ota.bundle_test` (5 tests, validates bundle-zip layout + exec-bit preservation + missing-dir failure)
- [ ] `cd packages/browseros && python -m build.modules.storage.download_test` and `upload_test` — pre-existing tests still pass
- [ ] On macOS release box: `browseros ota server release --version <X> --channel alpha`, then unzip the published `browseros_server_<X>_darwin_arm64.zip` from `cdn.browseros.com/server/` and confirm it contains `resources/bin/browseros_server` + `resources/bin/third_party/{bun,rg,podman/*}`; run `codesign -dv` on a few inner binaries (especially `vfkit`, `krunkit`) to confirm entitlements are applied.
- [ ] `browseros ota server release-appcast --channel alpha`, install BrowserOS alpha on a clean machine, trigger auto-update, verify `~/Library/Application Support/BrowserOS/.browseros/versions/<X>/resources/bin/third_party/podman/podman` exists and runs.
- [ ] Promote to `prod` channel only after alpha verification.

### DIFF
--- a/packages/browseros-agent/apps/server/package.json
+++ b/packages/browseros-agent/apps/server/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@browseros/server",
-  "version": "0.0.85",
+  "version": "0.0.87",
   "description": "BrowserOS server",
   "type": "module",
   "main": "./src/index.ts",
--- a/packages/browseros-agent/bun.lock
+++ b/packages/browseros-agent/bun.lock
@@ -156,7 +156,7 @@
     },
     "apps/server": {
       "name": "@browseros/server",
-      "version": "0.0.85",
+      "version": "0.0.87",
       "bin": {
         "browseros-server": "./src/index.ts",
       },
--- a/packages/browseros/build/common/server_binaries.py
+++ b/packages/browseros/build/common/server_binaries.py
@@ -1,0 +1,58 @@
+#!/usr/bin/env python3
+"""Shared sign metadata for BrowserOS Server binaries.
+
+Consumed by both the Chromium-build signing path (build/modules/sign/) and the
+OTA release path (build/modules/ota/). Adding a new third-party binary here
+means both paths pick it up automatically.
+"""
+
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Dict, List, Optional
+
+
+@dataclass(frozen=True)
+class SignSpec:
+    """Per-binary codesign metadata.
+
+    ``entitlements`` is the filename of the plist under
+    ``resources/entitlements/``; ``None`` means no extra entitlements.
+    """
+
+    identifier_suffix: str
+    options: str
+    entitlements: Optional[str] = None
+
+
+MACOS_SERVER_BINARIES: Dict[str, SignSpec] = {
+    "browseros_server": SignSpec(
+        "browseros_server", "runtime", "browseros-executable-entitlements.plist"
+    ),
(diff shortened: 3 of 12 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #736, opened 2026-04-16
fix: use separate BrowserOS dir in development

shadowfax92:
## Summary
- Use `~/.browseros-dev` when `NODE_ENV=development` so local dev state does not mix with the production BrowserOS home directory.
- Preserve the existing production path by continuing to use `PATHS.BROWSEROS_DIR_NAME` outside development.
- Log the resolved development BrowserOS directory during startup and cover the behavior with focused regression tests.

## Design
Keep the change local to `apps/server/src/lib/browseros-dir.ts` by branching on `process.env.NODE_ENV` inside `getBrowserosDir()` and emitting a development-only info log during BrowserOS directory initialization. This follows the server's existing development-environment pattern without changing shared constants or adding new config.

## Test plan
- `bun --env-file=.env.development test apps/server/tests/browseros-dir.test.ts`
- `bunx biome check apps/server/src/lib/browseros-dir.ts apps/server/tests/browseros-dir.test.ts`
- `bun run --filter @browseros/server typecheck`
- `bun run test` currently fails in existing unrelated `apps/server/tests/tools/input.test.ts` CDP disconnect/timeouts (`fill`, `click`, `check`, `press_key`).

### DIFF
--- a/packages/browseros-agent/apps/server/src/lib/browseros-dir.ts
+++ b/packages/browseros-agent/apps/server/src/lib/browseros-dir.ts
@@ -6,8 +6,19 @@ import { PATHS } from '@browseros/shared/constants/paths'
 import type { ServerDiscoveryConfig } from '@browseros/shared/types/server-config'
 import { logger } from './logger'
 
+const DEV_BROWSEROS_DIR_NAME = '.browseros-dev'
+
 export function getBrowserosDir(): string {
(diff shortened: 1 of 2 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #737, opened 2026-04-16
fix: randomized tweet variations + referral fixes

felarof99:
## Summary
- Add 10 feature-specific tweet variations (5 Roy style, 5 Dax style) and pick one at random in `getShareOnTwitterUrl`
- Cap daily referral earnings at 500 credits client-side to prevent unlimited farming
- Fix vite dev-server crash by declaring `@browseros/shared` as a workspace dependency of the agent app

## Test plan
- [ ] `bun install` succeeds and `@browseros/shared` resolves in the agent app
- [ ] `bun run dev:watch -- --new` starts without the "Failed to resolve import" error
- [ ] Share on Twitter button produces a different tweet across clicks
- [ ] Once credits ≥ 500, submit is blocked client-side with the daily-cap message

### DIFF
--- a/packages/browseros-agent/apps/agent/components/referral/ShareForCredits.tsx
+++ b/packages/browseros-agent/apps/agent/components/referral/ShareForCredits.tsx
@@ -1,3 +1,4 @@
+import { REFERRAL_LIMITS } from '@browseros/shared/constants/limits'
 import { ExternalLink, Loader2, Send } from 'lucide-react'
 import type { FC } from 'react'
 import { useState } from 'react'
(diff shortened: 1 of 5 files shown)


felarof99:
Addressed both Greptile P1 comments:

**#1 Daily cap uses total balance** — Correct finding, but the fix requires a server change to expose `dailyReferralCredits` in the `/credits` response. The existing `dailyLimit` field represents the total daily free-agent allowance (see `UsagePage.tsx`), not daily referral earnings, so it can't be reused directly. Filed as follow-up; keeping the coarse balance check for now since it's a minor over-restriction (users at 500+ total can't double-dip).

**#2 Tweet variation chosen at render time** — Fixed in f310d376. Moved `getShareOnTwitterUrl()` into an `onClick` handler that updates `e.currentTarget.href` so a new random variation is picked on every click.

PR #745, opened 2026-04-17
fix: rerun dev port cleanup before server restarts

shadowfax92:
## Summary
- rerun `kill -9` port cleanup before every watched server launch, not only once at `watch` startup
- wait for the default dev ports to clear in both `watch` and `cleanup` so failures surface immediately instead of looping on `9105`
- add supervisor tests covering the new pre-start hook behavior

## Design
This keeps the fix local to `tools/dev`. The Go supervisor now supports an optional pre-start hook, and the watched server uses it to clear `BROWSEROS_SERVER_PORT` before every restart attempt. That matches the existing `kill -9` behavior the dev tooling already intended to use, while closing the restart-time race that left `bun --watch` stuck on a reclaimed port.

## Test plan
- `go test ./...` in `packages/browseros-agent/tools/dev`
- bind `127.0.0.1:9105`, run `./tools/dev/run.sh cleanup --ports`, confirm the listener is killed

### DIFF
--- a/packages/browseros-agent/tools/dev/cmd/cleanup.go
+++ b/packages/browseros-agent/tools/dev/cmd/cleanup.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"fmt"
+	"time"
 
 	"browseros-dev/proc"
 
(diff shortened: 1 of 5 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #746, opened 2026-04-17
fix: pin OpenClaw image to 2026.4.12

shadowfax92:
## Summary
- Pin the default OpenClaw image to `ghcr.io/openclaw/openclaw:2026.4.12` instead of `latest`
- Update both the compose template fallback and the generated compose env default so runtime startup uses the same pinned image
- Add a focused test covering `buildComposeEnvFile()` and document the reason for the pin inline

## Design
This keeps the change minimal and coherent by updating the two actual default sources that control the gateway container image: the compose template fallback and the env file generated by `buildComposeEnvFile()`. The pin is documented inline because the downgrade is intentional: newer OpenClaw releases regress OpenRouter chat streams, while `2026.4.12` is the known-good image we want BrowserOS to use by default unless an explicit override is supplied.

## Test plan
- `bun test ./tests/api/services/openclaw-env.test.ts`
- `bun run typecheck`

### DIFF
--- a/packages/browseros-agent/apps/server/resources/openclaw-compose.yml
+++ b/packages/browseros-agent/apps/server/resources/openclaw-compose.yml
@@ -1,6 +1,7 @@
 services:
   openclaw-gateway:
-    image: ${OPENCLAW_IMAGE:-ghcr.io/openclaw/openclaw:latest}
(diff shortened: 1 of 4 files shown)


shadowfax92:
@claude review

shadowfax92:
@greptileai review

PR #771, opened 2026-04-21
fix(openclaw): allocate gateway host port dynamically + name the two ports distinctly

shadowfax92:
## Summary

- The gateway container always listens on **18789 internally**, but that port may be taken on the host (seen in the wild: a stale `gvproxy` from a previous container). The host-side mapping port is now **allocated dynamically** on each lifecycle transition and persisted to `~/.browseros/openclaw/.openclaw/runtime-state.json` so the choice is stable across restarts.
- The shared constant becomes `OPENCLAW_GATEWAY_CONTAINER_PORT` (container-internal, fixed) and the service/spec/runtime probes/chat-client now use `hostPort` for the dynamically mapped host-side port — so the two sides of the `-p 127.0.0.1:<host>:<container>` mapping are no longer ambiguous.
- Small UI follow-up: removed a duplicate Podman overrides card from the status panels.

## Design

Simple linear port scan modeled on `scripts/dev/start.ts`: `isPortAvailable` via `net.createServer.listen('127.0.0.1')`, `findAvailablePort` scans upward from `OPENCLAW_GATEWAY_CONTAINER_PORT`. `allocateGatewayPort(openclawDir)` prefers the persisted port when still free, otherwise scans and persists the winner. Lives in a new `apps/server/src/api/services/openclaw/runtime-state.ts`, mirroring the shape of `podman-overrides.ts`.

Allocation runs once per lifecycle method:
- `setup()` allocates right after `pullImage` so `setConfigBatch` writes `gateway.controlUi.allowedOrigins` with the correct host port, and `runOnboard`'s `gatewayPort` argument switches to the container-internal constant (always 18789).
- `start()` and `restart()` allocate just before `runtime.startGateway` / `restartGateway`.
- `tryAutoStart()` loads the persisted port first so `isReady()` probes the right port; it only reallocates when the gateway isn't already running.

The chat client captures the port at construction, so it's rebuilt whenever `setPort` changes the host port. `/status` and `/setup` both return the real allocated port via `getOpenClawService().getPort()`; the extension already reads `port` from `/status`, so no client changes were needed.

## Test plan

- [x] `bun run typecheck` — clean across all packages
- [x] `bun test apps/server/tests/api/services/openclaw/` — 73/73 pass
- [x] Verified allocator end-to-end: on a machine where `gvproxy` held 18789, the service correctly allocated **18790** and persisted it to `~/.browseros/openclaw/.openclaw/runtime-state.json`
- [ ] Manual: fresh setup on a clean host → observe `runtime-state.json` created with `{"gatewayPort": 18789}`, gateway reachable at `http://127.0.0.1:18789`
- [ ] Manual: bind 18789 with `nc -l 18789` before setup → observe allocator picks 18790 and persists it; restart with 18789 released → observe persisted 18790 is reused
- [ ] Manual: confirm the status panels no longer render two Podman overrides cards

### DIFF
--- a/packages/browseros-agent/apps/agent/entrypoints/app/agents/AgentsPage.tsx
+++ b/packages/browseros-agent/apps/agent/entrypoints/app/agents/AgentsPage.tsx
@@ -229,17 +229,13 @@ const ProviderSelector: FC<ProviderSelectorProps> = ({
   )
 }
 
-interface PodmanOverridesCardProps {
-  variant: 'inline' | 'standalone'
-}
-
-const PodmanOverridesCard: FC<PodmanOverridesCardProps> = ({ variant }) => {
+const PodmanOverridesCard: FC = () => {
   const { overrides, loading, saving, error, saveOverrides, clearOverrides } =
     usePodmanOverrides()
 
   const [value, setValue] = useState('')
   const [touched, setTouched] = useState(false)
-  const [collapsed, setCollapsed] = useState(variant === 'standalone')
+  const [collapsed, setCollapsed] = useState(true)
   const [localError, setLocalError] = useState<string | null>(null)
 
   useEffect(() => {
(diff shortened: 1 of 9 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #774, opened 2026-04-21
feat(openclaw): make podman machine CPUs + memory configurable

shadowfax92:
## Summary

- `podman machine init` previously hardcoded **8 CPUs / 8096 MB RAM**, which over-allocates on 16 GB dev laptops and forced devs to re-init the machine by hand.
- Exposes two env vars — `BROWSEROS_PODMAN_CPUS` and `BROWSEROS_PODMAN_MEMORY_MB` — read at init time.
- Lowers defaults to **4 CPUs / 4096 MB** so a fresh `ensureReady()` boots a machine that runs the gateway comfortably without starving the host.

## Design

Values are read directly inside `initMachine()` via a small helper (`readMachineResources()`) rather than threaded through `configurePodmanRuntime()`. Rationale: the two values are only consumed at one call site, `configurePodmanRuntime` already has two callers (`main.ts` + `openclaw-service.applyPodmanOverrides`), and plumbing them through both would be churn without a testability win — the existing `FakePodmanRuntime` already overrides `initMachine` entirely. Invalid or non-positive env values fall back to defaults so malformed config never reaches the CLI. Linux path is unchanged (machine init is a no-op there). Disk size stays at 10 GB — not in scope.

`.env.example` documents both vars under a "Podman machine (macOS/Windows VM)" section. `.env.development` is gitignored; devs copy from the example.

## Test plan

- [x] `bun x @biomejs/biome check` on changed files — clean
- [x] `bun run typecheck` — all workspace packages pass
- [x] `bun test apps/server/tests/api/services/openclaw/` — 80/80 tests pass (9 in `podman-runtime.test.ts`)
- [ ] On macOS: `rm -rf ~/.local/share/containers/podman/machine/applehv/*` (or destroy the existing machine), set `BROWSEROS_PODMAN_CPUS=2` + `BROWSEROS_PODMAN_MEMORY_MB=2048`, start the server, confirm the init log line shows the configured values and `podman machine inspect` reflects them.
- [ ] On Linux: start the server, confirm init still no-ops (no CLI spawned, no log line).
- [ ] With `BROWSEROS_PODMAN_CPUS=abc` (malformed), confirm default `4` is used.

### DIFF
--- a/packages/browseros-agent/apps/server/.env.example
+++ b/packages/browseros-agent/apps/server/.env.example
@@ -26,5 +26,9 @@ LOG_LEVEL=info
 # Debug — captures every LLM call to .devtools/generations.json (view with `npx @ai-sdk/devtools`)
 # BROWSEROS_AI_SDK_DEVTOOLS=true
 
+# Podman machine (macOS/Windows VM). No-op on Linux.
+BROWSEROS_PODMAN_CPUS=4
+BROWSEROS_PODMAN_MEMORY_MB=4096
+
 # Testing
 BROWSEROS_TEST_HEADLESS=false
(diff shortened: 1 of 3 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #794, opened 2026-04-23
feat(build-tools): add cache:sync:dev for local tarball seeding

shadowfax92:
## Summary
- Adds `cache:sync:dev` script that seeds `~/.browseros-dev/cache/vm/` from `./dist/` without touching R2, so devs can test the server against a freshly-built tarball before anything is published to `cdn.browseros.com`.
- Hardcoded to `arm64` (all devs on Apple Silicon); refuses to run unless `NODE_ENV=development`; idempotent — skips the copy if the cache already has a tarball with matching sha256.
- Fixes `R2_BUCKET` default in `.env.sample` from `browseros-artifacts` to `browseros` to match the actual bucket name.

## Design
Small, dev-only sibling to `cache:sync`. Reads `bundle.json` + `./dist/images/<name>-<version>-arm64.tar.gz`, hashes the tarball, copies it into `~/.browseros-dev/cache/vm/images/` under the manifest key, and writes an arm64-only manifest last so a partial run never leaves a manifest that claims files that aren't on disk. No R2 credentials, no `emit-manifest` step, no x64 build — minimal dev loop is `build:tarball --arch arm64` then `cache:sync:dev`.

## Test plan
- [x] Fresh seed from empty cache: `rm -rf ~/.browseros-dev/cache/vm && NODE_ENV=development bun run --filter @browseros/build-tools cache:sync:dev` → prints `seeded vm/images/openclaw-2026.4.12-arm64.tar.gz` and writes `manifest.json`.
- [x] Idempotent second run: same command → prints `cache hit: vm/images/openclaw-2026.4.12-arm64.tar.gz`, no copy.
- [x] Production-mode refusal: `NODE_ENV=production bun run --filter @browseros/build-tools cache:sync:dev` → exits 1 with explicit error.
- [x] Missing tarball: delete `./dist/images/openclaw-*.tar.gz` → exits with pointer to `build:tarball` command.
- [x] `bun run typecheck` clean.

### DIFF
--- a/packages/browseros-agent/packages/build-tools/.env.sample
+++ b/packages/browseros-agent/packages/build-tools/.env.sample
@@ -2,7 +2,7 @@
 R2_ACCOUNT_ID=
 R2_ACCESS_KEY_ID=
 R2_SECRET_ACCESS_KEY=
-R2_BUCKET=browseros-artifacts
+R2_BUCKET=browseros
 
 # Public CDN base - used by cache:sync to GET manifest and artifacts
 R2_PUBLIC_BASE_URL=https://cdn.browseros.com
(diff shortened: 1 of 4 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #830, opened 2026-04-27
fix(alpha): write balpha process logs

shadowfax92:
**Summary**
- Tee `balpha start` BrowserOS/Chromium output to `chromium.log` and Bun server output to `server.log`.
- Store logs under the copied alpha profile at `dev_user_data_dir/logs` and rotate files older than one day to `<name>.old`.
- Add `balpha logs` so users can print the log directory and file paths for debugging.

**Design**
The change keeps logging inside the existing Go alpha CLI supervisor. `Config` now derives a profile-local log directory, managed processes optionally tee tagged stdout/stderr to file, and startup wires BrowserOS and the server to separate log files while preserving terminal output. Chromium is launched with `--enable-logging=stderr` so browser logs flow through the same process stream.

**Test plan**
- `cd packages/browseros-agent/tools/alpha && go test -count=1 ./...`
- `git diff --check`

### DIFF
--- a/packages/browseros-agent/tools/alpha/README.md
+++ b/packages/browseros-agent/tools/alpha/README.md
@@ -92,6 +92,7 @@ Each start:
 - Runs `tools/dev/setup.sh`.
 - Builds the WXT extension.
 - Starts BrowserOS and the local Bun server.
+- Tees BrowserOS and server output to log files under the copied profile.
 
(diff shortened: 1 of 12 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #836, opened 2026-04-28
fix(dogfood): separate BrowserOS state root

shadowfax92:
**Summary**
- Adds a `browseros_dir` dogfood config field defaulting to `~/.browseros-dogfood`.
- Launches the dogfood server with `BROWSEROS_DIR` set from config so server, OpenClaw, VM cache, and Lima state stay separate from `.browseros` and `.browseros-dev`.
- Shows the BrowserOS dir in start/status output and documents the config field.

**Design**
Dogfood now uses the existing server `BROWSEROS_DIR` override instead of changing server defaults. The CLI owns the dogfood-specific state root, injects it into the server process environment, and strips any inherited `BROWSEROS_DIR` so the dogfood environment is deterministic while still allowing users to edit the generated config.

**Test plan**
- `cd packages/browseros-agent/tools/dogfood && go test ./...`


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #837, opened 2026-04-28
fix(dogfood): use alpha dock icon

shadowfax92:
## Summary
- Pass `--browseros-dock-icon=alpha` whenever `browseros-dogfood` launches BrowserOS.
- Cover the launch argument in `browser.BuildArgs` tests.
- Document that dogfood sessions use the alpha Dock icon so they are visually distinct.

## Design
The dogfood launcher hard-codes the alpha Dock icon category in its existing BrowserOS argument builder, alongside the other dogfood-only launch switches. This keeps the behavior automatic for both foreground and background dogfood starts without adding user config or changing process orchestration.

## Test plan
- `git diff --check`
- `go test -count=1 ./...` from `packages/browseros-agent/tools/dogfood`


shadowfax92:
@claude review

shadowfax92:
@greptileai review

PR #845, opened 2026-04-28
feat(openclaw): migrate chat to WS chat.send + add chat abort

shivammittal274:
## Summary

Three things shipped together because they touch the same file:

- **Fix observer handshake (production breakage).** Was failing with `CONTROL_UI_DEVICE_IDENTITY_REQUIRED` on every reconnect. Switched `client.id` from `'openclaw-tui'` to `'gateway-client'` and `mode` from `'ui'` to `'backend'` — OpenClaw's documented "trusted backend self-connect" identity. Dropped the explicit `operator.read` scope so the gateway grants default operator scopes for shared-secret auth. Dashboard SSE / queue dispatch now work again.
- **Migrate chat send from HTTP to WS.** Was `POST /v1/chat/completions` (SSE). Now WS `chat.send`. The reason isn't speed — it's that HTTP-initiated runs are NOT in OpenClaw's `chatAbortControllers` registry, so closing the SSE fetch never aborted the agent. PoC verified: agent kept generating for 70+ seconds after `fetch.abort()`. WS `chat.send` does register, which is what makes the Stop button actually stop things.
- **Add `POST /claw/agents/:id/chat/stop`.** New abort endpoint. Calls `chat.abort` over WS via the OpenClaw CLI from inside the gateway container so the connection appears as `direct_local` (required for `operator.write` scope). Live-verified end-to-end: returns `{aborted: true, runIds: […]}` and the in-flight stream closes with `lifecycle:aborted`.

## Notable design choices

- **Why exec the CLI inside the container instead of opening our own WS?** OpenClaw clears requested operator scopes for non-`direct_local` connections. Calling `chat.send` / `chat.abort` from outside the container loses `operator.write`. Calling via `nerdctl exec node dist/index.js gateway call <method>` keeps the connection direct_local → full operator scope.
- **No device pairing.** Per the static-UUID + bearer-auth constraint, we don't pair a device.
- **HTTP path deleted, no fallback.** `OpenClawHttpClient.streamChat` and ~290 lines of SSE/chunk parsing are gone. `isAuthenticated()` and session history reads stay on HTTP. Single chat path, no feature flag.
- **Idempotency.** `chat.send` requires `idempotencyKey`; we generate a per-send UUID, so retries are safe.
- **Multimodal.** OpenAI-shaped `messageParts` translate to OpenClaw's `attachments` shape inline.

## Richer event stream

The new stream emits more than the old `text-delta` / `done` / `error`: `reasoning-delta`, `tool-start` (with arguments + label + subject), `tool-end` (with `durationMs` + `isError`), `lifecycle` (start/end/aborted/error), `usage` (tokensIn / tokensOut / costUsd). Existing extension UI gracefully ignores unknown types — no extension changes required to ship this server-side; surfacing the new types is a follow-up UI PR.

## Test plan

- [x] `bun run typecheck` passes
- [x] `bun --env-file=apps/server/.env.development test apps/server/tests/api/services/openclaw` passes (one pre-existing failure on `restart moves off a persisted ready port when auth rejects the current token` — verified failing on `dev` before this PR)
- [x] Live PoC against running gateway: handshake succeeds with `gateway-client` + `backend`
- [x] Live PoC: `chat.send` returns `runId` and broadcasts events to the observer
- [x] Live PoC: `chat.abort` actually aborts (saw `aborted: true` and stream close)
- [ ] Manual: smoke-test chat from BrowserOS extension (text + image attachment)
- [ ] Manual: click Stop mid-execution, verify agent halts within ~1s
- [ ] Manual: queue dispatch on idle still works (it watches `ClawSession` which is fed by the observer)


shivammittal274:
Pivoting: Nikhil's browseros-openclaw fork adds gateway.auth.mode=none + private-ingress no-auth (https://github.com/browseros-ai/openclaw/pull/1). With that we can run chat.send/abort directly over the observer's WS — no CLI exec, no token. Reworking on a fresh branch off dev. Closing in favor of the simpler approach. Branch preserved for reference / cherry-picking the observer fix + dedupe + stop button if useful.

PR #856, opened 2026-04-29
fix: hide BrowserOS ACP wrapper in history

shadowfax92:
## Summary
- Hide BrowserOS ACP wrapper instructions from persisted user-message history.
- Render only the original user request while preserving the wrapped prompt sent to acpx/runtime.
- Add regression coverage for escaped user requests inside the wrapper.

## Design
The fix stays in the BrowserOS acpx history mapping layer: persisted text content is unwrapped only when it exactly matches the BrowserOS ACP prompt wrapper, then XML entities from the wrapped user request are decoded for display. Runtime prompt submission is unchanged until acpx/runtime exposes system prompt support.

## Test plan
- `bun test apps/server/tests/lib/agents/acpx-runtime.test.ts`
- `bun run --filter @browseros/server typecheck`

### DIFF
--- a/packages/browseros-agent/apps/server/src/lib/agents/acpx-runtime.ts
+++ b/packages/browseros-agent/apps/server/src/lib/agents/acpx-runtime.ts
@@ -274,10 +274,30 @@ function mapToolUseToHistoryToolCall(
 }
 
 function userContentToText(content: AcpxUserContent): string {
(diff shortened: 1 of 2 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #857, opened 2026-04-29
feat: add sidepanel ACP chat targets

shadowfax92:
## Summary
- Adds sidepanel chat targets so normal LLM providers and ACP-backed Claude Code/Codex models share one selector.
- Adds a sidepanel ACP chat route that converts ACP runtime events into AI SDK UI-message SSE for the existing sidepanel chat renderer.
- Routes sidepanel sends by selected target while preserving LLM /chat behavior and resetting active conversations on LLM/ACP target switches.

## Design
The sidepanel now resolves a selected chat target as either an LLM provider or an ACP model. LLM sends keep the existing /chat request body. ACP sends post explicit adapter/model/reasoning fields to /agents/sidepanel/chat, where the server creates a virtual conversation-scoped ACP agent, formats browser context consistently with /chat, and streams AI SDK UI-message chunks back to useChat.

## Test plan
- [x] bun run typecheck
- [x] bun run test:agent
- [x] bun ./tests/__helpers__/run-test-group.ts api (from apps/server)
- [x] bun --env-file=.env.development test --preload=./tests/__helpers__/test-env.ts ./tests/lib/agents/acp-ui-message-stream.test.ts (from apps/server)
- [x] Sidepanel CDP smoke test: selector opens, ACP options render, Codex GPT-5.5 selection updates header, no sidepanel JS errors

### DIFF
--- a/packages/browseros-agent/apps/agent/components/chat/ChatProviderSelector.helpers.ts
+++ b/packages/browseros-agent/apps/agent/components/chat/ChatProviderSelector.helpers.ts
@@ -31,8 +31,8 @@ export function getProviderSearchValue(
     .join(' ')
 }
 
-export function getProviderSubtitle(provider: Provider): string | null {
-  if (provider.kind !== 'acp') return null
+export function getProviderSubtitle(provider: Provider): string | undefined {
+  if (provider.kind !== 'acp') return undefined
   return provider.modelControl === 'best-effort'
     ? 'ACP model · best effort'
     : 'ACP model'
--- a/packages/browseros-agent/apps/agent/components/chat/ChatProviderSelector.helpers.ts
+++ b/packages/browseros-agent/apps/agent/components/chat/ChatProviderSelector.helpers.ts
@@ -1,0 +1,39 @@
+import type { Provider } from './chatComponentTypes'
+
+export interface ProviderOptionGroup {
+  key: 'llm' | 'acp'
+  label: string
+  options: Provider[]
+}
+
+export function groupProviderOptions(
+  providers: Provider[],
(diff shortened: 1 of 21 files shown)


shadowfax92:
@claude review

shadowfax92:
@greptileai review

shadowfax92:
Claude review pass addressed in dcdda78c.

Accepted and pushed:
- Loosened conversationId UUID validation to accept future UUID versions while retaining shape/variant checks.
- Made Provider.kind required and updated scheduled-task provider option construction; simplified the sidepanel target lookup accordingly.
- Switched getProviderSubtitle absent value to undefined.
- Added the ACP request-builder note explaining why previousConversation/approval-only LLM fields are not forwarded.
- Added ACP stream interleaving handling and failed-tool errorText fallback from the Greptile findings.

Not changing:
- Rate limiting: /chat currently has no rate-limiting middleware to mirror; adding a one-off limiter only to /agents/sidepanel/chat would diverge from current route policy.
- Sidepanel ACP session pruning: valid retention concern, but acpx exposes load/save for this store today, not a delete/prune API. Route-level filesystem pruning would be a separate storage-policy change.
- Re-emitting running tool description updates: the AI SDK UI-message stream does not expose a safe update primitive for the same toolCallId. The first event stays as input; final completed/failed events carry output/error text.

shadowfax92:
Greptile P2 findings addressed in dcdda78c:

- Selection persistence failures now report to Sentry instead of being silently dropped.
- ACP text/reasoning interleaving now closes/reopens parts with segment IDs and has a regression test.
- Failed tool calls now emit a non-empty fallback errorText and have a regression test.

PR #860, opened 2026-04-29
feat(eval): align Clado action executor with new endpoint contract

shivammittal274:
## Summary

Updates the Clado action executor to match the spec David Shan shared in the new README.

- **Endpoint + model bump** — points at the `000159-merged` checkpoint at `clado-ai--clado-browseros-action-000159-merged-actionmod-f4a6ef.modal.run`. Old URL is stale.
- **Cold-start timeout** — `CLADO_REQUEST_TIMEOUT_MS` 120 s → 360 s. Per Clado, cold start is ~5 min with 30 min cooldown; the old ceiling was failing every cold-start request.
- **Parse-failure handling** — when Clado returns HTTP 200 with `action: null` and `parse_error`, log it as an `invalid` step in history (so the model self-corrects) instead of aborting the executor loop. Cap consecutive parse failures at 3 to avoid infinite loops.
- **`final_answer` capture** — `end` actions carrying `final_answer` now propagate the answer text into the executor's observation back to the orchestrator.
- **Missing key mappings** — added `M-a`, `M-c`, `M-v`, `M-x` (Meta+A/C/V/X) for macOS Cmd shortcuts.
- **Screenshot format** — `webp` → `png` to match the documented "PNG or JPEG" contract.

Skipped pre-warm: David offered `min_containers=1` if we coordinate the day before, so we don't need to ping the health endpoint from each worker on construction.

## Test plan

- [x] `bun run typecheck` — clean
- [ ] CI: trigger `eval-weekly.yml` with `configs/browseros-oe-clado-weekly.json` once OpenClaw Linux fix lands so the server can boot

### DIFF
--- a/packages/browseros-agent/apps/eval/README.md
+++ b/packages/browseros-agent/apps/eval/README.md
@@ -66,9 +66,9 @@ The orchestrator works with any LLM provider. The executor can be another LLM, o
     },
     "executor": {
       "provider": "clado-action",
-      "model": "qwen3-vl-30b-a3b-instruct",
+      "model": "Qwen3.5-35B-A3B-action-000159-merged",
       "apiKey": "",
-      "baseUrl": "https://clado-ai--clado-browseros-action-actionmodel-generate.modal.run"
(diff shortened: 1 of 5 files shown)


shadowfax92:

PR #863, opened 2026-04-29
feat(agents): background-resilient chat — turns survive tab disconnect

DaniAkash:
## Summary

- New `ActiveTurnRegistry` decouples each agent turn's lifecycle from the SSE response that initiated it, so chat tabs that close, refresh, or navigate away no longer cancel an in-flight turn.
- New endpoints: `GET /agents/:id/chat/active` (discover an in-flight turn on mount), `GET /agents/:id/chat/stream` (subscribe / resume via `Last-Event-ID`), `POST /agents/:id/chat/cancel` (explicit Stop). `POST /chat` now returns `409 + attachUrl` when the agent already has a running turn.
- The chat hook captures `X-Turn-Id`, tracks `lastSeq` from per-event SSE `id:` lines, reattaches on mount when `/chat/active` reports a running turn, and routes Stop through the cancel endpoint. Streaming UX on the happy path is unchanged.

## Why this matters

Previously, closing the chat tab — or even a brief network blip — fired the request's AbortSignal into the runtime and killed the in-flight turn. The user came back to a half-rendered message and had to retry. The runtime call now uses the registry's per-turn AbortController instead of the HTTP request signal: the turn keeps running on the server, buffered events fan out to whoever is currently subscribed, and reconnects pick up via `Last-Event-ID`.

## What changes for users

| Scenario | Before | After |
|---|---|---|
| Stream a message to completion | Works | Identical |
| Switch tabs / refresh mid-turn | Turn cancelled, half-rendered output | Reattaches, replays buffered events, message completes |
| Press Stop | SSE close cancels by accident | Explicit `/chat/cancel` |
| Send while a turn is running | Silent queue/double-send | Server returns 409, UI attaches to the existing turn |

## Test plan

- [x] `bun test` for the registry: ring buffer overflow + truncated flag, terminal-frame preservation, replay from arbitrary `lastSeq`, cancel emits synthetic terminal, retain-window sweep.
- [x] `bun test` for the routes: chat emits per-event seq + `X-Turn-Id`, 409 collision returns `attachUrl`, `/chat/active` reports running turn, `/chat/stream?turnId=…` replays buffered events, `/chat/cancel` flips the turn to `cancelled` with a synthetic `done`, 404 on unknown turn.
- [x] Server typecheck, agent typecheck, agent lint, full server `test:api` (142) + `test:lib` (118), agent `bun test` (42).
- [ ] Manual: open chat, send a long-running message, close the tab mid-turn, reopen — message resumes and completes.
- [ ] Manual: send a message, press Stop — server-side turn cancels and any reattaching subscriber sees the cancellation terminal.
- [ ] Manual: open chat in two tabs on the same agent — both render the same stream in sync.

## Notes

- Server-restart resilience is out of scope. The retain window after a turn ends is 5 minutes; bounded buffer is 5000 frames with drop-oldest (terminal frame held in a separate slot).
- Sidepanel ACP route (`POST /agents/sidepanel/chat`) still uses the legacy stateless flow — it doesn't have a persistent agent identity.
- The openclaw bridge cancel-propagation gotcha is unchanged here; tracked separately.

### DIFF
--- a/packages/browseros-agent/apps/agent/entrypoints/app/agent-command/useAgentConversation.ts
+++ b/packages/browseros-agent/apps/agent/entrypoints/app/agent-command/useAgentConversation.ts
@@ -1,7 +1,10 @@
 import { useEffect, useRef, useState } from 'react'
 import {
   type AgentHarnessStreamEvent,
+  attachToHarnessTurn,
+  cancelHarnessTurn,
   chatWithHarnessAgent,
+  fetchActiveHarnessTurn,
 } from '@/entrypoints/app/agents/useAgents'
 import type { OpenClawChatHistoryMessage } from '@/entrypoints/app/agents/useOpenClaw'
 import type {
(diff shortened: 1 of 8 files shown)


DaniAkash:
@greptileai

PR #864, opened 2026-04-29
fix(server): MCP tool fidelity — drag dnd, click react fallback, fill validate, snapshot carousel

shivammittal274:
## Summary

Four MCP tool fixes identified by the K2.5 + Opus 4.6 head-to-head deep-dive as structural blockers (both models hit the same walls regardless of model capability). Split out from #844 (which now scopes to dataset hygiene only) for cleaner review.

### 1. `drag` → HTML5 dnd events (`apps/server/src/browser/mouse.ts`)
**Bug:** CDP `Input.dispatchMouseEvent` does NOT elevate to HTML5 `dragstart`/`dragover`/`drop` events. React-DnD calendar libs (used by `evals-gocalendar.vercel.app`) listen for HTML5 events, so drag "succeeds" at the tool layer but creates a duplicate event instead of moving the original.
**Fix:** after the mouse event sequence, dispatch synthetic `DragEvent`s via `Runtime.evaluate` with a shared `DataTransfer`. Apps using either drag style now work.

### 2. `click` → JS click follow-up (`apps/server/src/browser/browser.ts`)
**Bug:** Existing JS click was a fallback only when CDP click *threw*. Some React-controlled submit buttons (staynb's "Confirm and pay") silently swallow the native click — the button visually disabled but no submission fired.
**Fix:** after CDP mouse sequence, also call `element.click()` via JS. Idempotent for most apps (React already saw native click); guarantees handler fires for the swallowed cases. Skipped for non-left/multi-click (right-click, double-click are mouse-semantic).

### 3. `fill` → post-validate value persistence (`apps/server/src/browser/browser.ts` + `tools/input.ts`)
**Bug:** For React-controlled date textboxes and MUI comboboxes, keyboard typing fires native input events but the controlled state ignores them. Tool reports "Typed N characters" with no signal that the value never changed; agent proceeds with stale state.
**Fix:** after typing, read `element.value`. If it doesn't contain the typed text, surface a `WARNING: value did not persist — element may be a controlled component...` in the response so the model knows to try the picker-click flow. Tool itself still succeeds (typing happened); the warning is informational.

### 4. `snapshot` → surface horizontal-carousel scroll hints (`apps/server/src/browser/snapshot.ts` + `browser.ts`)
**Bug:** Snapshot returns AX-tree elements visible in viewport. Carousels show 3-4 items at a time; offscreen slides aren't included. Models then judge "all X" against an incomplete enumeration (e.g. `opendining-10` missed "The Vegan Table" because the homepage carousel was never horizontally-scrolled).
**Fix:** detect horizontally-scrollable carousel containers via `scrollWidth > clientWidth + 50px`; count visible vs total children; append `[scroll-hint] carousel "X": N items, ~3 visible — scroll right to reveal more` lines.

## Risk

These changes affect every server tool call (eval + production agent extension), not just eval. Each was scoped to be additive:

- **drag**: hybrid mouse + dnd. Apps listening to one style get the other as a no-op. Only risk: apps listening to BOTH (rare and already buggy).
- **click**: JS click is idempotent for most apps. Risk: forms without proper disabled-on-submit could double-fire — but those are already broken.
- **fill**: warning-only, no behavior change. Caller decides what to do with the warning.
- **snapshot**: purely additive. New `[scroll-hint]` lines easy for models to ignore if irrelevant.

## E2E validation (already done)

Pre-pushed-to-this-branch validation against a synthetic test page (`/tmp/tool-test.html`) hitting each failure mode in isolation:

| Fix | Result |
|---|---|
| #1 drag HTML5 dnd | `drop-status: "DROPPED"` — synthetic `drop` event fired ✅ |
| #2 click JS fallback | `click-status: "CLICKED"` — onclick handler fired ✅ |
| #3 fill post-validate | `WARNING: value did not persist` surfaced for controlled-input no-op ✅ |
| #4 snapshot scroll-hint | `[scroll-hint] carousel "Test items": 8 items, ~3 visible — scroll right to reveal more` ✅ |

## Test plan

- [x] `bun run typecheck` clean
- [x] `bun run lint` zero issues in touched files
- [x] E2E synthetic test (above)
- [ ] Live AGISDK eval validation — needs to be run after this lands

### DIFF
--- a/packages/browseros-agent/apps/server/src/browser/browser.ts
+++ b/packages/browseros-agent/apps/server/src/browser/browser.ts
@@ -759,6 +759,19 @@ export class Browser {
       // cursor detection is best-effort; AX tree results are still returned
     }
 
+    // Surface horizontally-scrollable carousel hints so the model knows
+    // offscreen content exists. Helps tasks like "find the most X among
+    // all restaurants/products" where the visible 3-4 items aren't the full set.
+    try {
+      const hints = await snapshot.findScrollHints(session)
+      if (hints.length > 0) {
+        lines.push('')
+        for (const h of hints) lines.push(`[scroll-hint] ${h}`)
+      }
+    } catch {
+      // best-effort
+    }
+
     return lines.join('\n')
   }
 
@@ -1044,6 +1057,16 @@ export class Browser {
 
     await elements.scrollIntoView(session, element)
 
+    // Two-phase click for React-controlled buttons:
+    //   Phase 1: CDP mouse-down/move/up. This is the realistic path —
+    //            triggers hover, focus, and most native event handlers.
+    //   Phase 2: element.click() via JS. Some React-controlled <button>s
+    //            (e.g. staynb's "Confirm and pay") have an onClick that
+    //            is silently swallowed when only native mouse events fire.
+    //            JS click() is idempotent for most apps because React's
+    //            event delegation already saw the native click; for the
+    //            cases where it didn't, this guarantees the handler runs.
+    let coords: { x: number; y: number } | undefined
     try {
       const { x, y } = await elements.getElementCenter(session, element)
       await mouse.dispatchClick(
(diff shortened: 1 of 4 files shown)


shadowfax92:

PR #865, opened 2026-04-29
feat: show created agents in sidepanel

shadowfax92:
## Summary
- Shows persisted harness agents from `GET /agents` in the sidepanel selector instead of adapter/model catalog targets.
- Routes sidepanel agent sends through `/agents/:agentId/sidepanel/chat`, preserving the created agent identity, session key, history, model, and reasoning settings.
- Removes product use of the legacy virtual `/agents/sidepanel/chat` path while keeping stale old selections recoverable to the default LLM provider.

## Design
The sidepanel target catalog now combines normal LLM providers with persisted harness agents and uses adapter metadata only for labels. Created-agent sends post contextual sidepanel payloads to an agent-id route that formats browser context and selected text, starts the persisted harness turn, converts turn frames to AI SDK UI-message stream chunks, and cancels the turn on request abort.

## Test plan
- [x] bun run test:agent
- [x] bun --cwd apps/server ./tests/__helpers__/run-test-group.ts api
- [x] bun run --filter @browseros/agent typecheck

### DIFF
--- a/packages/browseros-agent/apps/agent/components/chat/ChatProviderSelector.helpers.ts
+++ b/packages/browseros-agent/apps/agent/components/chat/ChatProviderSelector.helpers.ts
@@ -17,7 +17,7 @@ export function groupProviderOptions(
       ? [{ key: 'llm' as const, label: 'AI Providers', options: llm }]
       : []),
     ...(acp.length
-      ? [{ key: 'acp' as const, label: 'ACP Models', options: acp }]
+      ? [{ key: 'acp' as const, label: 'Agents', options: acp }]
       : []),
   ]
 }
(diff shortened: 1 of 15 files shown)


shadowfax92:
@claude review

shadowfax92:
@greptileai review

shadowfax92:
Addressed the two concrete P2 findings from the bot reviews in 793ee5e5:

- `formatAdapterName` now handles `openclaw` explicitly and falls back to the raw adapter id instead of mislabeling unknown adapters as OpenClaw.
- `turnFramesToAgentEvents` now releases the frame reader lock if `read()` rejects before rethrowing.

Not changing the broader notes here:
- `conversationId` validation is intentional request-contract validation from the plan; persisted harness history remains keyed by `agentId`.
- Text length caps would be a broader API policy change because the existing LLM chat path also accepts unbounded text fields.
- The agents/adapters loading gate is intentional to avoid resolving a persisted created-agent selection as stale before agents load; TanStack Query leaves loading on error.
- The remaining minor notes are either deliberate selector metadata, pre-existing code, or would add comments where tests already express the behavior.

PR #876, opened 2026-04-30
refactor(eval): split orchestrated executor backends

shadowfax92:
## Summary
- Split the orchestrated eval executor path into concrete ToolLoop and Clado backend classes.
- Removed the old mixed `Executor` adapter/switch path from `orchestrator-executor`.
- Added backend-boundary coverage that asserts the factory creates distinct backend implementations.

## Design
The orchestrator now depends only on the shared `ExecutorBackend` interface. `createExecutorBackend` resolves the provider and returns either `ToolLoopExecutorBackend` or `CladoExecutorBackend`, so Clado and tool-loop behavior are clear, independently named paths instead of one adapter toggled by `isCladoAction`.

## Test plan
- `bun test apps/eval/tests`
- `bun run --cwd apps/eval typecheck`


shadowfax92:
@claude review

shadowfax92:
@greptileai review

shadowfax92:
Addressed review feedback in `dd73a4f5`:

- Removed unused Clado window/tab constructor properties instead of storing public `_windowId` / `_tabId` fields.
- Removed the dead `backendKind` option from the fake-backend forwarding test.
- Documented the no-op tool-loop `close()` lifecycle.
- Reworked `browserContext()` to receive a non-null `Browser` local instead of relying on optional chaining or a non-null assertion.

Local verification after the fixes:
- `bun test apps/eval/tests`
- `bun run --cwd apps/eval typecheck`

PR #939, opened 2026-05-04
feat(openclaw): aggregate sub-session history into agent main session

shivammittal274:
## Problem

Autonomous OpenClaw runs (cron, hook, channel) execute and persist on disk, but the BrowserOS chat panel never sees them. Each run gets its own ephemeral session file under the parent agent's directory:

\`\`\`
/home/node/.openclaw/agents/<agentId>/sessions/<runId>.jsonl
\`\`\`

The panel queries \`agent:<agentId>:main\` and gets back only that one file. The hellos a user scheduled earlier today fired correctly but were invisible.

## Change

When \`getSessionHistory\` is called with a main-session key (\`^agent:[^:]+:main$\`), enumerate every session under that agent via the existing \`sessions.list\` gateway RPC and merge their messages into one chronological response.

Each merged message gets:

- \`source\`: \`main\` | \`cron\` | \`hook\` | \`channel\` | \`other\` (parsed from the third segment of the session key)
- \`subSessionKey\`: only when the message originates outside the main session, so the UI can group / render section markers without re-parsing

Single-session keys (anything not matching the main pattern) still take the existing per-session HTTP path, so callers asking for a specific sub-session key keep working untouched.

## Why no agent-collision risk

\`sessions.list({ agentId })\` is **filesystem-scoped**, not pattern-matched. Confirmed in \`browseros-ai/openclaw\` at \`src/config/sessions/combined-store-gateway.ts:90\`:

\`\`\`ts
const targets = requestedAgentId
  ? resolveAgentSessionStoreTargetsSync(cfg, requestedAgentId)  // only this agent's dir
  : resolveAllAgentSessionStoreTargetsSync(cfg);
\`\`\`

Each agent has its own directory. Querying \`research\` cannot return a \`writer\` session.

## Behavior choices

- **Sub-session fetch failures** → log + drop. Partial timeline preferable to hiding the main session entirely.
- **\`limit\`** → applied post-merge across the unified timeline (single global limit, not per-sub-session).
- **Streaming variant** (\`Accept: text/event-stream\`) → unchanged for v1; aggregation only on the JSON path.
- **No new endpoint** → modified the existing \`/claw/session/:key/history\` route; no new query flag.
- **Wire shape** → backward compatible. Existing callers just see \`source\` / \`subSessionKey\` as new optional fields on each message; old clients ignoring them keep working.

## Files

| File | Change |
|---|---|
| \`openclaw-service.ts\` | New \`fetchAggregatedAgentHistory\` branch in \`getSessionHistory\`; helpers \`extractAgentIdFromMainSessionKey\`, \`parseSessionSource\`, \`compareMessageOrder\` |
| \`openclaw-http-client.ts\` | Add optional \`source\` and \`subSessionKey\` fields on \`OpenClawSessionHistoryMessage\` |

## Stats

\`\`\`
openclaw-http-client.ts | 11 ++
openclaw-service.ts     | 147 ++++++++++++++++++-
2 files changed, 156 insertions(+), 2 deletions(-)
\`\`\`

## Validation

- \`bun run typecheck\` clean
- \`bunx biome check\` clean
- \`bun test apps/server/tests/api/services/openclaw/openclaw-service.test.ts apps/server/tests/api/routes/openclaw.test.ts\` — **44 pass / 0 fail**

## Test plan

- [ ] Schedule cron jobs in an OpenClaw agent (with delivery: announce or sessionTarget targeting main)
- [ ] Wait for cron runs to fire
- [ ] Reload the chat panel
- [ ] Confirm: cron-spawned messages appear inline with main session messages, ordered by timestamp, tagged with \`source: 'cron'\`
- [ ] Confirm: a single sub-session fetch failure (e.g. corrupt JSONL) doesn't break the response — main session still loads
- [ ] Confirm: \`getSessionHistory\` for a non-main key (e.g. \`agent:X:cron:...\`) still returns just that session

## Notes

Independent of the cleanup PRs (#934, #935, #936). Branched off \`dev\` directly. No conflicts expected.

Reference design: design doc at https://gist.github.com/shivammittal274/b7c200c8845096e750ac3d060c1f3f69

### DIFF
--- a/packages/browseros-agent/apps/server/src/api/server.ts
+++ b/packages/browseros-agent/apps/server/src/api/server.ts
@@ -46,6 +46,7 @@ import {
   connectKlavisInBackground,
   type KlavisProxyRef,
 } from './services/klavis/strata-proxy'
+import { convertOpenClawHistoryToAgentHistory } from './services/openclaw/history-mapper'
 import { OpenClawGatewayChatClient } from './services/openclaw/openclaw-gateway-chat-client'
 import { getOpenClawService } from './services/openclaw/openclaw-service'
 import type { Env, HttpServerConfig } from './types'
@@ -159,6 +160,15 @@ export async function createHttpServer(config: HttpServerConfig) {
             }))
           },
           getStatus: () => getOpenClawService().getStatus(),
+          getAgentHistory: async (agentId) => {
+            // Aggregated across the agent's main + every sub-session
+            // (cron / hook / channel) so autonomous turns surface in
+            // the chat panel alongside user-initiated ones.
+            const raw = await getOpenClawService().getSessionHistory(
+              `agent:${agentId}:main`,
+            )
+            return convertOpenClawHistoryToAgentHistory(agentId, raw)
+          },
         },
       }),
     )
--- a/packages/browseros-agent/apps/server/src/api/services/agents/agent-harness-service.ts
+++ b/packages/browseros-agent/apps/server/src/api/services/agents/agent-harness-service.ts
@@ -120,6 +120,15 @@ export interface OpenClawProvisioner {
    * gateway is not configured at all).
    */
   getStatus?(): Promise<GatewayStatusSnapshot | null>
+  /**
+   * Optional. When wired, the harness uses this for `getHistory` on
+   * openclaw-adapter agents so the chat panel sees autonomous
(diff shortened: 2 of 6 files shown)


shivammittal274:
@greptile-ai review

PR #940, opened 2026-05-05
fix(patch): handle canonical workspace paths

shadowfax92:
**Summary**
- Match registered Chromium workspaces by both the logical cwd and canonical realpath so symlinked checkouts resolve correctly.
- Store canonical workspace paths on add and include cwd, resolved cwd, registered workspaces, and an example command when detection fails.
- Install `browseros-patch` into the user Go bin by default and add `make uninstall` for cleanup.

**Design**
The workspace resolver now canonicalizes both the current directory and registered workspace paths before matching, while preserving the existing longest-prefix behavior for nested workspaces. The registry stores canonical paths at add time so future commands compare stable values. The Makefile keeps `PREFIX` override support but defaults to `go env GOBIN` or `GOPATH/bin`, matching the local Go CLI install pattern.

**Test plan**
- `go test -count=1 ./...` from `packages/browseros/tools/patch`
- `git diff --check`
- `make install PREFIX=/tmp/browseros-patch-install-test`
- `make uninstall PREFIX=/tmp/browseros-patch-install-test`

### DIFF
--- a/packages/browseros/tools/patch/Makefile
+++ b/packages/browseros/tools/patch/Makefile
@@ -1,8 +1,12 @@
 BINARY := browseros-patch
-PREFIX ?= /usr/local/bin
+GOBIN := $(shell go env GOBIN)
+ifeq ($(GOBIN),)
+GOBIN := $(shell go env GOPATH)/bin
+endif
+PREFIX ?= $(GOBIN)
 VERSION ?= dev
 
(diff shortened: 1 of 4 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #941, opened 2026-05-05
fix: improve browseros-patch CLI ergonomics

shadowfax92:
## Summary
- Clarifies browseros-patch around patch repos and named Chromium checkouts.
- Makes checkout detection errors actionable with cwd, resolved cwd, registered checkouts, and named command examples.
- Adds concise command examples and `--llm-txt` guidance for coding agents.

## Design
This keeps the existing registry and patch engine behavior intact. User-facing copy now says Chromium checkout, named checkout commands work from anywhere through the existing resolver, and `--llm-txt` prints a static plain-text quick reference before app state is loaded.

## Test plan
- [x] `go test -count=1 ./...` from `packages/browseros/tools/patch`
- [x] `go vet ./...` from `packages/browseros/tools/patch`
- [x] Inspected `go run . --help`, `diff --help`, `status --help`, `apply --help`, `sync --help`, `extract --help`
- [x] Inspected `go run . --llm-txt`

### DIFF
--- a/packages/browseros/tools/patch/cmd/add.go
+++ b/packages/browseros/tools/patch/cmd/add.go
@@ -12,8 +12,8 @@ func init() {
 	command := &cobra.Command{
 		Use:         "add <name> <path>",
 		Aliases:     []string{"register"},
-		Annotations: map[string]string{"group": "Workspace:"},
-		Short:       "Register a Chromium checkout as a workspace",
+		Annotations: map[string]string{"group": "Chromium Checkouts:"},
+		Short:       "Register a named Chromium checkout",
 		Args:        cobra.ExactArgs(2),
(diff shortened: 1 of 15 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

shadowfax92:
Review feedback processed. Accepted: moved `--llm-txt` to a root-only local flag so it no longer appears in subcommand help; removed the dead empty-registry guard in `namedCheckoutExample`. Declined: JSON keys stay `workspace`/`workspaces` for compatibility; conflict commands still require active-conflict discovery because they do not accept checkout names in this PR; stdout capture remains because the current human list output writes through `fmt.Println`. Verified with `go test -count=1 ./...`, `go vet ./...`, and manual help checks.

PR #964, opened 2026-05-08
feat(runtime): introduce AgentRuntime abstraction (types, interface, registry, abstract bases)

DaniAkash:
## Summary

Introduces a unified `AgentRuntime` layer that sits above the existing container abstraction and parallels it for host-process adapters (claude, codex). No adapter migrates yet — this is foundation only.

- `AgentRuntime` interface — descriptor, status snapshot, capability list, action dispatcher, exec-argv builder, per-agent home dir, optional path translation
- Discriminated `RuntimeAction` union (variant-required args like `agentId` for `reset-wipe-agent`)
- `RuntimeState` superset covering both container states (`not_installed`/`installing`/.../`running`/`stopped`/`errored`) and host-CLI states (`cli_missing`/`cli_present`/`cli_unhealthy`)
- `ContainerAgentRuntime` — abstract base extending `ManagedContainer`, default lifecycle + reset capabilities, dispatcher that gates on `getCapabilities()` and routes to inherited methods
- `HostProcessAgentRuntime` — abstract base for host-PATH binaries with cached `--version` probe, `cli_missing → cli_present/cli_unhealthy` transitions, env-prefixed exec argv, subclass hooks for `checkAuth` + `handleReinstallCli`
- `AgentRuntimeRegistry` — duplicate-guarded register/get/list/unregister + process singleton with test reset

## Test plan

- [x] `bun run typecheck` clean across server package
- [x] `biome check` clean on new sources + tests
- [x] `bun test tests/lib/agents/runtime/` — 29 cases covering capability defaults + filtering, status snapshot mapping, subscribe wiring, action dispatch (incl. `ActionNotSupportedError` for cross-kind actions), probe success/non-zero/spawn-error transitions, cache + force-bypass, registry round-trip + duplicate guard + singleton reset
- [x] Full server test sweep (`bun test`) — 1010 pass, 0 fail

### DIFF
--- a/packages/browseros-agent/apps/server/src/lib/agents/runtime/agent-runtime.ts
+++ b/packages/browseros-agent/apps/server/src/lib/agents/runtime/agent-runtime.ts
@@ -1,0 +1,50 @@
+/**
+ * @license
+ * Copyright 2025 BrowserOS
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ *
+ * Top-level interface every adapter runtime implements. Two abstract
+ * subclasses (`ContainerAgentRuntime`, `HostProcessAgentRuntime`)
+ * cover the two kinds we ship today.
+ */
+
+import type {
+  ExecSpec,
+  RuntimeAction,
+  RuntimeCapability,
+  RuntimeDescriptor,
+  RuntimeStatusSnapshot,
+  StateListener,
+  Unsubscribe,
+} from './types'
+
+export interface AgentRuntime {
+  readonly descriptor: RuntimeDescriptor
+
+  // ── Status surface (Plane B feed) ────────────────────────────────
+  getStatusSnapshot(): RuntimeStatusSnapshot
+  subscribe(listener: StateListener): Unsubscribe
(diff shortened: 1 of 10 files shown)


DaniAkash:
Greptile #1 (unquoted shell values): non-blocking — matches the existing `ManagedContainer.buildExecArgv` precedent, which also returns `argv.join(' ')` and leaves quoting to the downstream consumer (the array form is used for actual `Bun.spawn`). Will revisit alongside the existing pattern in a Hermes follow-up.
Greptile #2 (capability gate): fixed in 849c5ea — `executeAction` now checks `getCapabilities()` before dispatch, matching the container base.
Greptile #3 (probe cache stamp): fixed in 849c5ea — `healthCheckedAt` is now stamped only after the probe resolves, so spawn errors don't lock out re-probes.
Greptile #4 (`logs` capability with no action): non-blocking — intentional. `RuntimeCapability` is the UI-affordance set (gates the "view logs" button); `RuntimeAction` is the executable subset. Logs/terminal stream through their own endpoints rather than `executeAction`.

PR #983, opened 2026-05-09
feat(agent): add silent scheduled task runs

### LINKED ISSUE #447: [Feature Request] Option for Scheduled Tasks to Run Silently (No New Window Popup)
**What problem does this solve? (describe the problem you're experiencing)**
* When a scheduled task is triggered in BrowserOS, it currently causes a new window to pop up every time. This can be distracting and disruptive, especially when running frequent or background tasks, and negatively impacts the user experience.

**How are you working around this today? (your current solution or workaround)**
* Currently, I have to manually minimize each pop-up window or try to avoid using scheduled tasks that create interruptions. There does not seem to be a setting to keep scheduled tasks from showing new windows.

**What's your proposed solution? (how should BrowserOS address this?)**
* I suggest adding an option or setting for scheduled tasks to run in a "silent" or "quiet" mode. This would allow scheduled tasks to run without opening a new window, perhaps by showing only a background notification or status indicator instead. The user should be able to choose whether tasks pop up a window or run quietly in the background.

**Additional context (optional - add screenshots, examples, or other helpful details)**
* This feature would help users who rely on background automations and reduce workflow interruptions for those who do not need to interact with every scheduled task window.

---
Thanks for considering this suggestion!

shadowfax92:
Fixes #447

Verification:
- Passed: bun test apps/agent/lib/messaging/server/buildChatRequestBody.test.ts
- Passed: bun run test:cleanup && bun --env-file=.env.development test ./tests/agent/prompt.test.ts ./tests/api/services/chat-service.test.ts ./tests/browser/browser.test.ts
- Passed: bunx biome check on changed files
- Passed: git diff --check origin/dev...HEAD
- Local blocked: server bun run typecheck cannot find tsc; tracked as bosmain-woi.
- Local blocked: agent bun run typecheck cannot find wxt; tracked as bosmain-090.

### DIFF
--- a/packages/browseros-agent/apps/agent/entrypoints/app/scheduled-tasks/NewScheduledTaskDialog.tsx
+++ b/packages/browseros-agent/apps/agent/entrypoints/app/scheduled-tasks/NewScheduledTaskDialog.tsx
@@ -57,6 +57,7 @@ const formSchema = z
     scheduleTime: z.string().optional(),
     scheduleInterval: z.number().int().min(1).max(60).optional(),
     providerId: z.string().optional(),
+    runSilently: z.boolean(),
     enabled: z.boolean(),
   })
   .superRefine((data, ctx) => {
@@ -107,6 +108,7 @@ export const NewScheduledTaskDialog: FC<NewScheduledTaskDialogProps> = ({
       scheduleTime: '09:00',
       scheduleInterval: 1,
       providerId: undefined,
+      runSilently: true,
       enabled: true,
     },
   })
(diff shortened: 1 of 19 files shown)


shadowfax92:
Refinery rejected this merge request after CI gate. GitHub Tests / server-browser failed because the new tests/browser/browser.test.ts mocks @browseros/shared/constants/limits with only CONTENT_LIMITS, causing tests/browser/backends/cdp.test.ts to fail importing CDP_LIMITS from the mocked module. Source issue bosmain-445 has been reopened for rework; this branch is not merged.

PR #1004, opened 2026-05-15
feat(mcp): X-BrowserOS-Default-Window-Id header + set_window_visibility tool

shivammittal274:
## Summary

Two additions for host applications that want to deterministically bind every browser tool call in an MCP session to a specific window.

- **Header `X-BrowserOS-Default-Window-Id`** — honored in `register-mcp.ts`. Before `executeTool` runs, the handler wrap reads the tool's zod input schema; if it has a `windowId` field and `args.windowId` is undefined, the header value is injected. Schema-driven — covers `new_page`, `new_hidden_page`, `show_page`, `move_page` today and any future tool with a `windowId` field with no per-tool allowlist. The agent's explicit `args.windowId` still wins when set (intentional, lets multi-window batch work continue).
- **`set_window_visibility {windowId, visible}` tool** — symmetric show/hide on an existing window, delegating to `cdp.Browser.showWindow` / `hideWindow`. Idempotent. Preserves the windowId and tabs.

The C++ side of `Browser.showWindow` / `hideWindow` isn't wired in the binary yet (`is_hidden_` is `const` and there's no handler in `BrowserHandler` for them; only declared in the `inspector_protocol_config.json` include list + generated client types). The tool returns the runtime error verbatim so callers can degrade gracefully; once the Chromium patch lands, the tool starts working with no changes here.

## Companion PR

[agent-company#TBD](https://github.com/browseros-ai/agent-company) uses this header to per-`(employee, thread)` bind windows and exposes a visibility toggle in the chat header.

## Test plan

- [x] `bun run dev:watch:new` boots; `set_window_visibility` appears in `tools/list` (61 tools registered).
- [x] `tools/call new_page` with `X-BrowserOS-Default-Window-Id: <id>` header and no `windowId` in args → page lands in `<id>`. Verified via follow-up `list_pages`.
- [x] `tools/call set_window_visibility` returns the expected runtime error from missing CDP wire-up — no crash, error text propagated to client.
- [x] Agent's explicit `args.windowId` is **not** overridden by the header.
- [x] Tools without `windowId` in their schema are untouched by the injection.

### DIFF
--- a/packages/browseros-agent/apps/server/src/api/routes/mcp.ts
+++ b/packages/browseros-agent/apps/server/src/api/routes/mcp.ts
@@ -28,6 +28,12 @@ interface McpRouteDeps {
   klavisRef?: KlavisProxyRef
 }
 
+function parseOptionalNumber(value: string | undefined): number | undefined {
+  if (value === undefined) return undefined
+  const n = Number(value)
+  return Number.isFinite(n) ? n : undefined
+}
+
 export function createMcpRoutes(deps: McpRouteDeps) {
   const app = new Hono<Env>()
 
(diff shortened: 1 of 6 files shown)


shivammittal274:
@greptile review

PR #1060, opened 2026-05-27
feat(server): /screencast WebSocket route + ScreencastManager

shivammittal274:
## Summary
- New `/screencast` WebSocket route on the agent server. Streams the agent's browser as base64 JPEG frames over `Page.startScreencast`, so callers can render a live canvas without popping the native window.
- Subscribers are ack-gated and reference-counted per CDP target: first viewer triggers `Page.startScreencast`, last triggers `stopScreencast`. A slow consumer can't queue frames unboundedly.
- `?pageId=N` follows a specific Browser-internal page; omitted, it falls back to the window's active tab. Sessions are keyed off `targetId`, so two viewers of the same tab share one CDP stream regardless of how each requested it.
- Late subscribers get the cached `lastFrame` immediately; if none exists (idle page that hasn't repainted), `Page.captureScreenshot` primes the canvas synthetically so the pane never sits black.
- Knobs (q=80, 1280×800, every-nth-frame) live in `@browseros/shared/constants/limits` as `SCREENCAST_LIMITS`.

## How it fits
agent-company mounts this WebSocket via an SSE proxy (paired PR https://github.com/browseros-ai/agent-company/pull/76) and renders frames into a canvas in the chat surface. The agent-side renderer lifts \`pageId\` off the latest browseros tool's input/output and threads it here, so the canvas follows whichever tab the agent is acting on.

## API
- \`ws /screencast?windowId=N\` — stream the window's active tab.
- \`ws /screencast?windowId=N&pageId=M\` — stream a specific Browser-internal pageId; falls back to active tab if the page can't be resolved.

## Test plan
- [x] \`bun --env-file=.env.development test apps/server/tests/api/screencast-manager.test.ts\` — integration test against live CDP: subscribe → frame within ~8s on a hidden window, second subscriber gets status + cached lastFrame, unsubscribe cleans up.
- [x] \`bun run lint\` and \`bun run typecheck\` clean for \`apps/server\`.
- [x] Manual: opening, closing, and reopening the agent-company pane against the same window — frames resume on each subscribe (no black canvas on idle pages).
- [x] Manual: agent calls \`new_page google.com\` after the pane is already streaming example.com — pane follows to google.com.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1074, opened 2026-05-30
feat(server): single-window tab-group routing + simplified ScreencastManager

shivammittal274:
## Summary

Two commits enabling the agent-company \"one window, N agents, N tab groups\" host model:

- **`feat(server): per-agent tab group routing for page-creating tools`** — `new_page`, `new_hidden_page`, `show_page`, `move_page` now accept an optional `tabGroupId` and call `groupTabs` after page creation. Pairs with `X-BrowserOS-Default-Tab-Group-Id` so hosts can pin every page-creating tool call from a given agent into a specific tab group without the agent picking groups itself.
- **`refactor(server): simplify ScreencastManager to a singleton + bringToFront wake-up`** — drops the sessions map / pendingStarts / primeWithScreenshot / lastFrame / retry-loop complexity; one `active` session at a time, `tearDown` displaces the prior subscriber, and `Page.bringToFront` is called right before `Page.startScreencast` so backgrounded tabs actually composite. Subscribe also catches `Unknown page` and emits a clean `detached` status instead of letting the route close with 1011 (which spirals EventSource reconnects).

## Why bringToFront

`Page.startScreencast` on a tab whose RenderWidget isn't visible returns zero frames — the compositor is paused. POC confirmed `setWebLifecycleState('active')` alone is not sufficient; `bringToFront` is the lightest wake-up that delivers frames. Side-effect: BrowserOS visibly switches its frontmost tab and steals macOS focus, so the `connected` status is emitted right after `bringToFront` to give the agent-company SSE proxy an early signal to counter-activate Electron.

## Test plan

- [x] `bun run typecheck` clean
- [x] WS subscribe to a backgrounded tab → frames flow after bringToFront (was 0 before)
- [x] WS subscribe to a non-existent pageId → single `detached` status, no spiral
- [x] Cross-target subscribe sequence (A → B → A) → singleton displacement, no cross-talk
- [x] Tested end-to-end with agent-company's `feat/single-window-and-screencast` PR

## Related

Companion PR: [browseros-ai/agent-company#104](https://github.com/browseros-ai/agent-company/pull/104) — single app window + per-pane screencast + visibility toggle that consume this server-side foundation.

### DIFF
--- a/packages/browseros-agent/apps/server/src/api/routes/mcp.ts
+++ b/packages/browseros-agent/apps/server/src/api/routes/mcp.ts
@@ -68,12 +68,19 @@ export function createMcpRoutes(deps: McpRouteDeps) {
       c.req.header('X-BrowserOS-Default-Window-Id'),
     )
 
+    // Same pattern for tab groups — the host pins every page-creating
+    // call to a specific tab group so concurrent agents don't race for
+    // the window's active group.
+    const defaultTabGroupId =
+      c.req.header('X-BrowserOS-Default-Tab-Group-Id') ?? undefined
+
     // Per-request server + transport: no shared state, no race conditions,
     // no ID collisions. Required by MCP SDK 1.26.0+ security fix (GHSA-345p-7cg4-v4c7).
     const mcpServer = createMcpServer({
       ...deps,
       observer,
       defaultWindowId,
+      defaultTabGroupId,
     })
     const transport = new StreamableHTTPTransport({
       sessionIdGenerator: undefined,
--- a/packages/browseros-agent/apps/server/src/api/services/mcp/mcp-server.ts
+++ b/packages/browseros-agent/apps/server/src/api/services/mcp/mcp-server.ts
@@ -31,6 +31,11 @@ export interface McpServiceDeps {
   // bind every browser tool call to a specific window without the
   // agent needing to be aware of it.
   defaultWindowId?: number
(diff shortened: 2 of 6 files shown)


shivammittal274:
@greptile review

shivammittal274:
**Note on `Tests / server-api` failure**

The failing test is `ScreencastManager > subscribes, emits frames, displaces a prior subscriber, and stops on unsubscribe`, but the failure mode is `CDP failed to start on port XXXX within timeout` in `withBrowser`'s spawn step — i.e. the test code never runs its assertions. The BrowserOS binary fails to bring up CDP within the 15s wait in this CI environment.

This is **pre-existing** — PR #1060 (which originally added this test file) had the exact same `CDP failed to start` failure on `Tests / server-api` when it was merged ([job 78501091904](https://github.com/browseros-ai/BrowserOS/actions/runs/26637397520/job/78501091904)). My commit `5dc8626e` updates the test to match the new singleton manager's contract (visible window + displacement-emits-detached); the test passes locally:

```
tests/api/screencast-manager.test.ts:
Starting BrowserOS on CDP port 9005...
CDP is ready
 1 pass
 0 fail
```

I'll leave the test fix in. If the reviewer wants the CI environment issue addressed in this PR, happy to dig — but it looks like a runner/binary problem orthogonal to the manager rewrite.

PR #1080, opened 2026-05-31
fix(server): drop CODEGEN_SERVICE_URL and source SENTRY_DSN from env file

shadowfax92:
## Summary
- The `release-server` workflow's `build:server` step was failing CI — `validateProductionEnv()` requires `CODEGEN_SERVICE_URL` and `SENTRY_DSN`, but neither is configured as a CI secret, so they resolved to empty strings and the build aborted (`Production build requires variables: CODEGEN_SERVICE_URL, SENTRY_DSN`).
- **Removes `CODEGEN_SERVICE_URL` end to end** — it was inlined into server config and threaded into `HttpServerConfig`, but never actually read; the `EXTERNAL_URLS.CODEGEN_SERVICE` constant also had zero consumers (dead code).
- **Sources `SENTRY_DSN` from the committed env template** instead of an (unconfigured) CI secret.

## Design
A Sentry DSN is a public client identifier — it's already inlined into every shipped binary, exactly like `BROWSEROS_CONFIG_URL`, which ships its real value in `.env.production.example`. So the DSN now lives in that template, which CI copies to `.env.production` before building. The subtle part: `buildInlineEnv` resolves values as `process.env[key] ?? fileEnv[key]`, and `'' ?? x === ''`, so an empty `SENTRY_DSN` injected by the workflow `env:` block would shadow the file value. The fix therefore also drops the now-empty `SENTRY_DSN` and `CODEGEN_SERVICE_URL` lines from the workflow so the file value actually takes effect. `POSTHOG_API_KEY` is untouched — its secret is configured and still flows from the workflow env.

## Test plan
- `bun run typecheck` — all packages pass
- `bunx biome check` on changed files — clean (exit 0)
- `bun test apps/server/tests/build.test.ts` — 2 pass (real compile + `--version`, plus `--ci` archive build)
- `go build ./...` in `tools/dogfood` — passes
- Reproduced the original failure against the real `loadBuildConfig` before the change; confirmed the fixed scenario (SENTRY from file, no CODEGEN anywhere) loads cleanly after.


shadowfax92:
@greptileai review

shadowfax92:
@claude review

shadowfax92:
Thanks @greptileai and @claude — addressed below.

### ✅ dogfood `DefaultProductionEnv` → `SENTRY_DSN` (Greptile P1) — done in `7244a05f`
Seeded the public DSN so dogfood's server prod-env map stays aligned with `.env.production.example`.

One correction on the framing, though: **this isn't a regression introduced by the PR.** `SENTRY_DSN: ""` predates it — the PR only removed the `CODEGEN_SERVICE_URL` line. And dogfood's pipeline (`tools/dogfood/pipeline/build.go`) builds the *agent* in development mode; it never invokes the validating `build:server`, so `validateProductionEnv` is never reached from dogfood. I still made the change because it's the right **consistency** call: `SENTRY_DSN` is a non-secret client identifier (the whole thesis of this PR), so it belongs with the other hardcoded non-secret default (`BROWSEROS_CONFIG_URL`). `POSTHOG_API_KEY` and `R2_*` stay empty — those are genuine secrets supplied via user config.

### ❌ README `SENTRY_DSN` "Default" column — declining
The "Default" column lists the **code-level fallback**, not the template value — only `R2_UPLOAD_PREFIX` has one (`'server/prod-resources'` in `config.ts`). `BROWSEROS_CONFIG_URL` also ships a committed value in `.env.production.example` yet correctly shows `-`, because there's no code default; the value is sourced from env at build time. `SENTRY_DSN` is in the identical situation, so `-` is consistent — pointing only its row at the template would make the table asymmetric.

### ℹ️ @claude observations
- **Public DSN rate-limiting / allowed origins** — good call; that's a Sentry-project config concern (inbound limits), no code change here.
- **`POSTHOG_API_KEY` empty in the template** — intentional. It's a genuine secret and flows from the CI secret, so a local production build does require it in the environment.

Verification for the dogfood change: `go build ./...` and `go test ./config/...` pass.

shadowfax92:
Correction on the dogfood item: reverted in `324490d5` — **declining after all.**

No env value should be hardcoded into the Go source here. As noted above, dogfood never runs the validating `build:server` (it builds the agent in dev mode), so `DefaultProductionEnv` doesn't need the DSN. Keeping `SENTRY_DSN: ""` alongside the other empty placeholders (`POSTHOG_API_KEY`, `R2_*`) is correct — values are supplied via user config, not baked into source.

Net: both Greptile suggestions declined; the core CI fix (commit `863521e2`) is unchanged.

PR #1083, opened 2026-06-01
fix(agent): remove magic-link sign-in, keep Google-only

shadowfax92:
## Summary

- Remove magic-link sign-in from the **onboarding** flow and the standalone **`/login`** page, leaving **Continue with Google** as the only sign-in option on both.
- Delete the now-dead `/auth/magic-link` callback route (`MagicLinkCallback.tsx`) and the better-auth `magicLinkClient()` client plugin.

## Why

Magic-link sign-in emails the user a link they have to paste back into BrowserOS to finish authenticating. In practice users paste it into whatever browser happens to be default/open — often **not** BrowserOS — so the session never completes and the flow dead-ends. Google OAuth redirects in-browser and avoids the cross-browser handoff entirely, so it's the more reliable path.

## Design

Both sign-in surfaces (`onboarding/steps/StepTwo.tsx` and `app/login/LoginPage.tsx`) shared the same structure: a Google button, a separator, an email→magic-link form, and a "check your email" success branch. The email form, separator, and success branch were removed from both, along with their `magic-link-sent` state and `handleMagicLink` handlers. Google sign-in (`signIn.social({ provider: 'google' })`) and the onboarding "Skip for now" affordance are unchanged. With nothing left to send a link, the `/auth/magic-link` callback component/route and the `magicLinkClient()` plugin in `auth-client.ts` are removed as dead code.

## Test plan

- Onboarding sign-in step renders only **Continue with Google** + **Skip for now**.
- `/login` renders only **Continue with Google** (still reachable from the new-tab "Sign in" hint).
- Google sign-in completes and lands on `/home`; onboarding variant routes through `/onboarding/demo`.
- `bun run typecheck` → exit 0 · `bunx biome check` → clean · no remaining `magic-link` references in source.

## Notes / follow-up

- The **server-side** better-auth config (separate `apps/server` package) still has the magic-link plugin + email sending enabled. It's now unreachable from the agent UI, so harmless — but if we want it gone end-to-end, that's a small server-side follow-up.
- Any magic links already sitting in inboxes will now hit the `*` fallback route → redirect to `/home` (expected, given the callback route is removed).


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1091, opened 2026-06-02
fix: hide Hub button by default

shadowfax92:
## Summary

- Hide the Hub toolbar button by default when `browseros.show_llm_hub` has no explicit user value.
- Preserve existing explicit user choices for the Hub visibility pref.
- Align the BrowserOS customization UI switch with the new off-by-default behavior.

## Design

Keep the existing `browseros.show_llm_hub` pref key, change the Chromium registered default to `false`, and update the agent settings fallback so missing pref values render as off. This avoids a migration and keeps explicit stored true/false values authoritative.

## Test plan

- `bun run --filter @browseros/agent lint`
- `bun run --filter @browseros/agent typecheck`
- `bun run --filter @browseros/agent test`
- `git diff --check`
- `autoninja -C out/Default_arm64 chrome` attempted; unavailable in this Grove worktree because `out/Default_arm64` does not exist.


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1095, opened 2026-06-02
fix(server): hide browser-context scaffolding in the new-tab agent chat

shadowfax92:
## Summary
- The new-tab **Agents** chat leaked the `## Browser Context … <USER_QUERY>…</USER_QUERY>` prompt scaffolding in the user-message bubble, while the side panel showed only the clean query (see before/after below).
- **Root cause:** the new-tab view renders the active turn's `prompt` (from `GET /chat/active`). For side-panel-initiated turns that stored prompt is the wrapped `formatUserMessage` output, and `getActiveTurn` returned it verbatim — the one read path that didn't strip.
- **Fix:** `getActiveTurn` now unwraps the prompt on read via the canonical `unwrapBrowserosAcpUserMessage`, the same read-time strip `getHistory` and `extractLastUserMessage` already apply.

## Design
Read-time normalization in `AgentHarnessService.getActiveTurn`, mirroring how `getHistory` de-scaffolds on read. The generic `TurnRegistry` keeps storing the faithful prompt (no layering violation), and the canonical unwrap helper is reused with zero duplicated logic. The change touches only the `/chat/active` read boundary — never the prompt sent to the LLM, and not the already-correct history/listing paths. A `null` prompt passes through unchanged, and the helper is idempotent so already-clean (new-tab-initiated) prompts are untouched.

## Test plan
- New unit tests in `agent-harness-service.test.ts`:
  - a wrapped `## Browser Context … <USER_QUERY>` prompt surfaces via `getActiveTurn` as the clean inner query;
  - a `null` prompt passes through;
  - an already-clean prompt is unchanged (idempotence).
- `bun test` across service + acpx-runtime + agents-route suites: **66/66 pass**.
- `bunx biome check` clean; `tsc --noEmit` exit 0.
- Manual: start a turn from the side panel, then open the same agent in the new-tab Agents page — the user bubble now shows only the typed query, matching the side panel.

## Out of scope
- Prompts prefixed with a non-empty `userSystemPrompt` (pre-existing limitation shared with `getHistory`; the `^## Browser Context` anchor won't match behind a prefix).
- Whether new-tab-initiated turns receive browser context (separate product concern).

### DIFF
--- a/packages/browseros-agent/apps/server/src/api/services/agents/agent-harness-service.ts
+++ b/packages/browseros-agent/apps/server/src/api/services/agents/agent-harness-service.ts
@@ -4,7 +4,10 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 
-import { AcpxRuntime } from '../../../lib/agents/acpx/runtime'
+import {
+  AcpxRuntime,
+  unwrapBrowserosAcpUserMessage,
+} from '../../../lib/agents/acpx/runtime'
 import type { AgentDefinition } from '../../../lib/agents/agent-types'
 import {
   getHermesHarnessHostDir,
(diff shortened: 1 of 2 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1098, opened 2026-06-02
fix: daily scheduled task fires on startup if created after its scheduleTime (#1018)

### LINKED ISSUE #1018: Daily scheduled task fires on browser startup if created after its scheduleTime
## Summary

A brand-new daily scheduled task fires on browser startup if the browser is restarted after its scheduled time on the day the task was created. The daily branch of `runMissedJobs` in `scheduledJobRuns.ts` doesn't gate on `createdAt`, while the hourly/minutes branch right below it does.

## Steps to reproduce

1. Create a daily task at 17:30 today with `scheduleTime: "09:00"`. `createAlarmFromJob` schedules the first alarm for tomorrow 09:00 (correct).
2. Restart the browser later that same day (or trigger `chrome.runtime.onStartup`).
3. `runMissedJobs` runs.

## Expected

The task doesn't execute. The first run should be the legitimate alarm at tomorrow 09:00.

## Actual

The task runs immediately. The daily check at lines 176-181 only asks "is 09:00 today in the past?" - for a job created at 17:30 today that's true, so the gate falls through to `executeScheduledJob`. The hourly/minutes branch right below (lines 183-193) gates on `createdAt`; the daily branch was just missing the same guard.

## Repro (literal Node output)

```
$ node -e '
function dailyGate(scheduleTime, now) {
    const [hours, minutes] = scheduleTime.split(":").map(Number);
    const scheduledToday = new Date(now);
    scheduledToday.setHours(hours, minutes, 0, 0);
    return now < scheduledToday.getTime();
}
const now = new Date("2026-05-17T18:00:00+02:00").getTime();
console.log("Gate returns skip=" + dailyGate("09:00", now));
'
Gate returns skip=false
```

`skip=false` means the daily check does not protect the job; execution continues to `executeScheduledJob`.

## Secondary case: malformed `scheduleTime`

Strings that don't parse as HH:MM (`"abc"`, `""`, `"9"`) produce `[NaN]` from `.split(':').map(Number)`. `setHours(NaN, NaN, 0, 0)` makes the Date invalid, `getTime()` is NaN, `now < NaN` is false, so the daily gate falls through on every alarm tick. This is reachable in practice via `apps/server/src/tools/nudges.ts`, which forwards any LLM-generated string as `scheduleTime` without validation.

## File references

- `packages/browseros-agent/apps/agent/entrypoints/background/scheduledJobRuns.ts:176-181` - daily gate missing the `createdAt` check
- `packages/browseros-agent/apps/agent/entrypoints/background/scheduledJobRuns.ts:183-193` - hourly/minutes branch with the right pattern
- `packages/browseros-agent/apps/server/src/tools/nudges.ts` - path that can produce malformed `scheduleTime`

cenktekin:
This PR fixes Issue #1018 by checking the creation time of daily scheduled tasks. If a task was created today AFTER its scheduled time, it will no longer trigger immediately as a 'missed' job, preventing unwanted execution floods upon task creation or browser startup. Fixes #1018.


cenktekin:
🛡️ Strengthened security guards and improved error reporting in host-process runtime. 🔱

cenktekin:
✅ **DeepScan Fix Applied:** Cleaned up codebase based on automated review. 🔱

PR #1101, opened 2026-06-02
fix(build): update bundled CLI signing metadata

shadowfax92:
**Summary**
- Keep the bundled Codex and Claude Code CLI binaries registered for macOS and Windows signing.
- Remove stale limactl signing metadata now that Lima is unshipped.
- Prevent macOS app signing discovery from signing unregistered BrowserOSServer executables.

**Design**
The shared BrowserOS Server signing table remains the single source of truth for both the Chromium-build signing path and OTA signing. macOS app signing now filters BrowserOSServer executables through that table, so registered binaries such as `codex` and `claude` are signed while stale VM binaries such as `limactl` are ignored.

**Test plan**
- `python -m unittest discover -s build -p "*test.py"`
- `python -m unittest build.modules.sign.macos_test`
- `git diff --check`

### DIFF
--- a/packages/browseros/build/common/server_binaries.py
+++ b/packages/browseros/build/common/server_binaries.py
@@ -32,7 +32,6 @@ MACOS_SERVER_BINARIES: Dict[str, SignSpec] = {
     "codex": SignSpec("codex", "runtime"),
     "claude": SignSpec("claude", "runtime"),
     "rg": SignSpec("rg", "runtime"),
-    "limactl": SignSpec("limactl", "runtime", "lima-vz-entitlements.plist"),
 }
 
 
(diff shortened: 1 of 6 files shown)


shadowfax92:
@claude review

shadowfax92:
@greptileai review

shadowfax92:
Review feedback processed.

Fixed:
- Added `rg` to the macOS BrowserOSServer signing discovery positive case.
- Removed the stale `lima-vz-entitlements.plist` file now that `limactl` is no longer registered for signing.

Declined:
- Not adding `build/modules/sign/__init__.py`. The premise that every other module directory has one is not accurate (`package`, `patches`, `resources`, and `setup` also lack one), and `build.modules.sign` is already explicitly listed in `pyproject.toml`. Adding only `sign/__init__.py` does not improve broad `unittest discover`; adding parent package markers changes discovery import names and breaks existing relative imports. Keeping the current package layout.

PR #1129, opened 2026-06-08
feat(release): auto-bump server version each nightly build

shadowfax92:
## Summary
- The nightly macOS build now **auto-increments the server version (patch++) before building the server**, so each nightly publishes a new, traceable server version to R2 instead of overwriting the current one.
- Mirrors the existing browser-version auto-bump; the server bump is **folded into the same "build vX" version PR** (one PR carries both browser + server versions).
- Follow-up to #1128 (which set the `0.0.98` baseline). The pipeline reads the server version from `apps/server/package.json`, so the bump is the new source of truth each night.

## Design
`release-server.yml` gains an `apply_bump` input (default `false`, so manual dispatch is unchanged). The bump must be **in-place**: a reusable-workflow checkout is pinned to the parent nightly's start SHA, so a mid-run commit would be invisible. So when `apply_bump=true`, the release job runs `bump_server_version.py` to write the next patch into `package.json` + `bun.lock`, builds/uploads **that** version (`latest/` + `<version>/`), and exposes it via `outputs.server_version`.

The `build` job then applies that exact version (`--set`, no recompute) to its local clone and stages `package.json` + `bun.lock` into the existing version-bump PR — so the committed version always equals the uploaded one, even across the two machines. `apply_bump` is gated on `bump_mode != 'none'`, and a guard forbids `apply_bump` + `publish_github_release`.

The bump script is Python (the mac builder runs `uv`/`python3`, not `bun` for scripts) and mirrors `bump_version.py`. One regex anchored on `"name": "@browseros/server"` edits the single version line in both files (asserts exactly one match → minimal diff, no JSON reformatting).

Note: the browser build already pulls `artifacts/server/latest/…` verbatim every nightly, so "latest gets pulled in" already worked — this change adds **unique per-version artifacts + a traceable incrementing version**.

## Test plan
- [x] `bump_server_version_test.py` (7 tests): patch bump, `--set`, multi-digit rollover (`0.0.99→0.0.100`), single-line-only diff, exactly-one-match raise, idempotent set, CLI prints version
- [x] Regex validated against the **real** `package.json` + `bun.lock` (one match each; idempotent `--set` = clean no-op)
- [x] Existing `bump_version_test.py` still passes (no regression)
- [x] `actionlint` clean on both workflows (apart from a pre-existing custom-runner-label false positive)
- [ ] Observe the next scheduled nightly: server version increments, one PR carries both bumps, `artifacts/server/<new>/…` appears

### DIFF
--- a/.github/workflows/nightly-macos-build.yml
+++ b/.github/workflows/nightly-macos-build.yml
@@ -137,6 +137,7 @@ jobs:
     uses: ./.github/workflows/release-server.yml
     with:
       version: ${{ needs.resolve_inputs.outputs.server_version }}
+      apply_bump: ${{ needs.resolve_inputs.outputs.bump_mode != 'none' }}
       targets: darwin-arm64
       publish_github_release: false
     secrets: inherit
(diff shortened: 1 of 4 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1138, opened 2026-06-08
chore(cli): remove unreliable install command

shadowfax92:
## Summary
- Remove the cobra `install` command from `browseros-cli` — it downloaded/installed the BrowserOS **browser** per-platform (macOS `.dmg`, Windows `.exe`, Linux AppImage/`.deb`) but was unreliable, so we're dropping it. Deletes `cmd/install.go` in full, including its file-private helpers and the `--dir`/`--deb` flags.
- Repoint the three in-product "not installed" hints (`cmd/root.go`, `cmd/launch.go`, `mcp/client.go`) and the docs (`apps/cli/README.md`, `apps/cli/npm/README.md`) to **download from https://browseros.com** instead of the now-dead command.
- Leave the **CLI self-installer** untouched (`cdn.browseros.com/cli/install.sh|ps1`, npm `postinstall`) — that installs the CLI binary, a different concern — along with all unrelated `BROWSEROS_INSTALL_METHOD` / `.deb` browser-detection code and `go.mod`.

## Design
Commands self-register via per-file `init()` → `rootCmd.AddCommand`, so deleting `cmd/install.go` fully unwires the command (it drops out of `--help` and `completion` automatically); every import in the file was shared with sibling files, so `go.mod`/`go.sum` are unchanged. The stale hints are **repointed rather than deleted** so the "how do I get BrowserOS" guidance stays useful — the same URL `install.go` itself printed as its manual fallback. The "Setup:" help group survives because `launch`/`init`/`update` still populate it.

## Test plan
- [x] `go build ./...`, `go vet ./...`, `go test ./...` green (incl. `-tags integration` compile + graceful skip)
- [x] built binary `--help` and `completion` no longer list `install`
- [x] repo-wide grep: no `browseros-cli install` or `cmd/install.go` references remain
- [x] CLI self-installer + npm distribution paths untouched


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1143, opened 2026-06-09
feat: spill large snapshots to temp files

shadowfax92:
## Summary
- Write very large active browser snapshots to markdown temp files instead of returning full snapshot text inline.
- Preserve small snapshot behavior and trust-boundary wrapping; saved files contain the full wrapped snapshot.
- Return structured metadata for large snapshots: page, path, contentLength, wordCount, and writtenToFile.

## Design
The active snapshot tool keeps the existing observe/origin/wrap/textResult flow. It branches only when the raw snapshot exceeds 5000 words or a conservative 50000-character fallback, then writes the wrapped content with writeTempToolOutputFile and returns a concise file-path message.

## Test plan
- [x] cd apps/server && bun test tests/tools/browser/register.test.ts
- [x] bun run lint
- [x] bun run typecheck
- [x] bun run test:main

Note: lint exits 0 with existing warnings outside this change.

### DIFF
--- a/packages/browseros-agent/apps/server/src/tools/browser/snapshot.ts
+++ b/packages/browseros-agent/apps/server/src/tools/browser/snapshot.ts
@@ -1,6 +1,14 @@
 import { z } from 'zod'
 import { defineTool, textResult } from './framework'
+import { writeTempToolOutputFile } from './output-file'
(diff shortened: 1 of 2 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1144, opened 2026-06-09
fix: handle browser diff URL changes

shadowfax92:
## Summary
- Track the URL with each browser snapshot observation and return the full current snapshot when the URL changed instead of emitting a misleading line diff.
- Keep URL-change responses inside the existing untrusted page-content wrapper, with generic trusted prose and URL-change metadata.
- Stabilize snapshot capture around URL reads so navigation during capture asks the caller to retry instead of committing an incoherent baseline.

## Design
The observer now owns URL-aware snapshot baselines, and the pure snapshot diff layer compares observations before line-diffing. Tool renderers preserve existing result conventions while handling the URL-change branch explicitly for both `diff` and `act`.

## Test plan
- [x] `bun test apps/server/src/browser/core/snapshot/diff.test.ts apps/server/tests/tools/browser/register.test.ts`
- [x] `bun run lint` (exits 0; existing warnings remain outside this change)
- [x] `bun run typecheck`
- [x] `bun run test:main`

### DIFF
--- a/packages/browseros-agent/apps/server/src/browser/core/observer/observer.ts
+++ b/packages/browseros-agent/apps/server/src/browser/core/observer/observer.ts
@@ -10,6 +10,7 @@ import type { FrameRegistry } from './frames'
 import { type ResolvedElement, resolveRefEntry } from './resolve'
 
 const MAX_FRAME_DEPTH = 5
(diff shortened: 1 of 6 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1146, opened 2026-06-09
fix: use CDP frame tree for observer URL reads

shadowfax92:
## Summary
- Replace observer URL reads from `Runtime.evaluate('location.href')` with `Page.getFrameTree()` main-frame metadata.
- Preserve URL fragments via `frame.urlFragment` and retain registry/`unknown` fallback behavior.
- Add focused observer regression tests under the server browser test group.

## Design
`readCurrentUrl` stays private to the observer and now reads the live main-frame URL from CDP frame metadata, recombining `frame.url` with `frame.urlFragment` before falling back to `PageManager.refresh(pageId)?.url` and then `unknown` during teardown.

## Test plan
- [x] `bun --env-file=apps/server/.env.development test --preload=./apps/server/tests/__helpers__/test-env.ts apps/server/tests/browser/core/observer.test.ts`
- [x] `cd apps/server && bun run test:browser`
- [x] `bun run lint`
- [x] `bun run typecheck`
- [x] `bun run test:main`

### DIFF
--- a/packages/browseros-agent/apps/server/src/browser/core/observer/observer.test.ts
+++ b/packages/browseros-agent/apps/server/src/browser/core/observer/observer.test.ts
@@ -1,0 +1,141 @@
+import { describe, expect, test } from 'bun:test'
+import type { ProtocolApi } from '@browseros/cdp-protocol/protocol-api'
+import type { PageManager, PageSession } from '../pages'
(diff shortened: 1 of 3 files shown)


shadowfax92:
@claude review

shadowfax92:
@greptileai review

PR #1150, opened 2026-06-09
chore: zero out the fallow dead-code report

shadowfax92:
## Summary
- Brings all 8 counted `bun run fallow` categories to zero: 45 unused files, 129 unused exports, 41 unused types, 27 unused deps, 7 dev/optional deps, 4 unresolved imports, 1 unlisted dep, 3 circular deps → all 0. No runtime logic changed — deletions, un-exports, verbatim code moves, package.json/lockfile edits, and fallow config only.
- Deletes 43 provably dead files (old agents-row UI cluster, replaced server/eval modules, orphaned scripts, eslint leftovers) — every deletion backed by a reference sweep over CI workflows, shell/build scripts, docs, and git history.
- Registers the false positives instead of deleting them: 4 live scripts become fallow entries (`inspect-ui`, `patch-windows-exe`, eval `generate-report`/`weekly-report` used by `eval-weekly.yml`), `schema.graphql` ignored as codegen input, and `ignoreDependencies` for deps fallow can't see — Tailwind CSS `@import`/`@plugin` packages, the `pino-pretty` transport string, and `@browseros/cdp-protocol` (fallow misses its subpath-exports imports; it is genuinely used by 19 server files).
- Prunes 30 dependency entries (+`@types` twins) across agent/server/eval/root/build-tools; `eventsource-parser` + `fuse.js` stay in apps/agent where they're imported and only leave apps/server.
- Breaks the 3 import cycles with pure moves: `useConversations` → own module, schedule sync functions → `syncSchedulesToBackend.ts`, `killProcessOnPort` → `tests/__helpers__/kill-port.ts`.

## Design
Six sequential waves (delete → config truth-up → deps → export strip → cycle breaks → verify), re-running `fallow check` after each so later waves work from fresh data. `fallow fix` was used only as a mechanical assist and its output filtered against the authoritative `fallow check` report (fix ignores the config overrides and edits exempt files). One deliberate inline suppression: `persistDefaultProviderId` stays exported for its colocated test, since fallow's graph skips test imports. An adversarial two-round review caught and fixed a real landmine: fallow false-positives `@browseros/cdp-protocol` (subpath imports), and removing it only worked locally via a stale isolated-linker symlink — restored + config-ignored.

## Test plan
- [x] `bun run fallow` — summary line empty for all 8 counted categories (only warn-level private-type-leaks remain: 150, down from 168)
- [x] `bun run lint` — 0 errors, 28 warnings (= baseline)
- [x] `bun run typecheck` — 6/6 workspaces
- [x] `bun run test:main` — 10/10 (spawns real BrowserOS + server)
- [x] `bun run --filter @browseros/eval test` — 91/91
- [x] `bun run --filter @browseros/agent test` — 114/114
- [x] `bun install --frozen-lockfile` consistent; fresh-install resolution of `@browseros/cdp-protocol` verified
- [x] `git diff --check` clean


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1152, opened 2026-06-10
fix(build): stop shipping stale universal app in macOS nightly

shadowfax92:
## Summary
- The nightly macOS DMG has been a re-signed **May 27 universal app** for ~2 weeks: `Context.get_app_path()` preferred `out/Default_universal/BrowserOS.app` whenever it existed, so the arm64-only nightly signed/packaged a stale leftover instead of the app it just built. That's why the Codex/Claude CLIs bundled in cf2a9feb (#1093) never appeared in releases — verified by mounting the June 9 nightly DMG: fat x86_64+arm64 binaries, stale `third_party/lima/`, no codex/claude, Chromium build number dating it to build-offset 146 (May 27). The agent→R2→artifact-zip pipeline is fully correct (the R2 zip contains all binaries, `rwxr-xr-x`).
- Fix: `get_app_path()` now resolves strictly from the context's own `out_dir` (or `_fixed_app_path`). The universal pipeline never relied on the removed auto-detect — `UniversalBuildModule` pins `_fixed_app_path` for every context, and `architecture: universal` configs already derive `out_dir = out/Default_universal`.
- Guard: signing now verifies the bundle's `BrowserOSServer/default/resources` matches the staged tree (`chrome/browser/browseros/server/resources`) — missing files or lost exec bits fail the build **before notarization**; bundle-only extras and `.DS_Store` warn/skip. Wired into both `MacOSSignModule.execute()` (also covers `UniversalBuildModule`, which bypasses `validate()`) and legacy `sign_app()`.

## Design
Two independent changes in `packages/browseros/build/`: (1) delete the universal-first probe in `common/context.py` so app-path resolution is single-sourced from `out_dir`; (2) add `verify_server_resources_bundle()` in `modules/sign/macos.py` comparing paths + executable bits only — deliberately not content/size, since universal bundles hold lipo-fat binaries that legitimately differ from the thin staged tree. The guard converts this entire silent-stale failure class into a loud pre-notarization stop.

## Test plan
- [x] `uv run python -m unittest` over all `packages/browseros` test modules — 109 pass (13 new: stale-universal regression, universal/out_dir + `_fixed_app_path` resolution, guard missing-file/exec-bit/match/skip/extras/junk, module wiring raise, source-rel pinned to `copy_resources.yaml`)
- [x] `ruff` + `pyright` clean on touched files
- [ ] After merge: next nightly DMG contains `Contents/Resources/BrowserOSServer/default/resources/bin/third_party/{codex,claude}` signed as `com.browseros.codex`/`com.browseros.claude`

## Follow-ups (not in this PR)
- Ops: delete the stale `out/Default_universal/` on the builder (inert after this fix; frees disk).
- CI gap (pre-existing): no workflow runs `packages/browseros` Python tests.

### DIFF
--- a/packages/browseros/build/common/context.py
+++ b/packages/browseros/build/common/context.py
@@ -439,26 +439,15 @@ class Context:
     def get_app_path(self) -> Path:
         """Get built app path
 
-        For universal builds, checks if out/Default_universal/BrowserOS.app exists
-        and returns that instead of the architecture-specific path.
-
-        This allows downstream modules (sign, package) to work on the universal
-        binary after UniversalBuildModule has run.
-
-        Note: If _fixed_app_path is set, returns that directly (used by
-        UniversalBuildModule to prevent auto-detection during arch-specific ops).
+        Resolves strictly from this context's own out_dir (or _fixed_app_path
+        when set, as UniversalBuildModule does). Never probes other out dirs:
+        a stale out/Default_universal app must not hijack arch-specific
+        builds' sign/package stages. Universal flows resolve here too, since
+        architecture="universal" already derives out_dir=out/Default_universal.
         """
         # If fixed path is set (for arch-specific operations), use it directly
         if self._fixed_app_path:
             return self._fixed_app_path
-
-        # Check for universal binary first (macOS only)
-        if IS_MACOS():
-            universal_app = join_paths(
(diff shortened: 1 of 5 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1161, opened 2026-06-11
fix(build): cap Windows autoninja parallelism to stop clang-cl OOM

shadowfax92:
## Summary
- Windows official builds were dying mid-run with `LLVM ERROR: out of memory` (clang-cl frontend crash, exception 0xC000001D) — autoninja's default one-job-per-core parallelism exhausts Windows commit memory under official+ThinLTO compile loads (~2-4 GB per clang-cl on mojom-heavy TUs).
- On Windows, all three autoninja call sites (`CompileModule.execute`, `build_target`, packaging's `build_mini_installer`) now pass `-j N` where `N = clamp(total_physical_RAM_GB // 4, 1, logical_cpu_count)`; RAM is read via stdlib `ctypes`/`GlobalMemoryStatusEx` (no new deps).
- `BROWSEROS_NINJA_JOBS=<n>` overrides the heuristic on any platform (invalid values warned and ignored); macOS/Linux behavior without the override is byte-identical to before (no `-j`); probe failure falls back to today's behavior with a warning — the heuristic can never break a build.

## Design
A shared `autoninja_command(out_dir, targets, env)` builder in `compile/standard.py` resolves parallelism through `compute_ninja_jobs()` (override → Windows RAM cap → autoninja default) and assembles the argv for every call site, so future call sites can't drift back to uncapped invocations. The decision and its source are logged for build-log auditability. Capping by physical RAM (not commit limit) is deliberate: pagefile size is admin-managed and volatile, and thrashing into it makes builds crawl anyway. `int()` truncation of reported RAM keeps one job of safety margin on firmware-reserved memory — Windows has no overcommit, and an OOM at step 36k/57k costs hours. Incidentally fixes a latent `if IS_WINDOWS` (missing parens, always-truthy) bug in the packaging module's old inline argv.

## Test plan
- [x] 16 new unit tests (`compile/standard_test.py`, `package/windows_test.py`): override valid/invalid on both platforms, RAM-cap math, CPU/floor clamping, probe-failure fallback, unknown CPU count, and argv routing for all three call sites
- [x] Full build-package suite: 114 tests green; ruff, pyright (repo config), and black clean on touched files
- [ ] On the Windows build host: re-run the official build; expect log line `Ninja parallelism: -j <N> (capped by <RAM> GB RAM / 4 GB per job; ...)` and no clang-cl OOM. If more parallelism is wanted, set `BROWSEROS_NINJA_JOBS`.

### DIFF
--- a/packages/browseros/build/modules/compile/standard.py
+++ b/packages/browseros/build/modules/compile/standard.py
@@ -83,7 +83,7 @@ def compute_ninja_jobs(env: Optional[Mapping[str, str]] = None) -> Optional[int]
     if cpus:
         jobs = min(jobs, cpus)
     log_info(
-        f"Ninja parallelism: -j {jobs} (capped by {total_gb:.0f} GB RAM / "
+        f"Ninja parallelism: -j {jobs} (capped by {int(total_gb)} GB RAM / "
         f"{GB_PER_COMPILE_JOB} GB per job; override with BROWSEROS_NINJA_JOBS)"
     )
     return jobs
--- a/packages/browseros/build/modules/compile/standard.py
+++ b/packages/browseros/build/modules/compile/standard.py
@@ -1,9 +1,12 @@
 #!/usr/bin/env python3
 """Standard single-architecture build module for BrowserOS"""
 
+import os
+import sys
 import tempfile
 import shutil
 from pathlib import Path
+from typing import Mapping, Optional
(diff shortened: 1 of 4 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1165, opened 2026-06-11
feat(patch-cli): churn-free extract + pull/rebase-style sync for browseros-patch

shadowfax92:
## Summary
- **Fixes the extract-churn bug at the root:** patch bytes are now a pure function of content — `git diff` runs with pinned config (`--full-index -U3 --no-ext-diff --no-textconv`, algorithm/prefix/context/quotepath `-c` overrides), commit-mode synthetic ADD patches are byte-identical to git's own output (computed blob SHA, real mode, `\ No newline`, header-only empties, single-line hunk headers), and `WriteRepoPatchSet` plans create/update/unchanged/delete keyed on the existing `signature()` so semantically-unchanged patches (including all 342 legacy abbreviated-index files) are **never rewritten**. `status` clean ⇒ `extract` is a byte-level no-op; the same change extracted from two differently-configured checkouts produces identical bytes.
- **`sync` (alias `pull`) is now a real `git pull --rebase` for patches:** local changes are stashed (tracked by commit SHA, immune to positional shifts), patches applied, then **rebased back by default** via a per-file 3-way merge that works where `git stash pop` refuses (apply-dirtied files). Conflicts leave standard markers (modify/delete and binary divergence are reported, never silently resolved), the stash entry survives, `--no-rebase` parks changes with an honest `pending_stash` record that the next rebase sync restores, and `continue`/`skip` finish the loop — restoring the parked stash only when the pause came from a rebase-mode sync.
- **Junk-free, previewable extraction:** built-in untracked-file ignores (`.llm/`, `.browseros-patch/`, `*.log`, `*.rej`, `*.orig`, `.DS_Store`) + repo-root `.browseros-patchignore` + repeatable `extract --exclude`; `extract --dry-run` previews created/updated/deleted/unchanged without touching disk or state; directory-scoped `extract ch1 -- chrome/dir` now works.
- **Agent-grade surface:** exit code 2 whenever apply/sync/continue/skip pause on a conflict (output rendered first, JSON intact), and `--llm-txt` rewritten as a full playbook (exit codes, status field semantics, daily sync flow, capture flow, Chromium-upgrade loop with continue/skip/abort).
- **Ergonomics:** TTY-aware single-line progress (`[i/N]` scans, plain lines in pipes, silent in `--json`), `status --summary` (orphans by top-level dir), `status --all` (one-line table per registered checkout), pending-stash visibility, plain-words sync-state explanations, in-sync-but-orphans reproduce hint, `make install` ships a `bpatch` symlink.

## Design
Canonical-write + skip-unchanged: the compare layer (`signature()`) already ignored volatile `index` lines, but the write path emitted whatever bytes the local git produced — re-extraction churned files even when `status` said up-to-date. This PR fixes both sides of the funnel (deterministic bytes in, signature-gated writes out) instead of stripping index lines, which would have de-fanged the `--3way` apply strategy the conflict loop depends on. Legacy patches converge to canonical form organically as their content genuinely changes — no 342-file migration commit. The Python `browseros dev` extract/apply path is untouched.

## Test plan
- [x] `go test ./...` (6 packages, `-count=1`), `go vet`, `gofmt` clean — round-trip invariants covered: extract→status-clean→extract no-op; cross-checkout byte-identity under hostile git config; legacy-byte preservation; synthetic-add byte-equality vs real git output; SHA-stash rebase/conflict/abort flows; ignore semantics; dry-run isolation; exit-code classification
- [x] End-to-end smoke with the built binary: add → status → extract --dry-run → extract → idempotent re-extract → apply → llm-txt
- [ ] Real-world: `browseros-patch status ch1 && browseros-patch extract ch1` twice against a Chromium checkout — second run must leave `packages/browseros/chromium_patches` untouched

### DIFF
--- a/packages/browseros/tools/patch/.gitignore
+++ b/packages/browseros/tools/patch/.gitignore
@@ -1,0 +1,2 @@
+browseros-patch
+bpatch
--- a/packages/browseros/tools/patch/Makefile
+++ b/packages/browseros/tools/patch/Makefile
@@ -19,11 +19,12 @@ ifneq ($(shell uname -s),Darwin)
 else
 	codesign --force --sign - $(PREFIX)/$(BINARY)
 endif
-	@echo "Installed $(BINARY) to $(PREFIX)/$(BINARY)"
+	ln -sf $(PREFIX)/$(BINARY) $(PREFIX)/bpatch
+	@echo "Installed $(BINARY) to $(PREFIX)/$(BINARY) (alias: bpatch)"
 
 uninstall:
-	rm -f $(PREFIX)/$(BINARY)
-	@echo "Removed $(PREFIX)/$(BINARY)"
+	rm -f $(PREFIX)/$(BINARY) $(PREFIX)/bpatch
+	@echo "Removed $(PREFIX)/$(BINARY) and $(PREFIX)/bpatch"
 
 test:
 	go test ./...
--- a/packages/browseros/tools/patch/cmd/apply.go
+++ b/packages/browseros/tools/patch/cmd/apply.go
@@ -46,7 +46,7 @@ func init() {
 			if err != nil {
 				return err
 			}
-			return renderResult(result, func() {
+			if err := renderResult(result, func() {
 				fmt.Println(ui.Title(fmt.Sprintf("Applied patches to %s", ws.Name)))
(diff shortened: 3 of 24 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1174, opened 2026-06-11
feat(agent): Remote Hermes provider — Cloudflare-managed Fly VM runtime

shivammittal274:
## Summary

Adds a new `remote-hermes` LLM provider that runs the Hermes agent in a managed Fly VM provisioned via the Cloudflare `agent-control-worker` control plane. The user adds the provider with a single click — no API key, base URL, or model fields — and chat turns proxy through the worker (HTTP + SSE) while the VM dispatches tool calls back to the laptop's local BrowserOS MCP over a single WebSocket held open by `apps/server`. Output streams to the side panel as the same AI SDK UI Message format any other provider produces, so the chat UI is unchanged.

One install = one Fly VM (keyed by `browserosId`). Provider save fires `/remote-hermes/start` to warm the VM; deleting the last `remote-hermes` provider fires `/remote-hermes/destroy` to release the Fly machine + volume. Chat turns auto-poll status until the VM is `running` (180s budget) so the first cold-start chat just shows a progress pill instead of erroring out.

## What changed

**Agent UI (`apps/agent/`)**
- `lib/llm-providers/types.ts`, `providerTemplates.ts`, `providerIcons.tsx` — new `remote-hermes` provider type, `Sparkles` icon, first card in the Settings template grid with the orange "Recommended" treatment.
- `screens/ai-settings/NewProviderDialog.tsx`, `provider-form-schema.ts`, `BrowserOsAiPane.tsx` — add-provider flow needs only a name; delete-provider hook fires `/remote-hermes/destroy` when removing the last entry.
- `modules/chat/chat-session.hooks.ts`, `screens/sidepanel/index/Chat.tsx`, `RemoteHermesBootPill.tsx` — inline boot pill reads transient `data-vm-status` parts and advances through worker progress stages (`pulling_image`, `booting`, `healthchecking`).

**Server (`apps/server/`)**
- `lib/remote-hermes/` (new): env loader, HS256 JWT minting (`jose`), frame parser, `partysocket`-backed WS bridge with mutex / refcount / 30-min idle close, RPC router dispatching `browseros` tool calls to `127.0.0.1:<serverPort>/mcp`, `ProtocolEvent` → AI SDK `UIMessageStream` translator (strips acpx namespace from tool names), and the turn streamer with cold-start poll loop (180s budget against `/vm/status`).
- `api/routes/chat.ts` forks on `provider === 'remote-hermes'` and returns `createUIMessageStreamResponse` so the side panel sees the same wire format.
[...]

PR #1176, opened 2026-06-12
chore(agent): trim mcp-manager dead exports and unused devDeps

DaniAkash:
## Summary

- Drives the `packages/browseros-agent` fallow report from **17 issues to 4** (`bun fallow` on the prior main was failing because of dead-code findings introduced by #1168).
- Slims the `mcp-manager` public surface, narrows two internal helpers, exports one props interface to fix a real private-type leak, removes two genuinely unused devDeps (`dotenv`, `picocolors`).
- The remaining 4 findings (`RemoteHermesVmStatus`, `PostTurnResult`, `RemoteHermesBootPillProps`, `SocketState`) are all in the brand-new `remote-hermes/*` surface from #1174 and intentionally **out of scope** here. They can be cleaned up in a follow-up by whoever owns that surface.

## What changed

`apps/server/src/lib/mcp-manager/index.ts` — slim the barrel to only what's actually imported through it:

  - kept: `humaniseInstallError`, `installInto`, `listAgents`, `uninstallFrom`, `reconcileUrl`, `resetMcpManagerForTesting`, `setMcpManagerForTesting`
  - dropped: `BROWSEROS_MCP_SERVER_NAME`, `BROWSEROS_MCP_STDIO_SERVER_NAME`, `getMcpManager`, `ReconcileUrlInput`, `InstallAgentResult`, `McpAgentId`, `McpAgentRow`, `ReconcileResult`, `UninstallAgentResult` (all of these are only used by sibling files inside `mcp-manager/` itself, which import directly from `./manager`, `./reconcile`, `./types` — never via the barrel)

`apps/server/src/lib/mcp-manager/service.ts`:

  - `BROWSEROS_SERVER_NAMES` and `planFor` narrowed to module-private (both are only used inside `service.ts`)
  - dangling `McpAgentIdentifier` type alias deleted (zero consumers)
  - removed the now-unused `McpAgentId` import that the alias kept alive

`apps/agent/screens/mcp-settings/IntegrationsSection.tsx`:

  - `IntegrationsSectionProps` is now `export interface` so the exported `IntegrationsSection` component no longer references a private type (purely additive — no consumer change)

`apps/agent/screens/mcp-settings/integrations-section.helpers.ts`:

  - `AGENT_PRESENTATION` narrowed to module-private (only used by `presentationFor()` in the same file)

`package.json`:

  - dropped `dotenv` and `picocolors` from `devDependencies`. Neither is imported anywhere in the package; the agent README explicitly notes `dotenv` isn't needed since env files are loaded automatically.
  - `bun.lock` regenerated

## Risk

- For every dropped/narrowed symbol I grepped the whole `packages/browseros-agent` tree to confirm zero external consumers (including the agent UI side and the test suites).
- Removing `export` from a const/function that's only used in the same file is purely a visibility narrowing.
- Adding `export` to an interface is purely additive.
- Removing devDependencies that nothing imports has no runtime effect.

## Test plan

- [x] `bun fallow` in `packages/browseros-agent` — was 17 issues, now 4 (all four are in `remote-hermes/*`, out of scope)
- [x] `bun run typecheck` — clean (`@browseros/agent`, `@browseros/server`, `@browseros/eval`, `@browseros/cdp-protocol`, `@browseros/build-tools` all pass)
- [x] `bun test apps/server/tests/lib/mcp-manager/` — 15/15 pass (both `reconcile.test.ts` and `service.test.ts`)
- [ ] Reviewer to sanity-check that nothing in the broader BrowserOS workspace imports the dropped barrel symbols


DaniAkash:
@greptileai

PR #1173, opened 2026-06-11
feat(patches): WinSparkle Windows updater chromium patches

shadowfax92:
## Summary
- Extracts the WinSparkle chromium-side work into `chromium_patches/` via `browseros dev extract commit` (4 stack commits: third_party vendoring, glue, integration, review fixes) — 15 per-file patches against `BASE_COMMIT` (148.0.7778.97 tag).
- `third_party/winsparkle/` carries only `BUILD.gn` + `README.browseros` (mirrors mac sparkle): the DLLs/headers come from the `winsparkle_setup` on-the-fly download, so the zip-provided text files were pruned to avoid new-file patch collisions in the release build.
- Registers `win-sparkle-updater` and `winsparkle-third-party` features in `features.yaml` (mirrors `mac-sparkle-updater` / `sparkle-third-party`).

## Verification
- Every extracted patch byte-matches `git diff BASE_COMMIT..HEAD -- <file>` in the chromium checkout (15/15).
- Every patch passes `git apply --check` against the bare `BASE_COMMIT` tree via a temp index (15/15) — same as the release build's apply-all onto the clean tag.

## Test plan
- [ ] Windows build will be triggered to validate end to end (this completes the patch-sync prerequisite from #1171; remaining ops step: `appcast-win.xml` on the CDN).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1178, opened 2026-06-12
fix(build): pin enable_winsparkle in Windows GN flag configs

shadowfax92:
## Summary
- Set `enable_winsparkle = true` explicitly in `flags.windows.debug.gn` and `flags.windows.release.gn` — the WinSparkle updater patches (#1173) gate all Windows updater code behind this GN arg, and the release flags file is what CI nightly Windows builds consume.
- Remove the stale "Windows doesn't support Sparkle (macOS update framework)" comment from the debug flags file — #1173 made it false.

## Design
Mirrors the established macOS convention byte-for-byte: `enable_sparkle = is_mac` already defaults true on mac, yet both macOS flag files still pin `enable_sparkle=true`. Same here — `enable_winsparkle = is_win` defaults true on Windows, but pinning it makes the config deterministic and survives any future tightening of the declare_args default (the originally intended Sparkle default was official-build-only). Behavior-neutral today; placement in the release file matches `flags.macos.release.gn` (right after `dcheck_always_on`).

## Test plan
- [x] `grep enable_winsparkle packages/browseros/build/config/gn/flags.windows.*.gn` — both files pin it true
- [x] GN args syntax sanity (`key = value` lines) + `git diff --check` clean
- [x] Adversarial diff review: clean (arg declared in `sparkle_buildflags.gni` patch; `patches` runs before `configure`, so release's `gn gen --fail-on-unused-args` sees the declaration)
- [ ] Next nightly Windows build compiles WinSparkle glue + stages WinSparkle.dll

### DIFF
--- a/packages/browseros/build/config/gn/flags.windows.debug.gn
+++ b/packages/browseros/build/config/gn/flags.windows.debug.gn
@@ -36,8 +36,7 @@ proprietary_codecs = true
 enable_platform_hevc = true
 disable_fieldtrial_testing_config = true
 
-# Windows doesn't support Sparkle (macOS update framework)
-# enable_sparkle = false
+enable_winsparkle = true
 
 # Disable Windows-specific features that macOS doesn't use
 enable_nacl = false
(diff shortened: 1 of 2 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1183, opened 2026-06-13
fix(ci): unstick WarpBuild nightly release builds

shadowfax92:
## Summary
- The WarpBuild nightly (#1170) has never run: all `warp-*` build jobs queue forever with `runner_id: 0`. Root cause is org-side — this repo is **public**, and GitHub blocks org-level self-hosted runners (which is how WarpBuild registers) on public repos until an org admin checks **"Allow public repositories"** on the runner group (see [WarpBuild docs](https://www.warpbuild.com/docs/ci/public-repos)). The `warpbuildbot` app itself is already installed org-wide.
- Fixes the secondary bug that would keep Windows stuck even after the toggle: `warp-windows-2025-x64-32x` is not a WarpBuild image (only Windows Server 2022 exists) → `warp-windows-2022-x64-32x`.
- Adds a `queue-watchdog` job so "no runner ever arrives" fails loudly in ~20 minutes instead of silently holding the `nightly-release` concurrency group for 24h (runs 27367077749 → 27407228486 lost two nights exactly this way; the second was cancelled as part of this work).
- Documents one-time WarpBuild setup (runner-group toggle with an inspect-group-first safety step, account/billing check) and a stuck-in-`queued` troubleshooting runbook.

## Design
The org toggle is deliberately documented rather than automated (needs an org-admin PAT and widens runner exposure for every public repo — a human decision). The watchdog polls the run's own jobs API every 2 min with a 20-min deadline and only job-level `actions: write`; at the deadline it cancels the run only when **no** build job is actually running (a stuck-queued job is then the only thing pinning the run), fails without cancelling while any build is live, tolerates 3 consecutive API failures, and refuses to cancel if its job-name filter matches nothing. Build/publish do not depend on it.

## Test plan
- [x] `actionlint` (incl. shellcheck of the watchdog script) and YAML parse green
- [x] Watchdog jq filter dry-run against the real stuck run's jobs payload (3/3 queued → cancel branch) plus simulated picked-up/partial/empty payloads
- [x] Job-graph asserted: `build` needs only `plan`; `publish` needs `[plan, build]`; top-level permissions still `contents: read`
- [ ] After merge + org toggle: `gh workflow run "Nightly Release Build" -f platforms=linux` and confirm the job leaves `queued` within ~5 min

### DIFF
--- a/.github/workflows/nightly-release.yml
+++ b/.github/workflows/nightly-release.yml
@@ -80,6 +80,48 @@ jobs:
             echo "publish=$publish"
           } >> "$GITHUB_OUTPUT"
 
+  # WarpBuild provisions runners on demand via webhook; when none arrive
+  # (org runner-group policy, bad label, account issue) the build jobs sit
+  # queued until GitHub discards them 24h later, holding the
+  # nightly-release concurrency group all day. Fail fast instead — see the
+  # troubleshooting section of
+  # packages/browseros/build/docs/nightly-warpbuild-ci.md.
+  queue-watchdog:
+    needs: plan
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    permissions:
+      actions: write
+    steps:
+      - name: Fail fast when no runner picks up the builds
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          set -euo pipefail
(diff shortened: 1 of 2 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1184, opened 2026-06-13
fix(ci): unresolvable setup-uv pin broke nightly; restore 2025 label

shadowfax92:
## Summary
Follow-up to #1183, driven by live verification after the org runner-group toggle was flipped (pickup now works — a queued job was claimed in ~60s):

- **The real remaining breaker**: `astral-sh/setup-uv@v8` is unresolvable — astral-sh publishes v8.x.y *releases* but no floating `v8` major tag (majors stop at `v7`). Every build job died in "Set up job" within seconds, on all platforms. Pinned to the exact existing tag `v8.2.0`. All other action refs were verified to resolve.
- **Windows label restored to `warp-windows-2025-x64-32x`**: the account's WarpBuild runner catalog (dashboard → Runners) lists Server 2025 tiers — the #1183 change to `2022` was based on WarpBuild's stale public preinstalled-software page. `2022` did provision (alias), but it's not a catalog-listed pinned label; `2025` is, and matches the original #1170 intent.
- **Doc corrections**: catalog is now documented as the canonical source for labels/specs; disk specs fixed (linux 256 GB, mac 500 GB); org toggle recorded as completed 2026-06-13 with live-verified pickup; the wrong "WarpBuild does not image windows-2025" claim removed; new troubleshooting entry for the `Unable to resolve action @vN` failure signature.

## Test plan
- [x] `actionlint` green; `git diff --check` clean
- [x] `v8.2.0`, `checkout@v6`, `*-artifact@v4`, `WarpBuilds/cache@v1` tags verified to exist via the GitHub API
- [x] Pickup already proven live post-toggle (linux + windows runners claimed jobs)
- [ ] Post-merge: dispatch full run — all three lanes should pass "Set up job" and start building (cold-cache, 4-7h)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1185, opened 2026-06-13
fix(ci): windows nightly died logging emoji to cp1252 stdout

shadowfax92:
## Summary
The first windows nightly to reach the build CLI (run 27451026914, after #1184 fixed runner pickup) crashed in 2 seconds at its very first log line: `UnicodeEncodeError: 'charmap' codec can't encode character '\U0001f680'`. Windows runners pipe python stdout as cp1252, and the build CLI emits ~300 emoji across its loggers — `log_warning`/`log_error`/`log_success` bake them in, so stripping is not viable. Linux/macOS lanes in the same run are unaffected and still building.

Three thin layers, each verified:
- **Workflow**: `PYTHONUTF8: "1"` job-level env on `build` — UTF-8 mode for the CLI and every python subprocess (gclient, hooks). No-op on Linux/macOS; other jobs untouched.
- **`logger.py`**: reconfigure stdout/stderr with `errors="backslashreplace"` at import, so logging degrades to escapes instead of crashing in any environment (matches stderr's native fallback; isinstance-guarded for pyright).
- **`utils.py` (review finding)**: `run_command`'s Popen decoded child output strictly — under UTF-8 mode, one ANSI-codepage byte from a native tool (MSVC, mt.exe) would `UnicodeDecodeError` a multi-hour ninja stream. Added `errors="replace"` to the decode side of the funnel.

## Test plan
- [x] Crash reproduced locally with `PYTHONIOENCODING=cp1252` (identical traceback to CI), fix verified green (emoji → escapes, exit 0)
- [x] ruff + actionlint clean; pyright error count identical pre/post (2 pre-existing in utils.py, none introduced)
- [ ] Tonight's 08:00 UTC scheduled run is the live verification (concurrency group serializes behind the in-flight linux/macOS builds, so no manual re-dispatch)

### DIFF
--- a/.github/workflows/nightly-release.yml
+++ b/.github/workflows/nightly-release.yml
@@ -152,6 +152,11 @@ jobs:
     defaults:
       run:
         shell: bash
+    env:
+      # Windows runners pipe stdout as cp1252, which crashes the emoji-heavy
+      # build CLI with UnicodeEncodeError; force UTF-8 for python and all
+      # its subprocesses (gclient, hooks). No-op on Linux/macOS.
+      PYTHONUTF8: "1"
     steps:
       - uses: actions/checkout@v6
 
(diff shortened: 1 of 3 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1201, opened 2026-06-15
fix: default boolean env parser to false

shadowfax92:
## Summary
- Replace the BROWSER_USE_NEW_TOOLS-specific parser with a local strict boolean env parser.
- Make absent boolean env values parse to false instead of undefined.
- Preserve strict true/false validation and env-specific error messages.

## Design
The config loader now gets a concrete boolean from parseBooleanEnv for BROWSER_USE_NEW_TOOLS. Config file and CLI precedence are unchanged because runtime env still merges before those higher-precedence sources.

## Test plan
- [x] cd apps/server && bun test tests/config.test.ts
- [x] bun run lint
- [x] bun run typecheck
- [x] bun run test:main


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1207, opened 2026-06-15
feat: add metrics sampling option

shadowfax92:
## Summary
- Add a third `metrics.log` options argument for local event sampling.
- Keep aggregated `mcp.request` and `tool_executed` rollups exact.
- Sample `chat.request` metrics at `1 / 5` and tag captured sampled events with `sample_rate`.

## Design
Sampling is per-call and only applies to immediate captures after the existing rollup checks. Missing or invalid sampling values preserve full capture, so existing call sites remain unchanged unless they opt in.

## Test plan
- [x] bun test apps/server/tests/lib/metrics.test.ts
- [x] bunx biome check apps/server/src/lib/metrics.ts apps/server/src/api/routes/chat.ts apps/server/tests/lib/metrics.test.ts
- [x] bun run typecheck
- [x] bun run test:main


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1218, opened 2026-06-16
fix: remove deprecated controller extension plumbing

shadowfax92:
## Summary
- Remove deprecated controller extension port config from the BrowserOS server config surface.
- Drop Chromium-side controller extension prefs while preserving the default runtime extension port internally.
- Add explicit deprecated BrowserOS extension cleanup so old managed extension IDs are uninstalled instead of reinstalled.

## Design
This keeps the active BrowserOS extension list focused on supported bundled extensions, moves the retired controller ID into an explicit deprecated-extension cleanup list, and removes `extensionPort` from persisted/user-facing server configuration while leaving the manager's default internal value available where existing runtime code still expects it.

## Test plan
- [x] git diff --check origin/main...HEAD
- [x] Chromium build confirmed by requester outside this session

### DIFF
--- a/packages/browseros-agent/apps/server/src/config.ts
+++ b/packages/browseros-agent/apps/server/src/config.ts
@@ -20,7 +20,6 @@ const ServerConfigSchema = z.object({
   cdpPort: portSchema.nullable(),
   serverPort: portSchema,
   agentPort: portSchema,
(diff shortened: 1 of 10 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1220, opened 2026-06-16
fix: remove generated bundled extensions patch

shadowfax92:
## Summary
- Remove the Chromium patch file for `bundled_extensions.json`.
- Keep `bundled_extensions.json` owned by the `bundled_extensions` build module, which generates it before the patch phase.
- Prevent patch application from failing when `git apply` tries to create a generated file that already exists.

## Design
The release pipeline runs `bundled_extensions` before `patches`, so the JSON manifest should be generated into the Chromium checkout and not also represented as a `chromium_patches` add-file patch. This leaves the bundled extension `BUILD.gn` patch intact while removing the duplicate generated manifest patch.

## Test plan
- [x] `git diff --check`
- [x] `PYTHONPATH=packages/browseros python3 -m unittest build.modules.apply.common_test build.modules.apply.apply_all_test build.modules.patches.patches_test`
- [x] `make test` in `packages/browseros/tools/patch`
- [x] `go test ./internal/modules/extensions ./internal/patchwork ./internal/modules/patchmod` in `packages/browseros/build_go`

Note: full `make test` in `packages/browseros/build_go` is currently red on stale resolver expectations already present on `origin/main` (`release.linux.yaml` expects arm64 instead of current x64, and `release.windows.yaml` expects one fewer pipeline module despite `sparkle_sign`).


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1236, opened 2026-06-17
chore(ci): make nightly release workflow manual

shadowfax92:
## Summary
- Make the Nightly Release Build workflow manual-only by removing the scheduled cron trigger.
- Remove the stale schedule-based publish branch so `publish_nightly` is the manual publish gate.
- Update the workflow header comment from daily builds to manual builds.

## Design
The workflow now starts only from `workflow_dispatch`. Manual operators keep the existing platform selector and can still set `publish_nightly` on the default branch to update the rolling prerelease.

## Test plan
- [x] `git diff --check`
- [x] `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/nightly-release.yml"); puts "yaml ok"'`\n- [x] `actionlint .github/workflows/nightly-release.yml`\n- [x] `bun run check`

### DIFF
--- a/.github/workflows/nightly-release.yml
+++ b/.github/workflows/nightly-release.yml
@@ -1,6 +1,6 @@
 name: Nightly Release Build
 
-# Daily UNSIGNED release builds of BrowserOS for all three platforms on
+# Manual UNSIGNED release builds of BrowserOS for all three platforms on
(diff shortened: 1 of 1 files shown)


shadowfax92:
@claude review

shadowfax92:
@greptileai review

PR #1238, opened 2026-06-17
fix: include drizzle migrations for all server targets

shadowfax92:
## Summary
- Include Drizzle migrations in every production server resource artifact by removing platform filters from the migration resource rule.
- Add staging test coverage that checks the production manifest selects the migration rule for every supported server target.

## Design
The resource manifest already treats missing `os` and `arch` fields as unfiltered. The migration resource is shared local data rather than a platform-specific binary, so it should use that existing all-target behavior.

## Test plan
- [x] `bun test scripts/build/server/stage.test.ts`
- [x] `bun run check`
- [x] `bun run build:server:test`


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1245, opened 2026-06-18
fix: include BrowserOS server assets in installer

shadowfax92:
## Summary
- Include BrowserOS server `third_party` runtime assets in the Windows mini-installer release manifest.
- Include BrowserOS server database migrations and migration metadata in the packaged version directory.
- Sync resolver test expectations with the committed Linux and Windows release configs so the BrowserOS build_go suite passes.

## Design
The installer patch keeps the existing BrowserOSServer resource mapping and adds the missing nested directories that the server needs at runtime. The test update only aligns expected release config resolution with current YAML: Linux release is pinned to x64, and Windows release includes the sparkle_sign module.

## Test plan
- [x] `git diff --check origin/main..HEAD`
- [x] `git apply --numstat packages/browseros/chromium_patches/chrome/installer/mini_installer/chrome.release`
- [x] `go test ./...` in `packages/browseros/tools/patch`
- [x] `go test ./...` in `packages/browseros/build_go`
- [ ] `autoninja -C out/Default_arm64 chrome` could not run in this checkout: `out/Default_arm64` does not exist


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1255, opened 2026-06-18
fix: preserve Windows bundled Bun launcher paths

shadowfax92:
## Summary
- Quote bundled Bun ACP launcher command tokens with acpx-safe single quotes so Windows backslashes survive command splitting.
- Add launcher coverage for Windows `C:\...\bun.exe` paths and bundled paths containing spaces.
- Keep resource packaging and custom ACP commands unchanged.

## Design
The failure was at the ACP command-string boundary: acpx treats backslashes as escapes outside single quotes, so a double-quoted Windows path could become `C:Users...`. The launcher now quotes the Bun executable token in the form acpx preserves before spawning.

## Test plan
- [x] `bun test apps/server/tests/lib/agents/host-acp/launcher.test.ts`
- [x] `bun run check`
- [x] local adversarial review, 2 rounds; first finding fixed, second clean

### DIFF
--- a/packages/browseros-agent/apps/server/src/lib/agents/host-acp/launcher.ts
+++ b/packages/browseros-agent/apps/server/src/lib/agents/host-acp/launcher.ts
@@ -56,15 +56,15 @@ export function resolveAcpSpawnCommand(
     platform: input.platform,
   })
   if (bunPath) {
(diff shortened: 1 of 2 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1262, opened 2026-06-18
feat: add side panel scope toggle

shadowfax92:
## Summary
- Add a Customization switch for one side panel per window, stored in `browseros.side_panel.per_window`.
- Route toolbar side panel toggles through BrowserOS tab APIs by default and Chromium window `sidePanel.open/close` when the setting is enabled.
- Keep programmatic/search opens on the BrowserOS API because Chromium `sidePanel.open` requires a live user gesture.
- Persist per-window open state in session storage so service worker restarts preserve toggle semantics.

## Verification
- `bun run --filter @browseros/agent test`
- `bun run lint`
- `bun run typecheck`
- `bun run test:main`
- `bun run build:agent`
- `git diff --check`


shadowfax92:
@claude review

shadowfax92:
@greptileai review

PR #1263, opened 2026-06-18
fix: store side panel scope in extension storage

shadowfax92:
## Summary
- moves the side panel scope toggle off the invalid BrowserOS pref and into WXT local extension storage
- keeps background hydration, runtime apply, and rollback behavior wired to the local setting
- adds coverage for true, false, absent, stale refresh, native-pref failure, and rollback cases

## Design
The side panel scope value is now owned by `sidePanelPerWindowStorage` in the side-panel storage module. Background routing hydrates from that storage, while the settings card reads/writes the same local item and only uses native prefs for the settings that actually have native pref backing.

## Test plan
- [x] `bun test apps/agent/lib/browseros/toggleSidePanel.test.ts apps/agent/screens/customization/ToolbarSettingsCard.test.tsx`
- [x] `bun run --filter @browseros/agent test`
- [x] `bun run --filter @browseros/agent typecheck`
- [x] `bun run --filter @browseros/agent lint`
- [x] `rg -n "SIDE_PANEL_PER_WINDOW|browseros\\.side_panel\\.per_window" apps/agent -g '!**/*.test.*'`

### DIFF
--- a/packages/browseros-agent/apps/agent/lib/browseros/prefs.ts
+++ b/packages/browseros-agent/apps/agent/lib/browseros/prefs.ts
@@ -9,7 +9,6 @@ export const BROWSEROS_PREFS = {
   RESTART_SERVER: 'browseros.server.restart_requested',
   SHOW_LLM_CHAT: 'browseros.show_llm_chat',
   SHOW_TOOLBAR_LABELS: 'browseros.show_toolbar_labels',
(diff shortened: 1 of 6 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1269, opened 2026-06-18
feat: share snapshot formatting for post-actions

shadowfax92:
## Summary
- Extract shared snapshot formatting into `formatSnapshotResult` so the direct `snapshot` tool and snapshot post-actions use the same output path.
- Route both legacy and compact snapshot post-actions through the shared formatter, preserving `[Page N snapshot]` prefixes.
- Add large snapshot post-action coverage, including a save-failure fallback that returns a bounded wrapped excerpt instead of silently dropping context.

## Design
Snapshot formatting now lives beside the browser snapshot tool, matching the existing `formatDiffResult` pattern. Callers still own page-specific response shape, while the formatter owns empty-page fallback, trust-boundary wrapping, large-output file writes, metadata, and save-failure fallback text.

## Test plan
- [x] `bun test apps/server/tests/tools/browser/framework.test.ts --timeout 30000`
- [x] `bun test apps/server/tests/tools/response.test.ts --timeout 30000`
- [x] `bun test apps/server/tests/tools/browser/register.test.ts --timeout 30000`
- [x] `bun run check`

### DIFF
--- a/packages/browseros-agent/apps/server/src/tools/browser/snapshot-format.ts
+++ b/packages/browseros-agent/apps/server/src/tools/browser/snapshot-format.ts
@@ -3,6 +3,7 @@ import { wrapUntrusted } from './trust-boundary'
 
 const LARGE_SNAPSHOT_WORD_THRESHOLD = 5_000
 const LARGE_SNAPSHOT_CHAR_THRESHOLD = 50_000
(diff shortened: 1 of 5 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1272, opened 2026-06-19
feat: stabilize snapshot refs

shadowfax92:
## Summary
- Reuse public `eN` snapshot refs for the same backend DOM nodes within a stable document.
- Reset the public ref namespace on top-level navigation, URL changes, reloads, or unavailable document identity.
- Keep diff matching ref-sensitive while preventing insertion-before-existing-node ref cascades.

## Design
`Observer` now owns a committed document/url ref scope and renders each capture into a transactional `RefMap` fork. `RefMap` keeps document-scoped backend-node assignments separate from latest-snapshot resolution metadata, while no-document or unstable child-frame captures fall back to traversal-order refs.

## Test plan
- [x] `bun test apps/server/src/browser/core/snapshot/refs.test.ts apps/server/src/browser/core/snapshot/render.test.ts apps/server/src/browser/core/snapshot/diff.test.ts apps/server/src/browser/core/observer/resolve.test.ts apps/server/src/browser/core/observer/observer.test.ts`
- [x] `bun run check`
- [x] Local adversarial review pass: clean after fixes

### DIFF
--- a/packages/browseros-agent/apps/server/src/browser/core/observer/observer.test.ts
+++ b/packages/browseros-agent/apps/server/src/browser/core/observer/observer.test.ts
@@ -1,0 +1,245 @@
+import { describe, expect, test } from 'bun:test'
+import type { ProtocolApi } from '@browseros/cdp-protocol/protocol-api'
+import type { PageManager } from '../pages'
+import type { AXNode } from '../snapshot/ax-types'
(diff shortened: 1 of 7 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1273, opened 2026-06-19
fix: set wait default timeout to 2s

shadowfax92:
## Summary
- Changed the browser wait tool omitted-timeout default from 10s to 2s.
- Updated the tool schema description so callers see the 2000ms default.
- Added regression coverage through the registered browser tool path.

## Design
The existing wait tool already centralizes timeout handling through DEFAULT_WAIT_TIMEOUT_MS and clampTimeout, so this keeps the change scoped to that constant plus the public schema metadata.

## Test plan
- [x] cd apps/server && bun --env-file=.env.development test ./tests/tools/browser/register.test.ts
- [x] bun run check


shadowfax92:
@claude review

shadowfax92:
@greptileai review

PR #1274, opened 2026-06-19
fix(build): drop claude from server signing

shadowfax92:
## Summary
- Remove Claude from BrowserOS server binary signing metadata for Python and Go release paths.
- Keep Codex signing expectations intact and update package/storage contract tests for the unshipped Claude CLI.
- Add regression coverage for Windows OTA signing without `third_party/claude.exe`.

## Design
The release signing paths both derive expected Windows server binaries from shared metadata. The fix removes Claude from that shared metadata instead of weakening missing-binary checks, so Windows signing still fails for truly missing shipped binaries while no longer requiring an unshipped Claude executable.

## Test plan
- [x] `uv run python -m unittest discover -s build -p '*_test.py'`
- [x] `go test ./...`


shadowfax92:
@claude review

shadowfax92:
@greptileai review

PR #1275, opened 2026-06-19
fix: use token thresholds for browser output

shadowfax92:
## Summary
- Replaces snapshot word+char cutoffs with a single 15,000 estimated-token threshold.
- Replaces diff word cutoffs with a single 10,000 estimated-token threshold.
- Reports `tokenEstimate` for saved browser output and keeps save-failure excerpts bounded.

## Design
Browser snapshot and diff formatters now use a local plain-text helper that mirrors compaction's `Math.ceil(chars / 3)` heuristic. They estimate the wrapped browser output payload before deciding whether to inline it or write it to a BrowserOS output markdown file.

## Test plan
- [x] `cd apps/server && bun test tests/tools/browser/register.test.ts tests/tools/browser/framework.test.ts tests/tools/response.test.ts`
- [x] `bun run check`
- [x] local adversarial review, round 2 clean


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1277, opened 2026-06-19
fix(server): make wait time optional with a 2s default

shadowfax92:
## Summary
- The `wait` tool's pause time is now **optional with a 2000ms default** — weaker models that fumble the millisecond `value` (the screenshot showed one sending `">"`, hitting `wait: value must be milliseconds.`, then abandoning the wait for a snapshot) now just pause instead of erroring.
- `for` defaults to `"time"`, so `wait({ page })` is a valid ~2s pause.
- For `for="time"`, garbage/empty/negative `value` falls back to the default rather than erroring; `value` also accepts a number (not just a string).
- `for="text"`/`"selector"` still require a `value`, now with an instructive error pointing to `for="time"` for a plain pause. The paused duration is reported (`waited <n>ms` / `waitedMs`).

## Design
The fix lives entirely in `apps/server/src/tools/browser/wait.ts`. The Zod schema advertises the optionality (`for` `.default('time')`, `value` `union(string|number).optional()`), and `parseWaitMs` became a total function that returns the fallback for missing/empty/garbage/negative input instead of `null` — removing the only hard-error path for `for="time"`. No other tools touched; no shared-constant churn.

## Reconciliation with #1273
#1273 ("set wait default timeout to 2s") landed on `main` while this was open. It achieved the 2s default by lowering `DEFAULT_WAIT_TIMEOUT_MS` 10000 → 2000, which also caps every wait and shortens text/selector polling to 2s. This PR **keeps #1273's `DEFAULT_WAIT_TIMEOUT_MS = 2000`** (the giving-up / poll cap) and adds a **separate `DEFAULT_PAUSE_MS = 2000`** for the `for="time"` default. They're intentionally distinct: a model that crams the duration into `timeout` (a classic wrong-input case) still only pauses 2s by default, rather than blocking for the whole timeout.

## Test plan
- [x] `bun test apps/server/tests/tools/browser/wait.test.ts` — schema defaults, `parseWaitMs` leniency/boundaries, handler pauses (and reports `waitedMs`) on garbage and on no value.
- [x] `bun run check` (lint · typecheck · fallow · server tools + integration tests) — green, before and after merging `main`.


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1282, opened 2026-06-19
feat: restore per-window side panel conversation on reopen

shivammittal274:
## What
When the side panel scope is **per-window**, persist each window's active conversation (`windowId → conversationId`) and restore it when the panel remounts — so a closed/reopened panel resumes that window's chat instead of starting blank. No-op in per-tab scope.

Follow-up to the side-panel scope toggle (#1263) — the "conversationId on the new pane to pull" piece.

## Changes
- `lib/browseros/perWindowConversationStorage.ts` (new): session-scoped `windowId → conversationId` map.
- `modules/chat/chat-session.hooks.ts`: on mount in per-window scope, resolve the window id and restore its stored conversation via the existing `?conversationId` path (else claim it); persist on conversation change. Gated on `sidePanelPerWindowStorage`, so **per-tab behavior is unchanged**.

## Why it's small
Verified in a dev build that BrowserOS's window-scoped panel already persists across tab switches (single instance). So the only gap was a fresh remount (close+reopen / new window) starting blank — which this fixes.

## Testing
- typecheck ✅, lint ✅, 197/197 agent tests ✅
- Live dev-build checks (per-window scope):
  - Tab switches keep the same panel instance — conversation persists.
  - Close + reopen panel → restores the window's conversation.
  - Fresh window → claims/persists a conversation id.


shivammittal274:
@greptile review

PR #1285, opened 2026-06-19
feat: gate hermes on alpha support

shadowfax92:
## Summary
- Move `HERMES_AGENT_SUPPORT` from development-only gating to the existing alpha feature gate.
- Keep Hermes enabled automatically in development through the existing alpha/static shortcut.
- Update capability tests and the duplicated customization test mock to cover the alpha-gated behavior.

## Design
Hermes visibility already flows through `Capabilities.supports(Feature.HERMES_AGENT_SUPPORT)`, so the central capability config is the right ownership point. This changes only that config entry from `requiresDevelopmentFlag` to `requiresAlphaFlag` and aligns the tests with production alpha-on/off behavior.

## Test plan
- [x] `bun test apps/agent/lib/browseros/capabilities.test.ts apps/agent/screens/customization/ToolbarSettingsCard.test.tsx`
- [x] `bun run check`

### DIFF
--- a/packages/browseros-agent/apps/agent/lib/browseros/capabilities.test.ts
+++ b/packages/browseros-agent/apps/agent/lib/browseros/capabilities.test.ts
@@ -67,7 +67,7 @@ describe('resolveStaticFeatureSupport', () => {
 })
 
 describe('resolveFeatureStaticSupport', () => {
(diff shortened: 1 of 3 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1287, opened 2026-06-19
feat(patches): browseros showInfoBar extension API

shadowfax92:
## Summary
- add `chrome.browserOS.showInfoBar(message, callback)` to the BrowserOS extension API patch
- add the BrowserOS extension infobar delegate and metrics IDs, keeping `showToast` at `1987` and using `showInfoBar` at `1988`
- extract corrected Chromium squash commit `d74636bf879f5b3bd6dd127466cc7143f5344b91` into BrowserOS patches

## Design
`showInfoBar` follows the existing BrowserOS extension API pattern and displays a simple alert infobar in the active BrowserOS/Chrome tab. It uses the same confirm-infobar creation path as Chromium's simple alert helper, but calls `InfoBarManager::AddInfoBar` directly so the callback reports whether the infobar was actually accepted.

## Test plan
- [x] `/Users/shadowfax/.skills/sf-ch-mux/scripts/chpool build -- autoninja -C out/Default_arm64 chrome`
- [x] `/Users/shadowfax/.skills/sf-ch-extract/scripts/verify-patches.sh --chromium-src /Users/shadowfax/ch2-src --worktree /Users/shadowfax/worktrees/main/feat-browseros-show-infobar-patches --tip d74636bf879f5b3bd6dd127466cc7143f5344b91`
- [x] `git diff --check`

### DIFF
--- a/packages/browseros/chromium_patches/chrome/browser/extensions/api/browser_os/browser_os_api.cc
+++ b/packages/browseros/chromium_patches/chrome/browser/extensions/api/browser_os/browser_os_api.cc
@@ -1,9 +1,9 @@
 diff --git a/chrome/browser/extensions/api/browser_os/browser_os_api.cc b/chrome/browser/extensions/api/browser_os/browser_os_api.cc
 new file mode 100644
-index 0000000000000..a136ab744a9e6
+index 0000000000000..62fdae9e171ae
 --- /dev/null
 +++ b/chrome/browser/extensions/api/browser_os/browser_os_api.cc
-@@ -0,0 +1,291 @@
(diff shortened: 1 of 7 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

shadowfax92:
Fixed the showInfoBar callback issue in bd887397 by using InfoBarManager::AddInfoBar's return value. Note: CreateSimpleAlertInfoBar returns void in this Chromium checkout, so the patch mirrors that helper's construction path and captures AddInfoBar instead.

PR #1288, opened 2026-06-19
fix: replace Record<string, any> with Record<string, unknown> in GraphQL hooks

Lothnic:
## Description:
- Remove stale TODOs from mediaUrls.ts (URLs already present)
- Rename ambiguous type params T/V to TResult/TVariables
- Replace Record<string, any> | undefined with Record<string, unknown> = Record<string, never>
- Remove no-longer-needed biome-ignore suppressions

### CLA Signature


Lothnic:

DaniAkash:
Thanks for the contribution, but we're going to pass on this one — it's a small type-only cleanup that doesn't meaningfully change behavior, and the surrounding code has moved on since June. Appreciate the effort.

PR #1290, opened 2026-06-19
fix: inline agent runner jwt secret in server packaging

shadowfax92:
## Summary
- Add `AGENT_RUNNER_JWT_SECRET` to the server production build inline env allowlist.
- Document the variable in `apps/server/.env.production.example`.
- Add focused build config coverage for file env, process env override, and optional absence.

## Design
The production server packaging path already builds an `envVars` map in `loadBuildConfig` and passes it into Bun `define` during bundling. This change extends that existing allowlist without adding the secret to required production validation, so the Remote Hermes path remains opt-in for now.

## Security note
This intentionally supports the requested temporary behavior. If set for distributable release artifacts, this embeds a shared HS256 signing secret in the packaged server binary. Before public/long-lived rollout, this should move to per-install credentials, server-minted short-lived tokens, or asymmetric auth.

## Test plan
- [x] `bun test scripts/build/server/config.test.ts`
- [x] `bun test scripts/build/server/stage.test.ts`
- [x] `AGENT_RUNNER_JWT_SECRET=test-secret bun scripts/build/server.ts --target=all --ci`
- [x] `bun run check`


shadowfax92:
@greptileai review

shadowfax92:
@claude review

shadowfax92:
@greptileai review

PR #1292, opened 2026-06-19
fix: gate Hermes support on server version

shadowfax92:
## Summary
- Keep Hermes agent support behind the alpha feature flag.
- Add a minimum BrowserOS server version gate of `0.0.116` for Hermes.
- Make satisfied static gates fall through when a feature also has version constraints, so alpha is necessary but not sufficient.

## Design
Hermes stays in the existing `FEATURE_CONFIG` map with `requiresAlphaFlag: true` plus `minServerVersion: '0.0.116'`. `resolveFeatureStaticSupport` now returns `null` when static gates pass but version gates still need evaluation, preserving the existing `Capabilities.supports()` flow without adding Hermes-specific branching.

## Test plan
- [x] `bun test apps/agent/lib/browseros/capabilities.test.ts`
- [x] `bun run lint`
- [x] `bun run typecheck`
- [x] `bun run test:main`
- [x] `bun run check`

### DIFF
--- a/packages/browseros-agent/apps/agent/lib/browseros/capabilities.test.ts
+++ b/packages/browseros-agent/apps/agent/lib/browseros/capabilities.test.ts
@@ -67,7 +67,7 @@ describe('resolveStaticFeatureSupport', () => {
 })
 
 describe('resolveFeatureStaticSupport', () => {
-  it('gates Hermes support on alpha features', () => {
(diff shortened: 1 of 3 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1298, opened 2026-06-19
fix: include full browser output in structured responses

shadowfax92:
## Summary
- Include full wrapped snapshot content in direct snapshot structured JSON responses.
- Include full wrapped diff content in direct diff structured JSON responses, with URL-change diffs exposing the current snapshot under `snapshot`.
- Preserve existing MCP/text behavior that writes oversized browser output to generated markdown files.

## Design
The shared snapshot and diff formatters already compute trust-wrapped browser output before deciding whether text should be inline or file-backed. This keeps the text/MCP path unchanged while adding the same wrapped content to structured JSON for direct API callers.

## Test plan
- [x] `cd apps/server && bun test tests/tools/browser/register.test.ts`
- [x] `bun run lint`
- [x] `bun run typecheck`
- [x] `bun run test:main`

### DIFF
--- a/packages/browseros-agent/apps/server/src/tools/browser/diff-format.ts
+++ b/packages/browseros-agent/apps/server/src/tools/browser/diff-format.ts
@@ -29,6 +29,7 @@ export async function formatDiffResult(
       beforeUrl: diff.beforeUrl,
       afterUrl: diff.afterUrl,
     }),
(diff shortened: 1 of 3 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1299, opened 2026-06-19
fix: include large browser output excerpts

shadowfax92:
## Summary
- Save full oversized browser snapshot and diff output to BrowserOS output files as before.
- Include the first 5000 estimated tokens inline on oversized success responses.
- Keep inline excerpts wrapped with untrusted-content markers and preserve save-failure fallback metadata.

## Design
The formatter layer remains the owner of oversized output handling. `formatDiffResult` and `formatSnapshotResult` now use a shared estimated-token prefix helper so oversized responses can include a bounded wrapped excerpt while the full wrapped payload is still written to disk for readback.

## Test plan
- [x] `bun --env-file=apps/server/.env.development test apps/server/tests/tools/browser/framework.test.ts apps/server/tests/tools/browser/register.test.ts apps/server/tests/tools/response.test.ts`
- [x] `bun run lint`
- [x] `bun run typecheck`
- [x] `bun run test:main`
- [x] `bun run check`

## Local review
- [x] `sf-review` context-free subagent: clean, 0 findings

### DIFF
--- a/packages/browseros-agent/apps/server/src/tools/browser/diff-format.ts
+++ b/packages/browseros-agent/apps/server/src/tools/browser/diff-format.ts
@@ -1,10 +1,13 @@
 import type { SnapshotDiff } from '../../browser/core/snapshot/diff'
 import { writeTempToolOutputFile } from './output-file'
-import { estimateTextTokens } from './token-estimate'
+import {
+  estimateTextTokens,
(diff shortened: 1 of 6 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1300, opened 2026-06-19
feat: add screenshot quality option

shadowfax92:
## Summary
- Defaults the browser screenshot tool to JPEG quality 80 for smaller inline visual captures.
- Adds optional `format`, `quality`, and `size` inputs; normal viewport captures scale toward 1024x768 with CDP `clip.scale`.
- Keeps PNG quality omitted and skips size clipping for `fullPage` captures.
- Replaces a leaking OAuth module mock in an LLM config test so the full root test suite can run green.

## Design
The screenshot tool now builds a typed `Page.captureScreenshot` request directly from validated Zod inputs. For non-full-page captures it reads layout metrics and sets a scale-only viewport clip, preserving aspect ratio without upscaling. JPEG uses default quality 80 unless overridden; PNG and WebP omit the JPEG-only quality field.

## Test plan
- [x] `bun --env-file=apps/server/.env.development test apps/server/tests/tools/browser/register.test.ts`
- [x] `bun run check`
- [x] `bun run test`
- [x] context-free local review: clean

### DIFF
--- a/packages/browseros-agent/apps/server/src/tools/browser/screenshot.ts
+++ b/packages/browseros-agent/apps/server/src/tools/browser/screenshot.ts
@@ -1,23 +1,89 @@
+import type {
+  CaptureScreenshotParams,
+  Viewport,
+} from '@browseros/cdp-protocol/domains/page'
+import type { ProtocolApi } from '@browseros/cdp-protocol/protocol-api'
 import { z } from 'zod'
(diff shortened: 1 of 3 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1315, opened 2026-06-22
fix: align Biome checks with pinned CLI

shadowfax92:
## Summary
- Switch BrowserOS agent Biome hook commands from npm's `npx` path to Bun's scoped `@biomejs/biome` runner.
- Update agent lint scripts to call `@biomejs/biome` instead of the unrelated `biome` package.
- Keep the existing Biome 2.5.0 config unchanged because package.json and bun.lock already pin 2.5.0.

## Design
The reported failure happens when the Biome check runs with CLI 2.4.8 against a 2.5.0 config. The repo already declares and locks `@biomejs/biome@2.5.0`, so the fix is to make hooks and scripts resolve that scoped package consistently instead of using a transient npm path or the unscoped `biome` package.

## Test plan
- [x] `npx @biomejs/biome@2.4.8 check --config-path biome.json --no-errors-on-unmatched --files-ignore-unknown=true --colors=off biome.json` reproduces the original error
- [x] `bunx @biomejs/biome --version && bunx @biomejs/biome check --no-errors-on-unmatched --files-ignore-unknown=true --colors=off biome.json package.json apps/agent/package.json ../../lefthook.yml lefthook.yml`
- [x] `git diff --check`
- [x] `bun install --frozen-lockfile`
- [x] `bun run codegen:agent`
- [x] `bun run check`
- [x] `bun run test`

### DIFF
--- a/lefthook.yml
+++ b/lefthook.yml
@@ -19,7 +19,7 @@ pre-commit:
     biome-check:
       root: "packages/browseros-agent/"
       glob: "*.{js,ts,cjs,mjs,d.cts,d.mts,jsx,tsx,json,jsonc}"
-      run: npx @biomejs/biome check --write --no-errors-on-unmatched --files-ignore-unknown=true --colors=off {staged_files}
+      run: bunx @biomejs/biome check --write --no-errors-on-unmatched --files-ignore-unknown=true --colors=off {staged_files}
       stage_fixed: true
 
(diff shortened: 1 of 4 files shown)


shadowfax92:
@claude review

shadowfax92:
@greptileai review

PR #1321, opened 2026-06-22
fix: preserve remote harness output file access

shadowfax92:
## Summary
- Keep remote harness MCP filesystem tools behind the generic `source=remote-agent-harness` gate.
- Use one `remoteAgentHarness` context to carry generated BrowserOS output-file access through MCP registration.
- Allow remote harness `filesystem_read` to read browser tool output files returned through that harness, while default MCP stays browser-only.

## Design
The route detects remote harness callers from the query parameter and passes a single optional `remoteAgentHarness` object into MCP server creation. Its presence is the only internal signal for registering filesystem tools, and the same object carries the output-file access used by browser tools and `filesystem_read`.

Workspace files remain scoped through `executionDir`. Browser-generated output files remain allow-listed by the output-file access object, so `filesystem_read` can read paths returned by browser tools without exposing arbitrary BrowserOS state paths.

## Test plan
- `cd apps/server && bun test tests/api/routes/mcp.test.ts tests/api/services/mcp/register-mcp.test.ts`
- `bun run lint`
- `bun run typecheck`
- `bun run check`
- `bun run test`

Not merging per review request.

### DIFF
--- a/packages/browseros-agent/apps/server/src/api/routes/mcp.ts
+++ b/packages/browseros-agent/apps/server/src/api/routes/mcp.ts
@@ -10,37 +10,19 @@ import type { BrowserSession } from '../../browser/core/session'
 import { logger } from '../../lib/logger'
 import { metrics } from '../../lib/metrics'
 import { Sentry } from '../../lib/sentry'
-import {
-  type BrowserOutputFileAccess,
-  createBrowserOutputFileAccess,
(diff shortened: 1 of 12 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1327, opened 2026-06-23
chore(deps): refresh package versions across workspaces

DaniAkash:
Routine `bun update` across the agent monorepo.

Quick follow-up tidy-ups: one hono request-param widening guard, a couple of eval callback shape cleanups, a shiki override to keep one version hoisted.

`bun run typecheck`, `bun fallow`, scoped mcp-manager + build-script tests all clean locally.


DaniAkash:
@greptileai

PR #1345, opened 2026-06-23
chore: document agent MCP UI API URL env

shadowfax92:
## Summary
- Documents the optional `VITE_BROWSEROS_AGENT_MCP_API_URL` override in `apps/agent-mcp-ui/.env.example`.
- Uses the same loopback `/cockpit` URL shape that the UI client accepts and falls back to.
- Updates the env-example header so defaults are described by consuming code path, not only `web-ext.config.ts`.

## Design
This keeps ownership in the existing agent MCP UI env example because the app's `dev` and `build:dev` scripts load `.env.development` from that package, and the UI reads this Vite-prefixed value through `import.meta.env`.

## Test plan
- [x] `git diff --check`
- [x] `bun run check`
- [x] `bun run test`
- [x] local Codex review round 2: clean

### DIFF
--- a/packages/browseros-agent/apps/agent-mcp-ui/.env.example
+++ b/packages/browseros-agent/apps/agent-mcp-ui/.env.example
@@ -2,6 +2,10 @@
 # loads these before wxt launches the browser. None are required —
 # every entry has a sensible default in web-ext.config.ts.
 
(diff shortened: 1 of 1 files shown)


shadowfax92:
@claude review

shadowfax92:
@greptileai review

PR #1358, opened 2026-06-24
fix: preserve CLI runtime structured output

shadowfax92:
## Summary
- Preserve `run` tool structured output over MCP HTTP by wiring optional browser tool output schemas through registration.
- Add a BrowserOS-backed CLI runtime test group that builds `browseros-cli` and runs it against `ensureBrowserOS()`.
- Cover the reported paths: `health --json`, `pages --json`, `active --json`, and `snap --json` without an explicit page.

## Design
The compact `run` tool already produced `{ ok, value, logs }` in process, but the real MCP HTTP path dropped dynamic `value` fields without an output schema. This change lets browser tools declare output schemas and gives `run` a permissive structured result shape so CLI commands can keep consuming `structuredContent.value` instead of parsing display text.

## Test plan
- [x] `cd apps/cli && gofmt -l .`
- [x] `cd apps/cli && go vet ./...`
- [x] `cd apps/cli && go build ./...`
- [x] `cd apps/cli && go test ./...`
- [x] `cd apps/server && bun --env-file=.env.development test --preload=./tests/__helpers__/test-env.ts ./tests/tools/browser/register.test.ts`
- [x] `cd apps/server && bun run ./tests/__helpers__/run-test-group.ts cli`
- [x] `cd apps/server && bun run typecheck`
- [x] `bun run check` (exited 0; reports existing unrelated Biome/Fallow warnings)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1363, opened 2026-06-24
fix: support scoped agent release tags

shadowfax92:
## Summary
- release the extension from `agent-extension/v*` tags and the server from `agent-server/v*` tags, using the pushed slash tag for GitHub Releases
- validate tag version, package.json version, default-branch reachability, and duplicate/non-incrementing versions across legacy and new component tags before publishing
- bump `@browseros/agent` to `0.0.100` and `@browseros/server` to `0.0.122`, with `bun.lock` kept consistent
- keep server nightly/manual non-release build behavior working, including `apply_bump` for `publish_github_release: false`

## Design
The workflows resolve release metadata from the default-branch checkout first, then detach to the validated release commit for install/build. Shared shell validation reads package versions from the tagged commit via `git show`, compares both legacy and slash tag schemes, and emits release metadata for notes and release publishing.

## Test plan
- [x] `actionlint .github/workflows/release-agent-extension.yml .github/workflows/release-server.yml`
- [x] `bun run lint` (exits 0; existing unrelated warnings remain)
- [x] `bun run typecheck`
- [x] `bun run test`
- [x] `bun run build:agent`
- [x] `bun test apps/server/tests/build.test.ts`


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1368, opened 2026-06-24
fix: remove MCP promo tool count

shadowfax92:
## Summary
- Removed the static `(66+ tools)` badge from the AI settings MCP promo banner.
- Added a static render regression test for the banner copy and setup action.

## Design
The MCP promo banner owns this copy directly, so the change removes only the static count span while preserving navigation, analytics, dismissal, and the rest of the banner content.

## Test plan
- [x] `bun run --filter @browseros/app test`
- [x] `bun run check`
- [x] `bun run test`
- [x] `bun run build:agent`


shadowfax92:
@claude review

shadowfax92:
@greptileai review

PR #1377, opened 2026-06-24
chore(cli): use structured MCP results

shadowfax92:
## Summary
- Use structured grep `matches` for direct and batch `find` instead of parsing wrapped text output.
- Use structured screenshot `image` data for file writes and structured JSON output.
- Remove the unused `run`-based page resolver from the Go MCP client.

## Design
The CLI already normalizes MCP `structuredContent` into `ToolResult.StructuredContent` and `output.JSON` prefers that structured map. This change moves command glue onto the new structured fields added in `b464fc17` while preserving the explicit page-id CLI contract and human output shape.

## Test plan
- [x] `gofmt -l .`
- [x] `go vet ./...`
- [x] `go build ./...`
- [x] `go test ./...`


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1382, opened 2026-06-25
chore: remove extra claude skills

shadowfax92:
## Summary
- Remove non-allowlisted root Claude skills, keeping only `ask-internal`, `write-docs`, and `writing-plans`.
- Remove non-allowlisted package Claude skills, keeping only `test-ui`.
- Update retained skill handoffs so they no longer reference deleted skill folders.

## Design
This is an allowlist cleanup of the two tracked `.claude/skills` trees. The retained skill docs were adjusted only where they pointed at removed skills, so the remaining surfaces stay self-contained.

## Test plan
- [x] `git diff --check`
- [x] `bun run check`
- [x] `bun run test`
- [x] local adversarial review, round 2 clean


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1389, opened 2026-06-26
fix: remove fake window chrome from claw-onboard, fill popup

shadowfax92:
## Summary
- The standalone `claw-onboard` app embeds directly into Chromium as the popup/page, so the simulated macOS-window shell was wrong. Removed the fake chrome: traffic-light title bar, centered "Welcome to BrowserOS" caption, and the fixed-size (`h-[720px] max-w-[1040px]`) rounded/shadowed floating card.
- Reframed `OnboardingShell` to a full-viewport flex row so the visual rail + step column fill the **full popup width and height**; the step column scrolls internally (`min-h-0 overflow-y-auto`) while the page itself never scrolls (`h-screen overflow-hidden`).
- Flattened the body background from the desktop-mimicking gradient (which only sat behind the fake window) to the flat canvas color, avoiding a load-time flash.

## Design
The fake shell was entirely contained in `OnboardingShell.tsx` as two outer wrapper layers (centering + fixed card) plus a title bar around the real content row. The change deletes those and promotes the rail + step column to the root, preserving the existing two-column layout, every step component, copy, and the rail unchanged — this is a framing change, not a redesign. Now that it's a full page (not a modal), `role="dialog"` is dropped and the scrollable step column is marked as the page's `<main>` landmark.

## Test plan
- [x] `bun test` (claw-onboard) — 32 pass / 0 fail; the old chrome-bar test is replaced with a guard asserting the `<main>` landmark and the absence of the dialog chrome (`role="dialog"`, the caption, the traffic-light color).
- [x] `bun run lint` + `bun run typecheck` — clean.
- [x] `bun run build` (tsc + vite) — builds clean.
- [x] Visual check via puppeteer at 1000×680 and 760×600 (Welcome + Import steps): no traffic lights, no title bar, no floating card; rail + content fill the popup; content scrolls internally.

## Notes
- `bun run check`'s **fallow** step reports a pre-existing monorepo-wide dead-code backlog (353 issues, all in untouched `apps/claw-app` & `apps/claw-server`; 3 suppressions already in place). This change adds **zero** fallow issues in `claw-onboard` — lint, typecheck, test, and build are all green.
- Out of scope (non-goals): rail redesign, responsive/mobile breakpoints, Chromium-side embedding wiring, and the rail's "Mac · v1.0 · signed build" footer.

### DIFF
--- a/packages/browseros-agent/apps/claw-onboard/src/onboarding/OnboardingV2.test.tsx
+++ b/packages/browseros-agent/apps/claw-onboard/src/onboarding/OnboardingV2.test.tsx
@@ -62,9 +62,10 @@ describe('OnboardingV2 shell', () => {
     expect(html).toContain('Under your control')
   })
 
-  it('renders the macwin chrome bar title', () => {
+  it('does not render the fake macOS window chrome', () => {
     const html = renderApp()
-    expect(html).toContain('Welcome to BrowserOS')
+    expect(html).not.toContain('Welcome to BrowserOS')
+    expect(html).not.toContain('#FF5F57')
   })
 
   it('renders four step dots', () => {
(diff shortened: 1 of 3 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1399, opened 2026-06-26
fix: open new tab after onboarding complete

shadowfax92:
## Summary
- Keep standalone BrowserClaw onboarding completion on the existing `browserosOnboardingComplete` Chromium bridge message.
- Replace the post-completion cockpit redirect with Chromium's canonical new-tab URL, `chrome://newtab`.
- Update the focused onboarding test to lock the new destination.

## Design
`OnboardingV2` still calls `bridge.complete()` before navigating away, so native onboarding completion stays at the Chromium bridge boundary. The only UI-side behavior change is that the finish action now leaves the standalone onboarding WebUI for the BrowserOS new-tab page instead of resolving and opening the cockpit API root. I used `chrome://newtab` rather than the requested `chrome://new-tab` because local BrowserOS launch configs and Chromium constants use `chrome://newtab` / `chrome://newtab/`, and no `chrome://new-tab` alias showed up in the local sources.

## Test plan
- [x] `bun test apps/claw-onboard/src/onboarding/OnboardingV2.test.tsx apps/claw-onboard/src/onboarding/browseros-onboarding-bridge.test.ts`
- [x] `bun run check` (exits 0; prints existing unrelated Biome/Fallow warnings)
- [x] `bun run test`
- [x] local Codex diff review, round 2 clean

### DIFF
--- a/packages/browseros-agent/apps/claw-onboard/src/onboarding/OnboardingV2.test.tsx
+++ b/packages/browseros-agent/apps/claw-onboard/src/onboarding/OnboardingV2.test.tsx
@@ -88,7 +88,7 @@ describe('OnboardingV2 shell', () => {
 
     openBrowserOsNewTab()
 
-    expect(getAssignedUrl()).toBe('chrome://new-tab')
+    expect(getAssignedUrl()).toBe('chrome://newtab')
   })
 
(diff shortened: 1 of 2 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1402, opened 2026-06-26
chore: add Claw server config sample

shadowfax92:
## Summary
- Add `apps/claw-server/config.sample.json` with the Claw JSON config shape.
- Point `apps/claw-server/.env.example` at copying the sample to `claw.config.json`.
- Tighten Claw config-file validation so unknown top-level keys fail instead of silently falling back to defaults.

## Design
The sample stays inside the Claw server app boundary and is validated through `loadClawConfig`, so future schema drift breaks the focused config test instead of leaving a stale sample behind.

## Test plan
- [x] `bun test apps/claw-server/tests/config.test.ts`
- [x] `bun run check`
- [x] `bun run test`


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1442, opened 2026-06-29
fix: repair server tag version releases

shadowfax92:
## Summary
- Add a server tag-release prepare step for pushed `agent-server/v*` tags so tag-first releases can repair a stale server package version before publishing.
- Keep strict release validation in `resolve-component-release.sh`, including semver, annotated tags, default-branch reachability, duplicates, monotonic versions, and final package/tag equality.
- Document the preferred pre-bump flow and the tag-first constraints in the server README and workflow summary.

## Design
Pushed server tags now run `prepare-server-tag-release.sh`. If the tagged commit already has the matching server package version, it delegates to the existing strict resolver. If the tag is newer than the package version and points at the current default-branch tip, it sets `apps/server/package.json` and `bun.lock`, commits the bump, recreates the annotated tag locally, and pushes the branch plus tag update atomically before running the strict resolver again. It refuses downgrades and older default-branch commits.

## Test plan
- [x] `bun test scripts/release/resolve-component-release.test.ts`
- [x] `python3 ../browseros/build/scripts/bump_server_version_test.py`
- [x] `actionlint ../../.github/workflows/release-server.yml`
- [x] `bash -n scripts/release/prepare-server-tag-release.sh scripts/release/resolve-component-release.sh`
- [x] `git diff --check`
- [x] `bun run check` (exits 0; existing Claw Biome/Fallow warning inventory still prints)
- [x] `bun run test`

## Notes
- The failed `agent-server/v0.0.123` run used the workflow from the old tagged commit, so this fixes future tag pushes. After this lands, retrying `0.0.123` means pushing the tag again at a commit that includes this workflow.

### DIFF
--- a/.github/workflows/release-server.yml
+++ b/.github/workflows/release-server.yml
@@ -74,6 +74,7 @@ jobs:
     needs: authorize_manual_dispatch
     if: |
       always() &&
+      github.event.deleted != true &&
       (github.event_name != 'workflow_dispatch' || needs.authorize_manual_dispatch.result == 'success')
     runs-on: ubuntu-latest
     outputs:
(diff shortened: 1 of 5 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1443, opened 2026-06-29
feat: browseros_product GN flag for BrowserOS/BrowserClaw identity

shadowfax92:
## Summary
- Add a build-time `browseros_product` GN arg (`"browseros"` | `"browserclaw"`, default `"browseros"`) so one source tree can ship as either product.
- Generate `buildflags.h` (`BROWSEROS_PRODUCT_BROWSEROS` / `BROWSEROS_PRODUCT_BROWSERCLAW`) and wrap it in `core/browseros_product.{h,cc}` (`GetProduct()` / `IsBrowserOSProduct()` / `IsBrowserClawProduct()`).
- Product identity is baked into the binary — runtime switches and field trials cannot flip it for shipped artifacts.

## Design
A single GN arg selects the product; a `buildflag_header` turns it into compile-time flags, and a thin C++ helper (with a `static_assert` that exactly one product is selected) hides the raw flags behind `browseros::Product`. Default stays `"browseros"`, so the normal build is unchanged. No dev runtime override switch was added — it would undercut the "identity is un-flippable at runtime" guarantee; it can be added later, gated off in official builds, if dev ergonomics need it.

Extracted as chromium patches under `chromium_patches/chrome/browser/browseros/`; `buildflags.gni` is registered under the `browseros-core` feature (the `core/` files are already covered by that feature's directory entry).

## Test plan
- [x] Default (`browseros`) build links green via the build lock; generated `buildflags.h` has `BROWSEROS_PRODUCT_BROWSEROS=1`, `..._BROWSERCLAW=0`.
- [x] `verify-patches.sh`: 5/5 byte-match + apply-check, 0 skipped.
- [ ] `browseros_product = "browserclaw"` flips the generated header and `GetProduct()` returns `kBrowserClaw`.

### DIFF
--- a/packages/browseros/build/features.yaml
+++ b/packages/browseros/build/features.yaml
@@ -7,8 +7,9 @@ features:
   browseros-core:
     description: "chore: browseros core infrastructure"
     files:
-      # Directory structure + core constants/prefs/switches
+      # Directory structure + core constants/prefs/switches + product identity
       - chrome/browser/browseros/BUILD.gn
+      - chrome/browser/browseros/buildflags.gni
       - chrome/browser/browseros/core/
       # Shared pref infrastructure (settings-page + metrics + importer)
       - chrome/browser/prefs/browser_prefs.cc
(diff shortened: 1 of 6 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1447, opened 2026-06-29
chore: remove unused shutdown route

shadowfax92:
## Summary
- Remove the unused `POST /shutdown` Hono route and delete its route factory.
- Drop the route-only `onShutdown` dependency from API route composition, HTTP server config, and application startup.
- Add route coverage asserting `/shutdown` is no longer mounted.

## Design
The cleanup removes the public route and the dead callback plumbing that only existed to service that route. Existing process signal handling remains unchanged, and `Application.stop()` keeps the non-zero signal exit behavior Chromium relies on for restarts.

## Test plan
- [x] `bun --env-file=apps/server/.env.development test --preload=./apps/server/tests/__helpers__/test-env.ts ./apps/server/tests/api/routes/index.test.ts`
- [x] `bun run check`
- [x] `bun run test:main`
- [x] `bun run test`


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1450, opened 2026-06-29
refactor(patches): drop agent-port switch, unify server config

shadowfax92:
## Summary
- Remove the unused `browseros-agent-port` switch (`kAgentPort`) — it had no live readers anywhere in the tree.
- Collapse the per-product managed-server config into a single `config.json` shape shared by BrowserOS and BrowserClaw.
- Drop `ServerConfigKind`, the `config_kind` fields, `ConfigKindToString`, and the two per-product JSON builders; one `BuildServerConfigJson` now serves both.

## Design
The prior BrowserClaw descriptor work introduced two near-identical config shapes (`server_config.json` full vs `claw_config.json` stripped) selected by a `ServerConfigKind` enum. This unifies them: both descriptors emit `config.json` and one builder produces the shape below. Genuinely product-specific descriptor fields are untouched — `product`, `log_name`, `bundle_dir` (`BrowserOSServer` vs `BrowserClawServer`), `binary_name` (`browseros_server` vs `browseros-claw-server`), `health_path` (`/health` vs `/system/health`), and `enable_updater` (`true` vs `false`). Product-specific resource roots are unchanged.

```json
{
  "ports":       { "cdp": ..., "server": ..., "proxy": ... },
  "directories": { "resources": "...", "execution": "..." },
  "flags":       { "allow_remote_in_mcp": false },
  "instance":    { "install_id": "...", "browseros_version": "...", "chromium_version": "..." }
}
```

## ⚠️ Claw config-loader follow-up
The Claw loader previously consumed only `ports.server`, `ports.cdp`, and `directories.resources` (the removed `BrowserClawConfigUsesStrictShape` test encoded that). The unified `config.json` now also sends Claw `ports.proxy`, `directories.execution`, `flags`, and `instance`. **This requires the `browseros-claw-server` parser to tolerate (ignore) unknown keys.** That parser is not in the Chromium tree and was not inspectable here. BrowserClaw is brand-new/unreleased, so nothing in production breaks today — but if the loader rejects unknown keys, a Claw-side parser change is required before Claw ships.

## Test plan
- [x] `autoninja -C out/Default_arm64 chrome` — green
- [x] `verify-patches.sh` — 5/5 byte-match + apply-check
- [x] Reworked `browseros_server_config_unittest.cc` (descriptors assert `config.json`; new `ServerConfigJsonIncludesSharedRuntimeKeys` regression guard)

### DIFF
--- a/packages/browseros/chromium_patches/chrome/browser/browseros/core/browseros_switches.h
+++ b/packages/browseros/chromium_patches/chrome/browser/browseros/core/browseros_switches.h
@@ -1,9 +1,9 @@
 diff --git a/chrome/browser/browseros/core/browseros_switches.h b/chrome/browser/browseros/core/browseros_switches.h
 new file mode 100644
-index 0000000000000..465149164d56b
+index 0000000000000..1080aa947bec2
 --- /dev/null
 +++ b/chrome/browser/browseros/core/browseros_switches.h
-@@ -0,0 +1,86 @@
+@@ -0,0 +1,83 @@
 +// Copyright 2024 The Chromium Authors
 +// Use of this source code is governed by a BSD-style license that can be
 +// found in the LICENSE file.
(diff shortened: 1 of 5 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1453, opened 2026-06-29
fix: upload claw server artifacts by default

shadowfax92:
## Summary
- Makes the normal Claw server build inherit the shared production upload default, matching `build:server` behavior.
- Keeps local-only verification explicit through `--no-upload`, including the existing `build:claw-server:test` path.
- Adds regression coverage for default upload, `--no-upload`, `--ci`, and the Claw artifact smoke build.

## Design
The Claw build script already uses `runProdResourceBuild` from `@browseros/build-server-tools`; upload behavior is resolved by `parseBuildArgs` and the product descriptor. The bug was the Claw descriptor's `defaultUpload: false`, which made a successful production build archive locally but skip R2 uploads. Removing that override lets Claw inherit the shared default while preserving explicit opt-outs.

## Test plan
- [x] `bun test apps/claw-server/tests/build.test.ts packages/build-server-tools/tests/cli.test.ts`
- [x] `bun run build:claw-server:test`
- [x] `bun run check`
- [x] `bun run test`

### DIFF
--- a/packages/browseros-agent/apps/claw-server/tests/build.test.ts
+++ b/packages/browseros-agent/apps/claw-server/tests/build.test.ts
@@ -2,7 +2,10 @@ import { afterAll, describe, it } from 'bun:test'
 import assert from 'node:assert'
 import { existsSync, readFileSync, rmSync } from 'node:fs'
 import { resolve } from 'node:path'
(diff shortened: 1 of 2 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1459, opened 2026-06-30
chore(ci): move component release resolver

shadowfax92:
## Summary
- Move the component release resolver from `packages/browseros-agent/scripts/release` to root `scripts/ci`.
- Update the extension release workflow to call `$GITHUB_WORKSPACE/scripts/ci/resolve-component-release.sh`.
- Keep the existing package-local resolver tests, pointed at the new CI script path.

## Design
The resolver remains the same executable Bash script; only its ownership path changes. The workflow uses `$GITHUB_WORKSPACE` so the root CI script remains explicit even though the step runs from `packages/browseros-agent`.

## Test plan
- [x] `bun test ./scripts/release/resolve-component-release.test.ts` from `packages/browseros-agent` - 18 pass
- [x] `bun run check` from `packages/browseros-agent` - pass after local `bun install --frozen-lockfile` and `bun run --filter @browseros/app codegen`; existing Biome/Fallow warnings are reported but non-fatal
- [x] `git diff --check HEAD~1..HEAD` - pass
- [x] `rg "scripts/release/resolve-component-release\.sh|packages/browseros-agent/scripts/release/resolve-component-release\.sh" .github packages scripts -S` - no active references
- [ ] `bun run test` from `packages/browseros-agent` - local macOS run fails in unrelated `scripts/build/server/native-addon-policy.test.ts` (`prevents Bun from opening hidden temp native addons` expects a stderr guard message but receives empty stderr); focused release tests pass, and the CI build matrix runs on Ubuntu where that assertion returns early.


shadowfax92:
@claude review

shadowfax92:
@greptileai review

PR #1464, opened 2026-06-30
fix: complete bpatch annotate feature pass

shadowfax92:
## Summary
- Make `bpatch annotate` process `features.yaml` in order with fresh checkout status for each feature.
- Normalize apply-created split index entries before committing so stale no-op paths are cleaned instead of failing with `commit feature ...:`.
- Report actual files from the created commit, including rename old/new paths, instead of pre-stage candidates.

## Design
The Go annotator now mirrors the maintained Python annotation pipeline: each feature is evaluated against the current checkout state, staged only when it still has matching changes, and committed only if staging creates a real HEAD delta. This removes the stale one-shot ownership snapshot that failed after the first feature commit.

## Test plan
- [x] `go test ./internal/engine -run 'TestAnnotate'`
- [x] `go test ./...` in `packages/browseros/tools/patch`
- [x] `gofmt -l .` in `packages/browseros/tools/patch`
- [x] `go vet ./...` in `packages/browseros/tools/patch`
- [x] `go build ./...` in `packages/browseros/tools/patch`
- [x] `git diff --check`
- [x] `uv run python -m unittest build.modules.annotate.annotate_test` in `packages/browseros`

### DIFF
--- a/packages/browseros/tools/patch/internal/engine/annotate.go
+++ b/packages/browseros/tools/patch/internal/engine/annotate.go
@@ -48,7 +48,6 @@ type annotateFeature struct {
 	Name        string
 	Description string
 	Files       []string
-	Index       int
 }
 
 type annotateFileSet struct {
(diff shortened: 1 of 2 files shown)


shadowfax92:
@claude review

shadowfax92:
@greptileai review

PR #1465, opened 2026-06-30
fix: warn when patches repo is stale

shadowfax92:
## Summary
- Add a startup warning when the configured or discoverable patches repo differs from `origin/main`.
- Print the drift count and patches repo path to stderr so JSON stdout stays parseable.
- Keep the check best-effort and non-fatal, including support for `add --patches-repo` overrides.

## Design
The root command pre-run now loads app state, resolves the relevant patches repo path, fetches `origin/main` with a short timeout, and compares `HEAD...origin/main`. Any fetch or inspection failure is ignored so patch commands keep working.

## Test plan
- [x] `cd packages/browseros/tools/patch && go test ./...`
- [x] `git diff --check`
- [x] Temporary local smoke test confirmed warning on stderr and normal command output on stdout.
- [x] Local read-only Codex review round 2: clean.

### DIFF
--- a/packages/browseros/tools/patch/cmd/root.go
+++ b/packages/browseros/tools/patch/cmd/root.go
@@ -111,7 +111,11 @@ Pass a checkout name to run from anywhere, for example "browseros-patch diff ch1
 		}
 		var err error
 		appState, err = app.Load(jsonOut, verbose, "")
-		return err
+		if err != nil {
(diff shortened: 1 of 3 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1467, opened 2026-06-30
feat: descriptor-driven OTA updater for both sidecar servers

shadowfax92:
## Summary
- Make the single `BrowserOSServerUpdater` serve both the BrowserOS and BrowserClaw sidecars, driven by `ManagedServerDescriptor` instead of hardcoded BrowserOS values.
- BrowserClaw now updates: its own appcast feeds, OTA state isolated under `.browseros/BrowserClawServer/`, `browseros-claw-server` binary, updater enabled.
- BrowserOS behavior is unchanged (legacy `.browseros` state, `/status` readiness, existing feeds).

## Design
`ManagedServerDescriptor` gains a nested `UpdaterConfig` (`state_dir`, `appcast_url`, `alpha_appcast_url`, `readiness_path`); the updater captures the selected descriptor at construction and reads appcast URL, execution/state dir, downloaded binary name, and readiness path from it. `enable_updater` stays a top-level flag the manager checks before starting an updater. It's nested (not flat) to keep the descriptor under the chromium-style implicit-constructor field-weight limit while remaining a `constexpr` aggregate. BrowserClaw's `readiness_path` is empty for now (it exposes `/system/health` but not the `{can_update}` JSON contract yet), so the updater skips the readiness fetch and restarts right after verification. Same Ed25519 key for both products; appcast URLs moved out of constants into the descriptor.

## Test plan
- [x] `browseros_server_config_unittest.cc` asserts both descriptors' fields (feeds, state dir, readiness path, `enable_updater`).
- [x] Full Chromium `chrome` target builds green (arm64), bundling both server resources.
- [x] Extracted patches verified: 6/6 byte-match vs source diff, 6/6 apply-check onto base tree.

### DIFF
--- a/packages/browseros/chromium_patches/chrome/browser/browseros/server/browseros_server_config.cc
+++ b/packages/browseros/chromium_patches/chrome/browser/browseros/server/browseros_server_config.cc
@@ -1,9 +1,9 @@
 diff --git a/chrome/browser/browseros/server/browseros_server_config.cc b/chrome/browser/browseros/server/browseros_server_config.cc
 new file mode 100644
-index 0000000000000..108ebf4b1cb5a
+index 0000000000000..6fe71062a9c96
 --- /dev/null
 +++ b/chrome/browser/browseros/server/browseros_server_config.cc
-@@ -0,0 +1,162 @@
+@@ -0,0 +1,178 @@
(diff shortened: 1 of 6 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1469, opened 2026-06-30
feat: add dogfood target flags

shadowfax92:
## Summary
- Add explicit global `--browseros/-b` and `--claw/-c` target selection for the dogfood CLI lifecycle surface.
- Split BrowserOS and BrowserClaw profiles, state, daemon sockets, locks, logs, and default ports.
- Add standalone Claw startup through `apps/claw-app` WXT plus `apps/claw-server --config`, with target-aware status/log/control messaging.

## Design
The CLI keeps lifecycle verbs at the root and resolves one global target before runtime commands execute. Common config stays shared, while per-target settings own runtime paths and ports. BrowserOS keeps the existing `apps/app` + `apps/server` path; Claw mirrors the existing `tools/dev --claw` stack and never routes through `apps/server`.

## Test plan
- [x] `cd tools/dogfood && gofmt -l . && go vet ./... && go build ./... && go test ./...`
- [x] `bun test apps/claw-server/tests/config.test.ts`
- [x] context-free local review clean after fixes

### DIFF
--- a/packages/browseros-agent/tools/dogfood/README.md
+++ b/packages/browseros-agent/tools/dogfood/README.md
@@ -1,6 +1,6 @@
 # browseros-dogfood
 
-Internal BrowserOS dogfooding CLI for running the current checkout against a copied BrowserOS profile.
+Internal BrowserOS dogfooding CLI for running BrowserOS or BrowserClaw against a copied BrowserOS profile.
 
 ## What It Does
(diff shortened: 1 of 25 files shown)


shadowfax92:
@claude review

shadowfax92:
@greptileai review

PR #1472, opened 2026-06-30
fix: default screenshot annotations off

shadowfax92:
## Summary
- Default omitted screenshot `annotate` to false in the MCP tool.
- Default omitted browser-core screenshot annotation options to plain capture.
- Updated screenshot tool, registration, and core capture tests for default-off behavior while preserving explicit `annotate: true`.

## Design
The MCP screenshot handler and browser-core helper are both defaulting boundaries, so both now treat omitted `annotate` as false. Explicit `annotate: true` continues through the existing overlay lifecycle and metadata path.

## Test plan
- [x] `bun test apps/server/tests/tools/browser/screenshot-tool.test.ts apps/server/tests/tools/browser/register.test.ts apps/server/tests/tools/browser/screenshot.test.ts`
- [x] `bun run check`
- [x] `bun run test`

### DIFF
--- a/packages/browseros-agent/apps/server/tests/tools/browser/register.test.ts
+++ b/packages/browseros-agent/apps/server/tests/tools/browser/register.test.ts
@@ -181,7 +181,7 @@ describe('registerBrowserTools', () => {
         format: 'jpeg',
         quality: 80,
         fullPage: false,
-        annotate: true,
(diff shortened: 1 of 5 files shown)


shadowfax92:
@claude review

shadowfax92:
@greptileai review

PR #1483, opened 2026-07-01
fix: avoid server release bump branch conflict

shadowfax92:
## Summary
- switch the server release reflection PR branch from `release/bump-server-v${VERSION}` to the flat `chore-bump-server-v${VERSION}` ref
- remove `continue-on-error` from the reflection step so post-release sync failures make the workflow red while the tag release remains authoritative
- recover reruns where the bump branch already exists but the PR was never created

## Design
The release publication path stays unchanged. Only the post-release reflection step now uses a flat branch name, writes a GitHub error/step summary on failure, and verifies an open PR exists before treating an existing remote branch as complete.

## Test plan
- [x] `bun test scripts/release/release-server-workflow.test.ts`
- [x] `bun run ./scripts/run-bun-test.ts ./scripts/release`
- [x] `actionlint ../../.github/workflows/release-server.yml`
- [x] `git diff --check`
- [x] `bun run check`
- [x] `bun run test`

## Review
- Local adversarial review round 1 found the branch-exists-without-PR rerun gap; fixed.
- Local adversarial review round 2 was clean.

### DIFF
--- a/.github/workflows/release-server.yml
+++ b/.github/workflows/release-server.yml
@@ -83,14 +83,11 @@ jobs:
               --notes-file /tmp/release-notes.md
           fi
 
-      # Reflect the released version back into the repo through a PR. Repository
-      # rules forbid pushing commits straight to the default branch, so the bump
-      # never touches it directly. This is best-effort (continue-on-error): the
-      # release above is authoritative, and releases are gated by tags, not by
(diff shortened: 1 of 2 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1494, opened 2026-07-01
feat: install newer bundled extensions over installed versions

shadowfax92:
## Summary
- Bundled CRXs now install when they are **newer** than the already-installed extension, instead of being skipped — previously users waited on the network OTA cycle (up to ~10-15 min: 60s initial maintenance delay + 15-min cycles + CDN lag) after an app update shipped a newer bundle.
- First-start installs kick off sooner: the 2s immediate-install delay is gone (plain `PostTask` — safe because `ExtensionUpdater::Start()` runs inside `ExtensionService::Init()` before `CheckForExternalUpdates()` triggers the loader).
- `TryLoadFromBundled()` no longer does blocking `PathExists` disk I/O on the UI thread during startup; the missing-manifest case is handled on the FILE-thread task with the existing remote-config fallback.

## Design
`InstallBundledExtensionsNow()` gains a version gate: skip if an install is already pending (dedupe against the external-provider path), skip if the installed version is same-or-newer, otherwise install/upgrade the bundled CRX via silent `CrxInstaller` with `install_immediately`. Downgrades are never installed. The pending-check ordering keeps the fast path idempotent with the `ExternalProviderImpl` file-found flow that runs at startup.

## Test plan
- [x] `autoninja -C out/Default_arm64 chrome` green on chromium `browseros` branch (squash `273d1e6e8f529`)
- [x] Patch extraction verified: 3/3 byte-match against `BASE_COMMIT..tip`, 3/3 apply-check onto bare BASE tree
- [ ] Manual: with an older extension version installed, launch a build with a newer bundled CRX and confirm log line `browseros: Upgrading <id> vX -> vY from bundled CRX` and the new version active within seconds

### DIFF
--- a/packages/browseros/chromium_patches/chrome/browser/browseros/extensions/browseros_extension_installer.cc
+++ b/packages/browseros/chromium_patches/chrome/browser/browseros/extensions/browseros_extension_installer.cc
@@ -1,9 +1,9 @@
 diff --git a/chrome/browser/browseros/extensions/browseros_extension_installer.cc b/chrome/browser/browseros/extensions/browseros_extension_installer.cc
 new file mode 100644
-index 0000000000000..8b976ef54e385
+index 0000000000000..ed44e6788a59e
 --- /dev/null
 +++ b/chrome/browser/browseros/extensions/browseros_extension_installer.cc
-@@ -0,0 +1,310 @@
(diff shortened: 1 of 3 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1495, opened 2026-07-01
fix: harden Chrome login importer on Windows (App-Bound v20)

shadowfax92:
## Summary
- Chrome 127+ encrypts Windows cookies with **App-Bound Encryption ("v20")**, whose key is sealed by Chrome's SYSTEM elevation service and cannot be decrypted by another application. The importer only understood **"v10"** and silently imported undecryptable "v20" blobs as **raw ciphertext** — corrupting the import.
- Detect, **skip, and report** v20 values (and genuine decrypt failures) instead of storing garbage. Passwords and v10 cookies keep importing on Windows.
- Windows: decrypt legacy unprefixed values via **DPAPI** before treating them as plaintext (mirrors `os_crypt_win.cc`).
- Copy the SQLite **`-wal` sidecar** with the DB so freshly written cookies/passwords (present when the source browser is running) aren't missed; share the copy/cleanup helper between the cookie and password importers.
- Fix Windows profile-dir path construction to convert UTF-8 profile ids.

## Design
Graceful degradation, **not** a v20 crack. Decrypting v20 requires SYSTEM privileges (the key is behind Chrome's elevation service, which byte-for-byte validates the caller's exe path — a fork is rejected). Even Brave (imports no cookies, disabled Chrome password import) and browser-use-desktop (disables its importer on Windows) punt on this. A new `DecryptResult` status lets the cookie/password importers skip-and-count v20 values; the Windows-only decryptor mirrors Chromium's `os_crypt_win.cc`. macOS behavior is unchanged.

Follow-ups (out of scope): Edge/Brave/Opera/Vivaldi sources; a user-facing "N cookies skipped" banner (currently logged); SYSTEM-based v20 decryption; OSCrypt-reuse refactor.

## Test plan
- [x] `autoninja -C out/Default_arm64 chrome` (macOS arm64) via the global build lock — green; all importer TUs recompiled + linked.
- [x] Patch verify: **9/9 byte-match + apply-check** onto BASE.
- [ ] Windows: import from Google Chrome — passwords + v10 cookies import; v20 cookies logged as skipped, not corrupted.
- [ ] macOS: import from Google Chrome — no regression.

_Note: `chrome_decryptor_win.cc` is `is_win`-gated and not compiled by the macOS build; reviewed manually against `os_crypt_win.cc`._

### DIFF
--- a/packages/browseros/chromium_patches/chrome/browser/importer/importer_list.cc
+++ b/packages/browseros/chromium_patches/chrome/browser/importer/importer_list.cc
@@ -1,5 +1,5 @@
 diff --git a/chrome/browser/importer/importer_list.cc b/chrome/browser/importer/importer_list.cc
-index 62546b572bab8..5e387e769912c 100644
+index 62546b572bab8..6ebad489e44bb 100644
 --- a/chrome/browser/importer/importer_list.cc
 +++ b/chrome/browser/importer/importer_list.cc
 @@ -6,10 +6,15 @@
@@ -26,7 +26,7 @@ index 62546b572bab8..5e387e769912c 100644
  
  #if BUILDFLAG(IS_MAC)
  #include "base/apple/foundation_util.h"
-@@ -29,6 +35,200 @@
+@@ -29,6 +35,203 @@
  
  namespace {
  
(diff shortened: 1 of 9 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1496, opened 2026-07-01
feat: browseros-version based update feed for sparkle/winsparkle

shadowfax92:
## Summary
- Sparkle (mac) and WinSparkle (win) now compare `10000.MAJOR.MINOR.BUILD.PATCH` — the BrowserOS version behind a fixed epoch — instead of chromium `BUILD.PATCH` inflated by the build offset.
- One string, one derivation: `Context.get_sparkle_version()` feeds the appcast for both platforms; macOS gets it stamped into `CFBundleVersion` before signing (`sign/macos.py`), Windows mirrors it in `winsparkle_glue.cc` (patch extracted from chromium `b2232c03b8bf8`).
- `chrome/VERSION` and `BROWSEROS_BUILD_OFFSET` are deliberately untouched on every platform — the Windows installer still needs a unique, monotonically increasing install version per release, and one uniform scheme beats a per-platform split. The updaters simply no longer read it.

## Design
Shipped fleets compare in the old offset space (mac appcast ≤ ~7948.97; win ≤ ~7949.97, live since Jun 20). Sparkle 2.7 deprecates custom comparators and enforces an install-time standard-comparator downgrade guard, so feed versions must sort above that space forever — bare semver (`0.47.x`) would strand every installed client. `10000.<version>` is the smallest encoding that clears the fence while keeping the BrowserOS version legible; users only ever see `0.47.0.2` (shortVersionString everywhere). The three lockstep sites cross-reference each other in comments. One consequence, documented in code: every release must bump the BrowserOS version — an offset-only bump is invisible to updaters.

## Test plan
- [x] `uv run python -m unittest discover -s build -p "*_test.py"` — 108/108 green
- [x] `verify-patches.sh` — 2/2 byte-match + apply-check against chromium tip `b2232c03b8bf8`
- [x] `autoninja -C out/Default_arm64 chrome` green in ch1 (winsparkle_glue.cc is win-only; first Windows CI build is its compile gate)
- [ ] Next release: appcast items carry `sparkle:version 10000.0.47.x` and shipped 0.46/0.47 clients are offered the update

### DIFF
--- a/packages/browseros/build/common/context.py
+++ b/packages/browseros/build/common/context.py
@@ -30,6 +30,28 @@ from .paths import get_package_root
 
 
 # =============================================================================
+# Update feed versioning
+# =============================================================================
+#
+# resources/BROWSEROS_VERSION is the single source of update identity. The
+# feed version is "10000.MAJOR.MINOR.BUILD.PATCH": carried in the appcast's
+# sparkle:version, stamped into CFBundleVersion before signing (what Sparkle
+# compares), and mirrored by chrome/browser/win/winsparkle_glue.cc for
+# WinSparkle. The fixed 10000 epoch sorts above the retired feed scheme
+# (chromium BUILD.PATCH inflated by BROWSEROS_BUILD_OFFSET, ~7950.97 at
+# cutover) so already-shipped clients keep seeing new releases as upgrades.
+#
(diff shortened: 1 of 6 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1497, opened 2026-07-01
fix(patches): open browser window when onboarding is dismissed

shadowfax92:
## Summary
- Closing/cancelling the BrowserOS onboarding window (or a crashed flow) left the user with no browser window at all — the app stayed open showing nothing.
- Any non-completion exit now marks onboarding done and opens the default new-tab window; completion behavior is unchanged.
- Quit-in-progress (`Cmd+Q`) and superseded/abandoned flows still skip the launch so we don't fight shutdown or double-launch.

## Design
`OpenNewWindowForBrowserOSOnboarding()` — the `ProfilePicker` first-run exit callback in `chrome/browser/ui/startup/startup_browser_creator.cc` — previously bailed for any status other than `kCompleted`, while window close delivers `kQuitAtEnd`. The fix mirrors upstream FRE semantics (`FirstRunService::OnFirstRunHasExited` treats `kQuitAtEnd` as proceed + mark-finished): on `kQuitAtEnd` we call `browseros::onboarding::MarkCompleted()` and then `LaunchBrowser()`. Marking must happen first because `LaunchBrowser()` re-checks `ShouldShow()` and would re-open the onboarding picker mid-teardown. `kAbortTask`/`kAbandonedFlow` and `browser_shutdown::IsTryingToQuit()`/`HasShutdownStarted()` still return without launching. Extracted from chromium squash commit `8e840137f3582` (ch2, branch `browseros`).

## Test plan
- [x] `autoninja -C out/Default_arm64 chrome` green on the chromium checkout
- [x] `verify-patches.sh`: 1/1 byte-match against `BASE_COMMIT..8e840137f3582`, 1/1 apply-check onto the bare BASE tree
- [ ] Manual: fresh profile → close onboarding via traffic-light → new-tab window opens; relaunch → onboarding does not re-show; completing onboarding still opens the browser with first-run tabs

### DIFF
--- a/packages/browseros/chromium_patches/chrome/browser/ui/startup/startup_browser_creator.cc
+++ b/packages/browseros/chromium_patches/chrome/browser/ui/startup/startup_browser_creator.cc
@@ -1,5 +1,5 @@
 diff --git a/chrome/browser/ui/startup/startup_browser_creator.cc b/chrome/browser/ui/startup/startup_browser_creator.cc
-index 597bd5bfdcbbf..9f4392215e04e 100644
+index 597bd5bfdcbbf..9b8998497b560 100644
 --- a/chrome/browser/ui/startup/startup_browser_creator.cc
 +++ b/chrome/browser/ui/startup/startup_browser_creator.cc
 @@ -39,6 +39,7 @@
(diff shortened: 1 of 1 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1498, opened 2026-07-01
fix: default third-party cookies to incognito only

shadowfax92:
## Summary
- Removes the BrowserOS patch that silently changed the third-party cookie controls default from `kIncognitoOnly` to `kBlockThirdParty`.
- Adds a regression patch asserting the registered default remains `kIncognitoOnly`.
- Updates `build/features.yaml` to track the regression-test patch instead of the stale source patch.

## Design
The final Chromium source should match upstream for `components/content_settings/core/browser/cookie_settings.cc`, so BrowserOS no longer carries a source patch for that file. The remaining patch adds test coverage in `cookie_settings_unittest.cc` to pin the incognito-only default while preserving explicit user/policy modes.

## Test plan
- [x] `/Users/shadowfax/.skills/sf-ch-mux/scripts/chpool build -- autoninja -C out/Default_arm64 chrome`
- [x] `/Users/shadowfax/.skills/sf-ch-extract/scripts/verify-patches.sh --chromium-src /Users/shadowfax/ch1-src --worktree /Users/shadowfax/worktrees/main/feat-thirdparty-cookies-incognito-patches --tip db693d466c326daeb535789c20ef79e934767962 components/content_settings/core/browser/cookie_settings_unittest.cc`
- [x] Confirmed `cookie_settings.cc` has no final Chromium diff from `BASE_COMMIT..db693d466c326daeb535789c20ef79e934767962` and the stale patch file is removed.


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1500, opened 2026-07-01
fix: ungate server resource staging

shadowfax92:
## Summary
- stage BrowserOS and BrowserClaw server resource downloads for every product while retaining OS/arch filters
- copy both server resource trees for every product so Chromium resource bundle targets can consume paths selected by GN flags
- update real-config tests for download/copy filtering across products and target platforms

## Test plan
- cd packages/browseros && uv run python -m unittest bos_build.steps.resources.resources_test bos_build.steps.storage.download_test
- cd packages/browseros && uv run python -m unittest discover -s bos_build -t . -p "*_test.py"

## Notes
- Adversarial review verdict: clean.
- Did not run autoninja because this patch does not change Chromium source code.


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1502, opened 2026-07-02
feat(bos_build): --show-plan, --skip, --from and modules profiles

shadowfax92:
## Summary
- `--show-plan` prints the composed step list + required env vars (marked set/missing, never values) for any preset/profile/product/arch — or a `--modules`/phase-flag pipeline — and exits without needing a chromium checkout. The projection is generated from `plan()`, so it can never drift.
- `--skip upload,series_patches` (CLI) and `skip:` (profile key) subtract steps **after** composition — commenting out a step as an operation. CLI and profile skips union; unknown names fail loudly listing valid steps; a valid-but-absent step is a no-op, so saved profiles inherit future preset changes automatically.
- `--from sign_macos` resumes the tail of the composed (post-skip) plan — re-run after a failure without recompiling. CLI-only, by design.
- Profiles may carry `modules: [...]` as an explicit enumerated-pipeline opt-in ("you own this list now") routed through the DIRECT-mode machinery; planner-owned keys and planner CLI flags are rejected beside it. Shipped profiles stay switch-based (drift-tested).

## Design
`plan()` stays the single source of truth; the file/CLI layer gets subtraction and visibility instead of enumeration. `skip` joins `Switches` (validated in `resolved()`, subtracted as `plan()`'s last act so composition rules never re-trigger — skipping `sign_windows` cannot resurrect `mini_installer`), `slice_from` is a planner helper the CLI applies, and `load_profile` returns a `Profile` carrier so modules-profiles can't masquerade as planner input. `--show-plan` exits before the banner, before chromium_src resolution, and before env preflight. An adversarial review verified `plan()` output for skip-free switches is byte-identical to base across a 324-combination switch/arch/platform matrix.

## Test plan
- [x] `uv run python -m unittest discover -s bos_build -t . -p "*_test.py"` — 333 tests green (46 new: skip/from goldens, Profile parsing + rejection rules, CliRunner projection tests that scrub `CHROMIUM_SRC` to prove the no-checkout property)
- [x] `uv run browseros build --preset release --skip upload --from configure --show-plan` with `CHROMIUM_SRC` unset — prints plan + env, exits 0
- [x] CI smoke now runs `--show-plan` on the checkout-less runner


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1503, opened 2026-07-02
chore(browseros): restructure bos_build into toolset layout

shadowfax92:
## Summary
- Re-house `packages/browseros/bos_build/` so the top level reads as three toolsets: **BUILD** (`steps/` on the `core/` engine), **RELEASE** (`release/` + `release/ota/`), **DEV** (`patchkit/`) — over shared plumbing (`lib/`) and product data (`products/`).
- Move-only: `git mv` + import rewrites + docstring/README path updates. No behavior change; history follows every file (`git log --follow`).
- Lands ahead of the feed-publisher and extension-release work so those are born in the right place.

## Moves
- `core/` → `lib/`: env, utils, logger, paths, notify, sparkle, versions (+test), testing (+test) — `core/` now holds only the engine (step registry, planner, runner, pipeline, resolver, context, events, product descriptor model)
- `core/server_binaries*` → `products/` (it is product data); `steps/storage/r2.py` → `lib/r2.py` (shared client used by release/ota/storage)
- `steps/release/` → `release/` (flattened); `steps/ota/` → `release/ota/`; `steps/` keeps only `@step` pipeline steps
- `core/__init__.py` exports engine names only; `EnvConfig`/`Notifier`/`get_notifier` moved to `lib/__init__.py`; `steps/storage/__init__.py` slimmed to step re-exports

## Design
Staged `git mv` in dependency order (plumbing → data/client → toolsets → docs), one green commit each, so rename detection stays clean and every intermediate state passes the full harness. Step registry contract untouched: `steps/__init__.py` import order = registration order, loaded via `import_module("bos_build.steps")`; release/ota modules were never registered steps. Lazy cycle-dodging imports keep their laziness (notify→events, env/logger→paths, server_binaries→products registry).

## Test plan
- [x] `uv run python -m unittest discover -s bos_build -t . -p "*_test.py"` → 352 tests OK after **every** commit
- [x] `uv run ruff check bos_build` clean after every commit
- [x] CLI smokes after every commit: `browseros build --list`, `product doctor`, `release --help`, `ota --help`, `dev extract --help`
- [x] Old-path sweep (dotted + slash forms, incl. mock.patch strings, workflows, docs): zero hits
- [x] Adversarial context-free review: clean apart from one stale docstring path, fixed


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1505, opened 2026-07-02
fix(browseros): product-aware release CLI with bounded listing

shadowfax92:
## Summary
- Thread `--product` through the whole release CLI (default browseros): `create_release_context` now resolves the product via the registry, and every `fetch_all_release_metadata` call site (list/appcast/download, joining publish/github) passes `ctx.product.id` — BrowserClaw releases are no longer invisible.
- Fix version listing under the productized R2 layout: `list_all_versions(release_prefix, env)` lists `releases/<prefix>/`, and `list_legacy_versions` surfaces pre-product bare `releases/<version>/` entries under browseros tagged `(legacy)` (dedupe prefers productized).
- Bound `browseros release list` to the newest 5 versions per product (both products by default, in short sections) with `-n/--limit` override and `--all` for everything; `release list <version>` keeps the product-aware artifact detail view.
- Convert the flag-soup callback (`--list/--appcast/--publish/--download`) into real typer subcommands: `release list|appcast|publish|download`, keeping `release github create` (+ `--product`, long-only since `-p` stays `--publish` there).
- Kill the last `if product.id == "browseros"` branch in `get_download_path_mapping` — the alias map is built purely from `product.artifact_prefix`, with a golden test pinning the browseros output byte-identical to the old constant.

## Design
Thin subcommands over product-parameterized helpers: the Step-module architecture stays (clean merge surface for the in-flight feed-publisher branch — appcast still only prints snippets, no feed publishing here). Listing splits into two composable R2 helpers sharing one paginated CommonPrefixes walker; merge/limit logic is pure and unit-tested without boto3. Legacy bare releases attach to the default product because they predate the registry; their detail view already works via `get_release_json`'s existing browseros fallback (untouched).

## Test plan
- [x] `uv run python -m unittest discover -s bos_build -t . -p "*_test.py"` — 384 tests OK (30 new: golden alias maps, prefix/delimiter/pagination listing, legacy shape filtering, merge/dedupe/limit, CLI surface incl. conflict + unknown-product errors)
- [x] `uv run browseros release list --help` / `release --help` / `build --list` / `product doctor` — all green
- [x] Live smoke against production R2: bounded sections with `… 112 more (use --all)` hint, browserclaw section, `(legacy)` tags, legacy version detail fallback


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1506, opened 2026-07-02
feat(browseros): update-feed publisher with safety rails

shadowfax92:
## Summary
- One owner for every update feed: a frozen `FeedSpec` table (`bos_build/release/feeds/spec.py`) covering all 13 client-polled R2 keys exactly once — browser appcasts, server appcasts (incl. the never-published `appcast-win-arm64.xml` and `appcast-claw-server*.xml`), and the extension manifests that were hand-edited in the api-worker repo until now. BrowserClaw *browser* feeds exist in the table but refuse `--publish` until the product-aware sparkle_glue chromium patch lands.
- A single rails-enforcing publisher (`feeds/publisher.py`) behind every write: well-formed check → channel title/link (or `external_update_url`) must match the spec → HEAD-200 on every enclosure/crx URL → downgrade guard vs the live object (fails closed on unparseable live; per-extension for gupdate manifests, incl. entry-removal refusal; `--allow-downgrade` to override) → backup of the live object to `feeds-history/<key>.<UTC ts>` **before** every PUT → local staging refresh (`bos_build/config/appcast/`, `updates/extensions/`). **Dry-run is the default everywhere**; `--publish` is the only writing path.
- CLI: `browseros release appcast --version X [--product P] [--publish]` (full-file generation replacing the old snippet printer), `browseros release extensions --channel alpha|prod --set name=ver … [--publish]` (regenerates update-manifest + extensions.json + bundled-manifest coherently, carry-over from live, two-phase preflight so a refusal writes nothing, bundled no-regress rule), and `browseros release feeds status` (live version + last backup per key).
- Server OTA aligned on the table: `ServerOTAModule` is bundle-aware via `--product` (browseros literals preserved byte-for-byte), the same-version platform merge now uses the **live** feed as merge base (stale checkout can't drop live platforms), `ota server release-appcast` goes through the rails, and a new `ota server promote` re-renders the live alpha item with prod title/link — fixing the live bug where prod `appcast-server.xml` is a byte-copy of the alpha file (title "(Alpha)", alpha link, verified live at v0.0.124). The metadata rail makes that byte-copy class impossible to publish again.

## Design
Descriptor-table approach mirroring `core/products.py`: `spec.py` owns what feeds exist, `render.py` produces complete single-item documents (golden-tested against today's live wire shapes), `publisher.py` is the one write path with injectable R2/HTTP for tests. Dependency direction is strictly `ota → feeds` / `cli → release → feeds`. CLI additions live in `cli/release_feeds.py` with a 2-line hookup in `cli/release.py` to minimize conflicts with the parallel per-product release-CLI work — at merge time keep both sides.

## Review
Three adversarial review rounds ran locally (see commit trail: two `fix(browseros)` hardening commits). No Critical findings at any round; every Important finding fixed with regression tests. Declined as not-worth-it, for the record: memoizing duplicate crx HEAD requests, sub-second backup-key granularity, and consolidating `github.py`'s informational snippet printer onto the new renderer (that surface belongs to the parallel release-CLI rework).

## Test plan
- [x] `uv run python -m unittest discover -s bos_build -t . -p "*_test.py"` — 444 tests green (golden XML/JSON per feed kind, downgrade + removal + metadata rails, backup-before-put ordering with mocked R2, extensions trio coherence + carry-over + preflight, spec-table exactly-once, ota bundle derivation + live-first merge base + promote re-render)
- [x] `uv run browseros release appcast --help` · `release --help` · `build --list` · `product doctor`
- [ ] Operator smoke after merge (human-only, never CI): `browseros release appcast --version <v>` dry-run against real R2 and diff-check vs live; `browseros release feeds status`

No R2 write happens in tests or CI — all R2/HTTP interaction is faked; `--publish` is for the human operator.


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1516, opened 2026-07-02
feat(browseros): --gn-arg overrides for configure

shadowfax92:
## Summary
- Adds a repeatable `browseros build --gn-arg key=value` passthrough, accepted in all three build modes (preset/profile, modules-profiles, and direct `--modules`/phase flags).
- `ConfigureModule` appends the values **last** in `args.gn` under a `# --gn-arg overrides` comment — GN last-write-wins makes them authoritative over the committed flags files, so iterating on `chromium_patches` no longer means editing `config/gn/*.gn` to flip `symbol_level`, `dcheck_always_on`, etc.
- Per-invocation by design: no profile key, never persisted. Overrides surface in `--show-plan` headers, the run banner, and a configure-time `Applying N gn-arg override(s)` log line; a plan without `configure` warns that the flag has no effect.

## Design
The CLI validates each value once (`^[a-zA-Z_][a-zA-Z0-9_]*=.+` — clear error naming the offending value) and threads a verbatim `tuple[str, ...]` through `Context.extra_gn_args`; both resolver paths (`resolve_config` for direct/modules-profile mode, the preset projection's `build_runs`) set it, and configure is the only consumer. Values pass through verbatim per GN syntax: bools/ints bare, strings with embedded quotes (`--gn-arg 'target_cpu="arm64"'`).

## Test plan
- [x] `uv run python -m unittest discover -s bos_build -t . -p "*_test.py"` — 489 tests green (new: append-last + override-ordering + log in configure_test; threading/default in resolver_test; validation, `--help` repeatability, 3× show-plan headers, preset + modules-profile `build_runs` plumbing in build_test)
- [x] `uv run browseros build --help | grep gn-arg` · `--list` · `product doctor`
- [x] Smoke: `--preset debug --gn-arg symbol_level=2 --gn-arg 'target_cpu="arm64"' --show-plan` shows the overrides header; `--gn-arg bogus` rejected with a clear error

### DIFF
--- a/packages/browseros/bos_build/README.md
+++ b/packages/browseros/bos_build/README.md
@@ -78,6 +78,9 @@ browseros build --preset release --skip upload,series_patches
 
 # Resume the tail after a failure without recompiling
 browseros build --preset release --from sign_macos
+
+# One-off GN overrides while iterating (appended last, so they win)
+browseros build --preset debug --gn-arg symbol_level=2 --gn-arg dcheck_always_on=true
 ```
 
 - `--skip` (and a `skip:` list in profiles) subtracts **after**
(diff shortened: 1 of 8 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1522, opened 2026-07-02
feat: add Claw app extension icon

shadowfax92:
## Summary
- Add the BrowserOS extension icon PNG assets to the Claw app public assets.
- Wire the Claw app action manifest to use the same 16/32/48/128 icon map as the BrowserOS app.
- Preserve the existing Claw extension title, permissions, and IDs.

## Design
The Claw WXT app now mirrors the BrowserOS app's public `icon/*.png` assets and uses the same `manifest.action.default_icon` paths. WXT packages those public assets into the extension root, which keeps this change app-local and avoids coupling the two app packages at runtime.

## Test plan
- [x] `bun run --filter @browseros/claw-app typecheck`
- [x] `bun run --filter @browseros/claw-app lint`
- [x] `bun run --filter @browseros/claw-app build`
- [x] `bun run check`
- [x] `bun run test`


shadowfax92:
@claude review

shadowfax92:
@greptileai review

PR #1526, opened 2026-07-02
fix: bake product identity buildflags for browseros/browserclaw release

shadowfax92:
## Summary
- Restores safe `buildflags.gni` defaults: `browseros_allow_runtime_product_override = false`, `browseros_package_all_server_resources` follows the override — resolving the pre-claw-release TODO from c840d957. Content is byte-identical to the pre-flip version (blob `0a4b34ea72336`), verified to apply cleanly.
- `get_product_gn_args()` now couples `browseros_package_all_server_resources` to the runtime override: debug builds get `true`/`true` (one debug binary can actually run as either product — previously the override was on but the other product's server resources were missing), release builds bake `browseros_product` with both flags `false`. Release args.gn output is unchanged from current main.
- Configure now warns per missing server-resource bundle on debug builds, so an unstaged tree fails with an actionable message at configure time instead of a cryptic ninja missing-input error.

## Design
Release safety lives in two layers on purpose: bos_build states managed-build policy explicitly in args.gn (auditable, belt-and-suspenders for shipped artifacts), while the .gni default protects anything built outside bos_build. The Jun 30 testing need (run either product from one build) is preserved through the managed debug path instead of unsafe global defaults.

## Notes for dev workflows
- Hand-managed args.gn (not written by `browseros build` configure) must add `browseros_allow_runtime_product_override = true` to keep the `--browseros-product` switch; the flipped default compiles it out otherwise. This is the intended pre-claw-release behavior.
- Existing debug out dirs pick up the corrected flags on their next configure run; plans that skip configure keep their frozen args.gn (true of every GN arg).
- Review triage: declined adding a GN assert against `override && !package_all` (legitimate for browser-only cross-product testing without server resources) and declined dropping the explicit package_all arg from args.gn (see Design).

## Test plan
- [x] `uv run python -m unittest discover -s bos_build -t . -p "*_test.py"` — 596 tests OK (new: release golden args for both products, debug wiring test, configure warning tests)
- [x] Edited patch applies cleanly in a scratch repo; `git hash-object` of the result matches the embedded index hash
- [ ] Next nightly debug build: confirm both `BrowserOSServer` and `BrowserClawServer` resources land in the output and `--browseros-product=browserclaw` works end to end

### DIFF
--- a/packages/browseros/bos_build/core/context.py
+++ b/packages/browseros/bos_build/core/context.py
@@ -296,13 +296,8 @@ class Context:
         return [base / "common", base / "products" / self.product.id]
 
     def get_product_gn_args(self) -> list[str]:
-        """Return BrowserOS product GN args for configure.
-
-        Release bakes product identity: runtime override off, only the
-        baked product's server resources packaged. Debug allows the
-        --browseros-product runtime switch, so it packages both products'
-        server resources — the switch is useless without them.
-        """
+        """Product GN args: release bakes identity; debug keeps the runtime
+        product switch working (it needs both server resource sets)."""
         dev_build = "true" if self.build_type == "debug" else "false"
         return [
             f'browseros_product = "{self.product.gn_product}"',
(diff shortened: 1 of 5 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1528, opened 2026-07-02
feat(claw-server): 24h rotating file logger

shadowfax92:
## Summary
- Adds a file sink to the claw-server logger so prod info logs on execution are persisted to `<browserosDir>/claw-server/claw-server.log`, mirroring the 24h startup rotation `apps/server` uses (stale log renamed to `.old`).
- Rotation keys on file creation time (mtime refreshes on every write and would never rotate an active log) with a 20MB size backstop for filesystems that don't track birthtime.
- stderr JSON-line output is unchanged; the file sink is dep-free (`node:fs`, sync per-line writes) — no pino, avoiding its Bun-compile transport caveats.

## Design
The existing minimal console-wrapper logger gains `setLogFile(dir)`: mkdir, rotate-if-stale, open an append fd; every log line is written to both stderr and the file. All failure paths degrade instead of crashing — a broken log dir keeps stderr-only logging, a failed re-point keeps the previous sink, a dead fd detaches the sink, and unserializable fields fall back to a stub line. `main.ts` attaches the sink only after `Bun.serve` binds the port, so a second accidental launch dies on EADDRINUSE before it can rotate the live instance's log, and `start()` now logs fatal startup errors (with stack) before exiting. Rotation renames first and only unlinks the old backup for genuine dest-exists failures, so history is never destroyed without enabling the rename.

Three review rounds ran locally (multi-agent); all confirmed findings fixed. Accepted tradeoffs: sync `writeSync` per line (low volume, crash-safe); rotation helper intentionally not extracted to `@browseros/shared` (implementations now diverge — follow-up if server adopts the same keying); cross-port multi-instance rotation needs a real file lock and is documented as a limitation.

## Test plan
- [x] `cd apps/claw-server && bun test` — 383 pass (13 logger tests: sink writes, dir creation, fresh-file append, stale rotation, creation-time vs mtime keying, size-cap rotation, backup replacement, unwritable dir, failed re-point keeps sink, re-point switches files, envelope precedence + key order, circular fields)
- [x] `bun run check` at monorepo root — exit 0
- [x] E2E: booted claw-server against a temp `BROWSEROS_DIR`; `claw-server listening` lands in the log file; a second launch on the held port exits 1 with a structured error and does not rotate the live log

### DIFF
--- a/packages/browseros-agent/apps/claw-server/src/lib/logger.ts
+++ b/packages/browseros-agent/apps/claw-server/src/lib/logger.ts
@@ -10,18 +10,26 @@
  * log views render both producers identically.
  *
  * `setLogFile` adds an optional file sink with startup-time rotation
- * (rename to `.old` when the file was created over 24h ago) so prod
- * runs keep an on-disk record. Deliberately dep-free: pino's async
- * transports bring Bun-compile caveats, and at this log volume sync
- * per-line writes are fine — and survive crashes, which is when the
- * file matters most.
+ * (rename to `.old` when the file was created over 24h ago or grew
+ * past the size cap) so prod runs keep an on-disk record. Deliberately
+ * dep-free: pino's async transports bring Bun-compile caveats, and at
+ * this log volume sync per-line writes are fine — and survive crashes,
+ * which is when the file matters most.
+ *
+ * Known limitation: rotation assumes one claw-server per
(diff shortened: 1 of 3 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1529, opened 2026-07-02
fix(claw-server): version binary from its own package.json

shadowfax92:
## Summary
- `browseros-claw-server --version` printed the BrowserOS server version (`0.0.124`) instead of the claw package version (`0.0.1`); the build now stamps the binary and `artifact-metadata.json` from `apps/claw-server/package.json`.
- Removes the vestigial `versionPackageDir` mechanism from `build-server-tools` — it was added in 189679fc for a release flow that built claw artifacts under the server release train, but later reworks removed claw from `release-server.yml`, leaving the coupling as a pure mislabel. Claw was its only user.
- Replaces the hardcoded `SERVER_VERSION = '0.0.1'` in both MCP entry points with the shared build-injected `VERSION`, so the MCP handshake reports the real artifact version.

## Design
Every product's build version now comes from its own `packageDir/package.json` (`readPackageVersion` in `build-server-tools/src/config.ts`), which is what the server and CLI products already did implicitly. The `apps/server` build is behavior-identical since it never set `versionPackageDir`. Regression coverage: a fast unit test pins `loadBuildConfig(rootDir, clawServerBuildProduct).version` to the claw package version (fails in ms if the descriptor ever points elsewhere), and the existing build smoke test now asserts binary `--version` + metadata against `apps/claw-server/package.json`.

## Test plan
- [x] Fast regression test red before the descriptor fix (`0.0.124 !== 0.0.1`), green after
- [x] `bun test apps/claw-server` — 372 pass, 0 fail (includes both build smoke tests)
- [x] `bun test packages/build-server-tools` — 17 pass, 0 fail
- [x] `bun run check` (lint + typecheck + fallow) — exit 0
- [x] End-to-end: built `darwin-arm64` binary prints `0.0.1`; `artifact-metadata.json` version is `0.0.1`

### DIFF
--- a/packages/browseros-agent/apps/claw-server/src/mcp/manager.ts
+++ b/packages/browseros-agent/apps/claw-server/src/mcp/manager.ts
@@ -22,10 +22,10 @@ import { McpServer } from '@modelcontextprotocol/sdk/server/mcp.js'
 import { WebStandardStreamableHTTPServerTransport } from '@modelcontextprotocol/sdk/server/webStandardStreamableHttp.js'
 import { logger } from '../lib/logger'
 import { findBySlug } from '../routes/agents/service'
+import { VERSION } from '../version'
 import { registerBrowserTools } from './register'
 
 const SERVER_NAME = 'browseros-claw-server'
-const SERVER_VERSION = '0.0.1'
 
 /**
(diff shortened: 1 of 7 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1531, opened 2026-07-02
fix(bos_build): isolate BrowserClaw user-data dir from BrowserOS

shadowfax92:
## Summary
- BrowserOS and BrowserClaw both shipped with `CrProductDirName = BrowserOS`, sharing `~/Library/Application Support/BrowserOS` and its SingletonLock — launching one blocked the other. Same hardcoding existed for Linux (`browser-os`) and Windows (`kProductPathName = L"BrowserOS"`).
- `ProductDescriptor` now models the user-data directory identity per platform: mac `product_dir_name` (BrowserOS / BrowserClaw), linux `user_data_dir_name` (browser-os / browser-claw), windows `product_path_name` (BrowserOS / BrowserClaw).
- New `product_identity` prep step runs right after `patches` and stamps those values into the three patched Chromium files; wired into release, debug, and universal presets plus phase-flag mode.

## Design
The chromium patches stay product-neutral (they carry the default BrowserOS values); per-product deviation happens in bos_build, same as the `chromium_files/products/<id>/` overlays. The step rewrites via structural anchors (the `CrProductDirName` key, the `#else`-guarded `data_dir_basename` assignment, the `kProductPathName` variable) — never the current value — so it is idempotent, and each anchor must match exactly once or the build dies (a stale patch can never silently ship a shared profile root). For `--product browseros` the stamp is a byte-level no-op, so default trees stay identical to the patch stack and BrowserOS artifacts are unchanged. Buildflag conditionals inside the patches were rejected because `chrome/common`/`chrome/install_static` can't depend on `chrome/browser/browseros:buildflags` (GN layering) and the plist is static text.

## Test plan
- [x] `uv run python -m unittest discover -s bos_build -t . -p "*_test.py"` — 601 tests green (goldens for both descriptors, stamp/no-op/idempotence/anchor-failure step tests, planner + phase ordering)
- [x] Anchors validated against the real patch stack content (exactly one match per file)
- [x] `uv run browseros build --preset release --product browserclaw --show-plan` lists `product_identity` right after `patches`
- [ ] Next claw build: confirm `BrowserClaw.app/Contents/Info.plist` has `CrProductDirName = BrowserClaw` and both browsers run side by side

## Follow-ups (out of scope)
- Windows registration identity (ProgIDs/CLSIDs/app GUID in `chromium_install_modes.h`) is still shared — needed for side-by-side Windows *installs*, not for running two profiles.
- `resources/entitlements/Info.plist.additions` is dead (referenced nowhere) and still says BrowserOS — candidate for deletion.
- A patch-stack lint could assert the three patches keep carrying the default product's values, guarding the extract workflow.


shadowfax92:
@greptileai review

shadowfax92:
@claude review

shadowfax92:
Note on the red `bos_build / unittest + ruff` check: the 3 failures (`test_help_documents_repeatable`, `test_help_documents_flags_and_read_only_guarantee`, `test_list_help_shows_bounding_options`) are pre-existing help-text assertions that fail on every PR since ~03:00Z today (7 consecutive unrelated branches) — rich/typer renders ANSI codes inside flag names under Actions, so plain-substring asserts like `'--product' in output` miss. The identical command (`uv run python -m unittest discover -s bos_build -t . -p '*_test.py'`) is green locally at this exact merged tree: 605/605 + ruff clean. Trunk CI needs a follow-up (e.g. `NO_COLOR: 1` in the workflow env or ANSI-stripping in the help tests).

PR #1534, opened 2026-07-02
feat: product-aware Sparkle/WinSparkle update feeds for BrowserClaw

shadowfax92:
## Summary
- Sparkle (mac) and WinSparkle (win) now select the update appcast by `browseros::GetProduct()`: BrowserClaw builds poll `appcast-claw[.-*].xml`; BrowserOS keeps today's keys.
- On Windows, the WinSparkle registry path (`Software\BrowserClaw\WinSparkle`) and app display name are product-scoped so side-by-side BrowserOS/BrowserClaw installs don't cross-contaminate skipped-version / last-check state.
- Release pipeline: browserclaw browser feeds are now **publishable** — the gate was waiting on exactly this client. Publishability is keyed on a new `_BROWSER_FEED_CLIENTS` set (the products whose updater selects their feed).

## Design
The feed key already existed on both sides — `_BROWSER_FEED_SLUGS = {"browseros": "", "browserclaw": "claw"}` in `release/feeds/spec.py` derives `appcast-claw*.xml`, and the chromium glue now emits those same keys. macOS picks the arch suffix at runtime (universal binary); Windows picks it at compile time (single-arch) and the product half from `GetProduct()`. The `publishable` gate stays as a general safety valve for any future product added to the slug map before its chromium updater client lands; browserclaw simply now qualifies.

## Test plan
- [x] `autoninja -C out/Default_arm64 chrome` — green (chromium patch compiles)
- [x] `verify-patches.sh` — both patches byte-match `BASE..tip` and apply-check onto the bare base
- [x] `python -m unittest` release (174) + products (29) suites green; feed/appcast/publisher tests updated for the new publishable state
- [ ] Cut a BrowserClaw build and confirm the updater polls `appcast-claw.xml` (mac) / `appcast-claw-win.xml` (win)

### DIFF
--- a/packages/browseros/bos_build/release/appcast_test.py
+++ b/packages/browseros/bos_build/release/appcast_test.py
@@ -2,12 +2,15 @@
 """Tests for the full-file browser appcast module."""
 
 import unittest
+from dataclasses import replace
 from types import SimpleNamespace
 from typing import cast
+from unittest.mock import patch
 
 from ..core.context import Context
 from ..core.step import ValidationError
 from .appcast import AppcastModule
+from .feeds.spec import browser_feeds_for_product
 
 
 def _artifact(filename: str) -> dict:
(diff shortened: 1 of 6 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1539, opened 2026-07-03
feat(patches): bpatch apply auto-annotates into feature commits

shadowfax92:
## Summary
- `bpatch apply` now runs `annotate` automatically after a conflict-free apply, so the common `apply && annotate` flow is one command. `--no-annotate` opts out; a filtered (`-- files`) or `--changed` apply never auto-annotates (they're surgical). `continue`/`skip` finish the annotation a paused apply asked for.
- Fixes the intermittent "`apply && annotate` doesn't work" cases: `annotate` now **refuses to run mid-conflict-resolution** (was committing half-applied trees + reject files into feature history), filters untracked junk (`*.rej`, `*.orig`, `.browseros-patchignore` patterns) the same way `extract` does, and reports changed files no feature claims as **unclaimed** instead of silently leaving the checkout dirty.
- Hardening from three adversarial review rounds: skipped conflicts (including a rename's old path) are excluded from auto-annotate; `apply`/`sync`/`annotate` refuse on unmerged index entries **or** leftover conflict markers; a parked `--no-rebase` stash still annotates (pure-patch tree); `resolve.Save` is atomic; annotate failures are recorded on the result (`annotate_error`) so the apply outcome is never discarded.

## Design
`ApplyOptions.AutoAnnotate` drives a shared completion step (`finishApply` for the no-conflict path, `finishResolvedRun` for `continue`/`skip`) that persists the intent through a conflict pause via `resolve.State`, so the whole flow finishes what `apply` started. Annotate takes one `git status` snapshot and partitions it per-feature in memory (no N+1 rescans). The one guard that matters for correctness — don't commit a conflicted/half-merged tree — lives in a single shared `requireNoPendingResolution` helper (pending resolve state, unmerged entries, or `git diff --check` conflict markers) that `apply`, `sync`, and `annotate` all call. Default-on annotation is per the explicit request ("apply should automatically do annotate as that is the most common way"); `--no-annotate` is the escape hatch.

## Test plan
- [x] `go test ./...`, `go vet ./...`, `make test && make build && make clean` in `packages/browseros/tools/patch` — all green
- [x] New engine tests: auto-annotate after clean apply, no-op apply still annotates, refuse during resolution / on unmerged files / on conflict markers, skip finishes annotation excluding the half-applied conflict, rename old-path exclusion, parked-clean stash still annotates, missing-registry skip, annotate-error recorded not fatal, unclaimed reporting, junk filtering
- [x] End-to-end against scripted fake patch-repo + checkout: fresh apply+annotate, idempotent re-run, patch-repo update, genuine conflict → skip → completion, unmerged-guard block+recover

### DIFF
--- a/packages/browseros/tools/patch/cmd/annotate.go
+++ b/packages/browseros/tools/patch/cmd/annotate.go
@@ -69,6 +69,13 @@ func printAnnotateResult(ws workspace.Entry, result *engine.AnnotateResult) {
 			fmt.Printf("  %-24s %s\n", skipped.Name, skipped.Reason)
 		}
 	}
+	if len(result.Unclaimed) > 0 {
+		fmt.Println(ui.Warning("Unclaimed changes (no feature owns these; left uncommitted):"))
+		for _, rel := range result.Unclaimed {
+			fmt.Printf("  %s\n", rel)
+		}
+		fmt.Println(ui.Hint(`Claim them in bos_build/features.yaml, then re-run; "browseros-patch feature lint" checks coverage.`))
+	}
 	if result.CommitsCreated == 0 {
 		fmt.Println(ui.Hint("No commits created."))
 	}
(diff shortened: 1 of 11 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1544, opened 2026-07-03
fix: inline production node env for claw server

shadowfax92:
## Summary
- Force Claw production resource builds to inline `NODE_ENV=production`.
- Add an optional build descriptor inline-env override so ambient `NODE_ENV=development` cannot leak into Claw bundles.
- Cover Claw descriptor behavior and shared build-config override precedence with focused Bun tests.

## Design
The Claw descriptor now declares `NODE_ENV` as an inline env key and owns a production override. `loadBuildConfig` applies optional descriptor inline overrides after file/process env and CI defaults, leaving existing products unchanged unless they opt in.

## Test plan
- [x] `bun test scripts/build/claw-server/descriptor.test.ts`
- [x] `bun test scripts/build/claw-server/descriptor.test.ts packages/build-server-tools/tests/config.test.ts`
- [x] `bun run check` from `packages/browseros-agent` (exits 0; existing Biome/fallow warning reports still print)

### DIFF
--- a/packages/browseros-agent/packages/build-server-tools/src/config.ts
+++ b/packages/browseros-agent/packages/build-server-tools/src/config.ts
@@ -101,6 +101,7 @@ export function loadBuildConfig(
       envVars[key] ??= value
     }
   }
+  Object.assign(envVars, product.env.inlineEnvOverrides ?? {})
   if (!options.ci) {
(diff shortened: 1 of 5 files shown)


shadowfax92:
@claude review

shadowfax92:
@greptileai review

PR #1547, opened 2026-07-04
refactor(claw): drop legacy /mcp/:slug, rename bind port to serverPort, canonicalize MCP URL

shadowfax92:
## Summary
- **Removed the legacy per-agent `/mcp/:slug` MCP route** and its `COCKPIT_LEGACY_PER_AGENT_MCP` env gate. The slugless `/mcp` (mcp-v2) route is now the only MCP HTTP endpoint. Deleted the dead `src/mcp/manager.ts` and legacy tests.
- **Renamed the claw-server bind port `env.port` → `env.serverPort`** (config field `serverPort`, mapped from sidecar `ports.server`) to make the server/proxy split explicit, matching the `apps/server` mental model. The server still listens only on `127.0.0.1:${serverPort}`.
- **Canonicalized every generated MCP URL to the slugless `/mcp`**, preferring the Chromium-owned proxy port for public URLs (`http://127.0.0.1:${proxyPort ?? serverPort}/mcp`) via a single shared `publicMcpUrl()` helper. Startup migration, agent profiles, harness installs, and the claw-app UI helpers all flow through it.

## Design
Chromium owns the MCP proxy port and forwards it to the sidecar's server port; claw-server must never *bind* the proxy port, only read it to advertise the public URL external clients should use. This PR keeps that split intact — API calls stay on the server port, MCP URLs prefer the proxy port — while deleting the legacy per-slug URL shape/route/migration that produced URLs pointing at a now-removed route. The three previously-divergent copies of the public-URL policy were consolidated into `shared/mcp-url.ts::publicMcpUrl()`.

A round of adversarial code review also caught a **fail-open bug in the one-shot startup migration**: it rewrote the profile JSON before the harness re-install and discarded the install outcome (which returns `{ installed: false }` rather than throwing), so a single transient re-install failure would permanently strand a user's harness entry with no retry. Fixed by reordering to install-then-persist: the profile URL only advances after the harness entry is in place, and failed rows retry on the next boot.

## Test plan
- [x] `cd packages/browseros-agent/apps/claw-server && bun test` — 385 pass / 0 fail
- [x] `cd packages/browseros-agent/apps/claw-server && bun run typecheck` — clean
- [x] `cd packages/browseros-agent/apps/claw-app && bun test` — 165 pass / 0 fail
- [x] `cd packages/browseros-agent/apps/claw-app && bun run typecheck` — clean
- [x] `rg` confirms no active `/mcp/:slug`, `COCKPIT_LEGACY_PER_AGENT_MCP`, or per-slug MCP URL generation remains
- [x] New test: migration leaves stored `mcpUrl` untouched and retries when a harness re-install fails

## Notes for reviewers
Two behaviors are intentional consequences of the shared slugless endpoint, not regressions in this PR:
- **"Regenerate URL" / agent delete no longer revoke access.** The v2 `/mcp` endpoint is a single shared URL with no per-agent identity gate; on `main` the per-slug route was already default-OFF (404) so per-slug revocation was already inert. This PR changes URL pointers, not the (pre-existing) enforcement model.
- **Per-agent Ask/Block approvals + site rules are enforced by the v2 dispatch path, not the deleted per-slug server.** The v2 single-server path (`register.ts`) documents the deliberate design; this cleanup does not change it.


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1557, opened 2026-07-04
feat: ship claw-onboard dist via R2 into the chromium build

shadowfax92:
## Summary
- Gives claw-onboard the claw-server treatment end to end: `bun run build:claw-onboard:resources` builds the chromium dist (`build:chromium`, contract-verified), stages it as a single platform-independent artifact with `artifact-metadata.json`, zips it, and uploads to R2 `claw-onboard/prod-resources/{latest|<version>}/browseros-claw-onboard-resources.zip`.
- `bos_build` downloads that zip (sha256-verified, ungated op in `download_resources.yaml`) to `resources/binaries/browseros_claw_onboard` and copies it into `chrome/browser/browseros/onboarding/resources` — the directory the existing onboarding grit pak already compiles and `chrome://`-serves.
- Retires the hand-committed built bundle: the three `chromium_patches/.../onboarding/resources/*` patch files (a stale 325KB minified app.js among them) are deleted, `features.yaml` pruned to match (repo doctor enforces the pairing), and the onboarding `BUILD.gn` grd list gains the `icon/*.png` files the dist contract requires. `nightly-macos-build.yml` stages the artifact hermetically like it does for claw-server.

## Design
claw-server's `runProdResourceBuild` is a compiled-binary pipeline (bun-compile per target, codesign, executable staging), so claw-onboard does **not** reuse it. Instead `build-server-tools` gains an assets-only sibling — `AssetBuildProductDescriptor` + `runProdAssetBuild` — that runs the app's own `build:chromium`, stages the dist under `<distRoot>/universal/resources`, and reuses the existing config/R2/metadata/zip helpers, so flags (`--ci`/`--no-upload`), env gating, metadata schema, and the `latest/`+versioned key layout stay identical to the binary flow. Because static assets are platform-independent, one universal zip replaces five per-target zips (one download op, no os/arch gates — the onboarding pak builds for every product, mirroring the claw-server ops which are likewise product-ungated). The copy feeds the **existing grit consumer** rather than a claw_server-style `bundle_data` dir, because the pak is the only runtime consumer; a parallel loose-file dir would ship dead bytes. `SetupWebUIDataSource` serves the full grit resource map, so adding icons needed no C++ change, and the resource-ID allotment (20) fits the 8 files. Deliberately **not** modeled as a `ServerBundle`: that abstraction requires signing metadata and forces OTA appcast feeds via `_server_feeds` — wrong for build-time-bundled static assets.

## ⚠ Post-merge action required (one-time seeding)
`chromium_patches` no longer carries the onboarding resources, so any `browseros build` with downloads enabled will look for `claw-onboard/prod-resources/latest/browseros-claw-onboard-resources.zip` in R2. **Immediately after merging, someone with R2 credentials must run `bun run build:claw-onboard:resources` once** (from `packages/browseros-agent`) to seed the artifact — before the next `nightly-release` (linux/windows) run. `nightly-macos-build` is unaffected (it stages the artifact locally via the step added here). This is the same bootstrap claw-server's download entries went through.

## Test plan
- [x] `packages/browseros-agent`: `bun run check` green; full `bun run test` green across all suites (server 1016, claw-app 171, claw-onboard 48 incl. a new e2e that runs the real resources build with `--no-upload` and asserts staged tree + metadata integrity + zip, claw-server 385, agent 282, eval 91, build scripts 31, release 33). The only failure anywhere is `scripts/build/server/native-addon-policy.test.ts` ("prevents Bun from opening hidden temp native addons"), which fails identically on pristine `main` on this machine (darwin-only test; it self-skips in ubuntu CI) — pre-existing, unrelated.
- [x] `packages/browseros`: `uv run python -m unittest discover -s bos_build -t . -p "*_test.py"` — 604 tests green, including new real-config assertions for the download op (all platforms/arches/products) and the copy op (both products), and `RepoTruthTest` for features.yaml ↔ patch-stack sync. `uv run ruff check bos_build` clean; CLI smoke (`browseros build --list`, `--preset release --show-plan`, `product doctor`) all pass.
- [x] Hand-edited `onboarding/BUILD.gn` patch byte-verified: hunk count recomputed (56→61), blob hash regenerated via `git hash-object`, `git apply --check` passes in a scratch tree, applied content byte-matches the post-image.
- [ ] The full chromium build cannot run locally — grit/pak integration is verified via the config/patch inspection and unit tests above; the first `nightly-macos-build` run after merge exercises it for real.


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1564, opened 2026-07-04
fix: agent/server instability — OTA gate, payload caps, UI recovery

shadowfax92:
## Summary

Root-causes and fixes the cluster of agent-instability reports (Discord: chat self-close w/ lost history; browseros_server restarting every ~14-15 min with port flips 9200↔9201; invalid auto-written claude-code MCP config; context overload on large-page extractions).

- **Server (OTA-shippable, fastest user reach):** `GET /status` now implements the `can_update` readiness contract the shipped C++ OTA updater already polls — updates defer while a chat stream, harness turn, or MCP tool call is in flight, instead of killing active work every 15-min check. `POST /shutdown` is restored (removed in aacdd4ad4 as "unused", but every shipped 0.46.x browser calls it on managed restarts; without it they fall back to SIGKILL). The claude-code `~/.claude.json` entry now gets `"type": "http"` (agent-mcp-manager@0.0.3 omits it at system scope → Claude Code parsed the entry as stdio and failed with "command: expected string").
- **browser-mcp:** `snapshot`/`diff`/`grep` stop duplicating full payloads into `structuredContent` (the large-snapshot guard was defeated by its own structured field carrying the entire snapshot); `grep` gains limit/line/total caps with full-fidelity file spill; `evaluate` gains the `read`-style 5k inline cap + spill (e.g. `return document.documentElement.outerHTML` no longer explodes context twice). No in-repo consumers of the removed fields; none of these tools declare output schemas.
- **Chromium patches (next browser release):** managed restarts keep the current server port (scan from current w/ SO_REUSEADDR so our own TIME_WAIT socket doesn't force the 9200↔9201 flip; a code-2 port-conflict exit advances instead — matches the documented EXIT_CODES contract) and health-check recycles need 2 consecutive failures instead of 1.
- **Side panel (apps/app):** agent-server URL resolved fresh per send (was resolved once and pinned in a ref — a port flip bricked the open panel until remount); ChatError's retry button actually wired (suppressed for non-retryable credit/rate-limit classes); conversations persist when a turn ends in `error` too, so interrupted turns survive a panel reopen.

## Design

Four independent lanes matched to where each root cause lives and how fast each surface ships (server = OTA within hours via the 15-min appcast; UI ships with server resources; C++ = next browser release). No architectural rework — every fix rides an existing mechanism: the shipped C++ already polls `/status.can_update` and POSTs `/shutdown`; the payload caps copy `read.ts`'s proven spill pattern; the UI reuses the harness screen's fresh-URL-per-call pattern. Full root-cause analysis with file:line evidence, requirements, and plan live in the branch workspace (`.llm/260704-0934_agent_server_instability/`, gitignored) — summary: the ~15-min cadence is `kUpdateCheckInterval` (OTA updater), the port flip is a no-SO_REUSEADDR probe hitting the killed server's TIME_WAIT socket, the "shutdown-endpoint" log reason is the C++ RequestShutdown path, and the missing readiness gate meant hot-swaps landed mid-turn.

## Test plan

- [x] `bun run test:main`, `test:api` (150), `test:lib` (264), browser-mcp suite (24), app unit tests (33+11) — all green with verified exit codes
- [x] `bun run typecheck` + fallow green; Biome clean on all 48 changed files (root lint has pre-existing failures on main in `apps/claw-app`/`codegen.ts`, untouched here)
- [x] Chromium patch files pass `git apply --check`; hunk counts validated; unittests updated (keep-port, exit-2 advance + flag reset, 2-strike health)
- [x] Adversarial context-free review ran: 0 critical, 1 important + 6 minor — all 7 fixed and re-verified (grep spill fallback, evaluate spill extension/length, dead reconcile calls, retry suppression, dep order, ref-write timing, C++ test gap)
- [ ] Post-OTA field check: with a pending server update and an active chat, `/status` returns `can_update:false` and the update applies at the next idle tick; claude-code entry in `~/.claude.json` carries `type: http` after an Integrations install

Note for ops (not in this PR): the stable appcast `appcast-server.xml` is currently serving the **alpha** channel feed (title "BrowserOS Server (Alpha)", pushed 0.0.124 on Jul 1 — the exact day of two reports); worth correcting on the CDN side.


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1567, opened 2026-07-04
feat(claw-server): session-scoped MCP identity + elicitation tab-group naming

shadowfax92:
## Summary
- Session-scope the v2 MCP agent identity so two sessions from the same client (e.g. two Claude Code windows) no longer share a tab group / page-ownership ledger. `agentId` is now `<client-slug>-<6hex FNV-1a(sessionId)>`; `slug` stays the plain client handle.
- Ask the user to name each session via MCP **elicitation** (capability-gated, fire-and-forget after `initialize`), and title the session's tab group `<client>/<small-name>` — e.g. `claude/invoice-processing`, `cursor/linkedin-jobs`.
- All failure modes (no capability, decline, cancel, transport error, user-ignore timeout, unusable input) fall back silently to prior naming behavior; naming never blocks init or tool calls.

## Design
Isolation happens in one place: `agentIdentityFromClient` in `lib/mcp-session/identity.ts` is the single bridge every consumer (ownership ledger, tab-group tracker, cross-agent page guard, tabs routes, dispatch-cancellation, replay) keys off, so session-suffixing the `agentId` there scopes everything with no consumer-signature changes. Naming is a pure module (`lib/mcp-session/naming.ts`: normalize small-name, derive client prefix, compose title) plus a thin glue (`mcp/session-naming.ts`) fired non-blocking from `oninitialized`: capability check → `elicitInput` (120s timeout, one ~2s retry only for the fast SSE-stream race, never on a `RequestTimeout`) → on accept, normalize, store `sessionLabel`, apply the group title. Titles set before the first `tabs new` survive the tracker's pre-seed refresh via a `titleExplicit` flag and are picked up at group creation; titles accepted after the group exists rename it via `tab_groups update` (awaiting any in-flight create).

## Test plan
- [x] `cd packages/browseros-agent/apps/claw-server && bun test` → 416 pass / 0 fail
- [x] `bun run typecheck` → clean
- [x] `bunx biome check` on changed files → clean
- [x] Wire-level test: two SDK clients both named `claude-code` get distinct agentIds; session B's page guard rejects session A's page id and B's `tabs list` hides it
- [x] Elicitation flow unit tests: no-capability / accept / empty-normalize / decline / cancel / double transport-failure / user-ignore-timeout-no-retry


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1574, opened 2026-07-04
fix(ci): point internal-docs sync at main, dev branch is gone

shadowfax92:
Every scheduled run of `sync-internal-docs.yml` has been failing at the checkout step with **"A branch or tag with the name 'dev' could not be found"** — the workflow still targets the removed `dev` branch (checkout ref, `gh pr list --base`, `gh pr create --base`).

This repoints all of them at `main`, the current default branch. No logic changes.

Verified: last 5 scheduled runs all failed at the same step (run 28718451951 and earlier); `git ls-remote` confirms no `dev` head exists.

https://claude.ai/code/session_01HeZKoAHyhGFYi47Wq4T59w

### DIFF
--- a/.github/workflows/bos-build-tests.yml
+++ b/.github/workflows/bos-build-tests.yml
@@ -4,7 +4,6 @@ on:
   pull_request:
     branches:
       - main
-      - dev
     paths:
       - "packages/browseros/bos_build/**"
       - "packages/browseros/tools/vendor_uploads/**"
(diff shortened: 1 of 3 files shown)


shadowfax92 (commented):
Reviewed the workflow branch-targeting fix. Verified sync-internal-docs now checks out main and creates/lists PRs against main, exact dev checkout/base scan is clean, and stale dev PR filters in sibling workflows were removed.

PR #1575, opened 2026-07-04
feat: auto-annotate bpatch refresh

shadowfax92:
## Summary
- Make `browseros-patch refresh` preserve tracked local checkout changes through the rebuild and auto-annotate them into feature commits by default.
- Add `refresh --no-annotate` to opt out, while keeping `--force` as tracked-state abandonment.
- Harden refresh safety around pending resolution, untracked files, stash replay conflicts, renames, and mode-only changes.

## Design
Refresh now stashes only tracked local changes after preflight checks, rebuilds the canonical `browseros` branch as before, restores the stash, and runs the shared annotation machinery scoped to the tracked paths it preserved. Stash replay conflicts are persisted in workspace state so later mutating commands refuse until the recorded stash is dropped. The CLI shares apply's annotate output/failure handling and docs cover the new default and opt-out.

## Test plan
- [x] `make test` from `packages/browseros/tools/patch`
- [x] Local adversarial review loop: 3 rounds, final clean

### DIFF
--- a/packages/browseros/tools/patch/cmd/apply.go
+++ b/packages/browseros/tools/patch/cmd/apply.go
@@ -64,11 +64,11 @@ func init() {
 					}
 					fmt.Println(ui.Hint(`Run "browseros-patch continue" after fixing the current conflict.`))
 				}
-				printAnnotateOutcome(ws, result)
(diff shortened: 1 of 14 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1577, opened 2026-07-04
refactor(patches): server proxy → pure TCP byte forwarder with HTTPS

shadowfax92:
## Summary
- Rewrites `BrowserOSServerProxy` from a request-re-issuing HTTP server (`net::HttpServer` + `SimpleURLLoader`: 5MB body cap, 300s timeout, header whitelist, WebSockets closed) into a transparent L4 byte forwarder between the stable proxy port and the sidecar backend — SSE/streaming, chunked, WebSocket upgrades, and unbounded bodies now traverse byte-intact with zero request/response mutation.
- Adds an HTTPS listener that terminates TLS with a per-start in-memory self-signed cert (EC P-256 + `net::x509_util::CreateSelfSignedCert` + `net::CreateSSLServerContext`), pref `browseros.server.proxy_https_port` (default 9001). Any TLS setup failure logs and degrades to HTTP-only.
- Manager simplification: the UI-thread `SharedURLLoaderFactory` clone dance is gone; proxy needs no network service at all.

## Design
Raw `net/` sockets on the IO thread, pump modeled on the in-tree devtools `SocketTunnel` (`port_forwarding_controller.cc`) but with proxy-owned connection lifetime so `Stop()` tears down deterministically. Both listeners share one protocol-agnostic `Connection` pump over `net::StreamSocket*`; the TLS path is just a handshake wrapper before the same pump. Preserved contracts: stable port, canned 503 during the no-backend startup window (now with lingering close so unread request bytes can't RST-clobber the response), loopback-only gate (moved to accept time — remote peers get connection-close instead of HTTP 403, deliberate for a forwarder that makes no protocol assumptions). Accept errors re-arm via delayed task instead of recursing.

## Test plan
- [x] 15 proxy unit tests incl. transparency proofs: streamed bytes observable while backend connection still open; 6MB+ bodies pattern-verified both directions; 101-upgrade then raw binary both ways; teardown drains connection map; TLS handshake+forward via `SSLClientContext`+`MockCertVerifier`; degrade-to-HTTP-only on occupied HTTPS port; construction-on-UI/start-on-IO thread affinity
- [x] `autoninja -C out/Default_arm64 chrome unit_tests` green; `unit_tests --gtest_filter='BrowserOSServerProxy*:BrowserOSServerManager*'` green (proxy filter ×5 repeats, no flakes)
- [x] Patch verification: 9/9 byte-match vs `git diff BASE..tip` and 9/9 apply-check onto bare BASE tree (`verify-patches: OK`)

## Notes
- Also repairs 3 `BrowserOSServerManager` tests that were broken at the current patch-store tip independent of this change (unit tests calling `Start()` tried to launch a real CDP server; restart test relied on running state the mock never set).
- Adversarial review ran locally pre-PR: 9 findings → 6 fixed (incl. a Critical thread-checker-binds-on-wrong-thread that crashed all `is_debug` builds), 3 dropped with rationale (buffer churn / sync recursion match SocketTunnel precedent; no-SAN cert is a documented non-goal).

### DIFF
--- a/packages/browseros/chromium_patches/chrome/browser/browseros/server/BUILD.gn
+++ b/packages/browseros/chromium_patches/chrome/browser/browseros/server/BUILD.gn
@@ -1,9 +1,9 @@
 diff --git a/chrome/browser/browseros/server/BUILD.gn b/chrome/browser/browseros/server/BUILD.gn
 new file mode 100644
-index 0000000000000..8f1f42fae7854
+index 0000000000000..aed05041c83b4
 --- /dev/null
 +++ b/chrome/browser/browseros/server/BUILD.gn
-@@ -0,0 +1,173 @@
+@@ -0,0 +1,175 @@
 +# Copyright 2024 The Chromium Authors
 +# Use of this source code is governed by a BSD-style license that can be
 +# found in the LICENSE file.
@@ -110,6 +110,7 @@ index 0000000000000..8f1f42fae7854
 +    "browseros_appcast_parser_unittest.cc",
 +    "browseros_server_config_unittest.cc",
 +    "browseros_server_manager_unittest.cc",
++    "browseros_server_proxy_unittest.cc",
 +    "browseros_server_utils_unittest.cc",
 +  ]
 +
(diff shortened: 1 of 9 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1598, opened 2026-07-05
fix(build): lock chromium checkout during release builds

shadowfax92:
## Summary
- Add an exclusive checkout lock keyed to the resolved Chromium src path, with holder metadata for pid/product/start time/host.
- Wrap `browseros build` before preflight and step execution so clean, git setup, patching, compile, sign, and package cannot run concurrently on the same checkout.
- Keep default behavior fail-fast with a clear contention error; add `--lock-wait` for deliberate serialization.

## Design
The lock file is stored beside the gclient checkout root under `.browseros-build-locks`, outside `src/`, so the existing `git clean -fdx` cannot delete it. A contending process reads the holder metadata and exits before touching the checkout unless `--lock-wait` is supplied.

## Test plan
- [x] `uv run python -m unittest bos_build.cli.build_test bos_build.core.checkout_lock_test`
- [x] `uv run --group dev ruff check bos_build/core/checkout_lock.py bos_build/core/checkout_lock_test.py bos_build/cli/build.py bos_build/cli/build_test.py`
- [x] `uv run --group dev ruff format --check bos_build/core/checkout_lock.py bos_build/core/checkout_lock_test.py bos_build/cli/build.py bos_build/cli/build_test.py`
- [x] `uv run python -m unittest discover -s bos_build -t . -p '*_test.py'`

### DIFF
--- a/packages/browseros/bos_build/cli/build.py
+++ b/packages/browseros/bos_build/cli/build.py
@@ -11,6 +11,7 @@ from typing import Callable, List, Optional, Tuple
 import typer
 
 from ..core.context import Context
+from ..core.checkout_lock import CheckoutLockError, ChromiumCheckoutLock
 from ..lib.paths import get_package_root
 from ..core.pipeline import validate_pipeline, show_available_modules
 from ..core.planner import (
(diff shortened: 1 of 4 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1601, opened 2026-07-05
fix(bos_build): two-pass windows cache extract for symlink entries

shadowfax92:
## Summary
- Add a Windows-only cache restore path that decompresses the downloaded `.tar.zst` to a rewindable `.tar` before extraction.
- Run Git tar with `MSYS=winsymlinks:nativestrict` and retry one full extraction pass after a first-pass failure.
- Keep POSIX restore and cache save on the existing streaming pipeline and preserve the cache object format.

## Design
Windows restore now uses a restore-only command runner for `zstd -d -f -o <tar> <tar.zst>` and `tar -xf <tar>` from the resolved chromium root. The compressed file is removed after successful decompression, the temporary tar is removed in cleanup, and tar extraction remains fatal if the retry also fails.

## Test plan
- [x] `uv run python -m unittest discover -s bos_build -t . -p "*_test.py"`

### DIFF
--- a/packages/browseros/bos_build/steps/source/cache.py
+++ b/packages/browseros/bos_build/steps/source/cache.py
@@ -115,6 +115,56 @@ def _run_pipeline(
         )
 
 
+def _run_command(
+    cmd,
+    *,
+    cwd: Path | None = None,
+    env: dict[str, str] | None = None,
+) -> int:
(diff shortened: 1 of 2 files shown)


shadowfax92:
@claude review

shadowfax92:
@greptileai review

PR #1641, opened 2026-07-06
fix: normalize rust mcp schemas for Claude Code

shadowfax92:
## Summary
- Normalize generated Rust MCP schemas at the `browseros-mcp` boundary so boolean JSON Schema forms never reach `tools/list`.
- Pin schema generation to Draft 2020-12 with inline subschemas, preserving MCP-aligned schema generation while avoiding `$ref`/`$defs` output for strict clients.
- Add catalog-wide regressions that reject boolean schema values/references and assert the `run` output schema keeps object-form `ok`, `logs`, `error`, and `value` properties.

## Design
The Rust server was exposing `run.outputSchema.properties.value` as the legal JSON Schema boolean form `true` because `serde_json::Value` derives to an any-schema under schemars. Claude Code's MCP tools-list validation rejects non-object schema nodes, so one boolean subschema caused the entire tools list to fail. This fix normalizes schema-position boolean forms at the single framework boundary (`true` to `{}`, `false` to `{ "not": {} }`), removes boolean `default` annotations from the emitted compatibility schema, and fails loudly if any unsupported boolean value remains.

## Test plan
- [x] `cargo test -p browseros-mcp schema -- --nocapture` failed before the fix on boolean schema output and boolean input defaults.
- [x] `cargo test -p browseros-mcp --lib -- --nocapture`
- [x] `bun run fmt:rust`
- [x] `bun run lint:rust`
- [x] `bun run test:rust`
- [x] `bun run check` passes with existing warnings in unrelated app/fallow output.
- [x] `bun run test`
- [x] Live curl initialize/tools-list probe against `target/debug/browseros-claw-server-rs`: `tool_count=16`, `boolean_schema_values=0`, `schema_refs=0`, `schema_drafts=https://json-schema.org/draft/2020-12/schema`, `run_value_schema={}`.

### DIFF
--- a/packages/browseros-agent/crates/browseros-mcp/src/framework.rs
+++ b/packages/browseros-agent/crates/browseros-mcp/src/framework.rs
@@ -230,18 +230,52 @@ pub fn input_schema<T>() -> Arc<JsonObject>
 where
     T: JsonSchema + std::any::Any,
 {
-    rmcp::handler::server::tool::schema_for_input::<T>().unwrap_or_else(|err| {
+    let schema = rmcp::handler::server::tool::schema_for_input::<T>().unwrap_or_else(|err| {
         panic!("invalid BrowserOS MCP input schema: {err}");
-    })
+    });
+    normalize_schema_object(schema)
 }
 
 pub fn output_schema<T>() -> Arc<JsonObject>
 where
     T: JsonSchema + std::any::Any,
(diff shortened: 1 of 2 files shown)


shadowfax92:
@claude review

shadowfax92:
@greptileai review

PR #1663, opened 2026-07-07
feat: make core ref input occlusion-aware

shadowfax92:
## Summary
- Add a `CoreError::ElementCovered` path that reports ref/role/name context and a compact blocker descriptor.
- Hit-test ref/node click points in the target frame before eligible click, hover, check/uncheck, and select interactions.
- Preserve existing behavior when hit-testing fails, and leave coordinate input and drag paths unchanged.

## Design
The pre-check resolves the target backend node to a runtime object and calls a small in-page `elementFromPoint` helper in that object context. The helper treats target ancestors, descendants, labels, and shadow hosts as safe, returning a blocker descriptor only when an unrelated element would receive the input.

## Test plan
- [x] `cargo build --workspace`
- [x] `cargo test --workspace`
- [x] `cargo clippy --workspace --all-targets -- -D warnings`

### DIFF
--- a/packages/browseros-agent/crates/browseros-core/src/error.rs
+++ b/packages/browseros-agent/crates/browseros-core/src/error.rs
@@ -1,5 +1,34 @@
 use crate::{PageId, Ref};
 use browseros_cdp::CdpError;
+use std::fmt;
+
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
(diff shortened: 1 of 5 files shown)


shadowfax92:
@claude review

shadowfax92:
@greptileai review

PR #1689, opened 2026-07-08
fix: correct onboarding patch hunk counts

shadowfax92:
## Summary\n- correct BrowserOS onboarding patch hunk headers whose counts were stale after the completion-crash fix\n- prevents bpatch refresh from truncating browseros_onboarding.cc and browseros_onboarding_prefs.cc\n\n## Verification\n- uv run python -m bos_build.cli.dev doctor\n- verified new-file hunk counts match added-line counts\n- verified patch object hashes match index headers\n- Chromium pool build in ch-3 is running with the local equivalent source fixes


shadowfax92:
@claude review

shadowfax92:
@greptileai review

PR #1747, opened 2026-07-09
fix: launch Windows server without a console window

shadowfax92:
## Summary
- route managed server startup through a BrowserOS-owned platform launcher instead of extending Chromium base APIs
- launch the Windows console-subsystem server with `CREATE_NO_WINDOW` while preserving the managed process handle and existing non-Windows behavior
- add multiprocess regression coverage that distinguishes no console from a merely hidden console

## Design
`start_hidden` only supplies `STARTF_USESHOWWINDOW` and `SW_HIDE`; it controls presentation but does not prevent Windows from allocating a console for a console-subsystem executable. The Windows launcher uses `CreateProcessW` with the same command line, inherited environment/current directory, non-inherited handles, and owned process handle as the prior default launch, adding only `CREATE_NO_WINDOW`. `SW_HIDE` remains as compatibility behavior if the packaged executable changes subsystem.

## Test plan
- [x] `autoninja -C out/Default_browseros_arm64 chrome/browser/browseros/server:server chrome/browser/browseros/server:unit_tests`
- [x] `out/Default_browseros_arm64/unit_tests --gtest_filter='ServerProcessLauncherTest.*:BrowserOSServerManagerTest.*'` (20 passed)
- [x] `autoninja -C out/Default_browseros_arm64 chrome`
- [x] extracted patch verification: 6/6 byte-match and 6/6 apply-check

## References
- [Process creation flags](https://learn.microsoft.com/en-us/windows/win32/procthread/process-creation-flags)
- [STARTUPINFO](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/ns-processthreadsapi-startupinfow)

### DIFF
--- a/packages/browseros/chromium_patches/chrome/browser/browseros/server/BUILD.gn
+++ b/packages/browseros/chromium_patches/chrome/browser/browseros/server/BUILD.gn
@@ -1,9 +1,9 @@
 diff --git a/chrome/browser/browseros/server/BUILD.gn b/chrome/browser/browseros/server/BUILD.gn
 new file mode 100644
-index 0000000000000..aed05041c83b4
+index 0000000000000000000000000000000000000000..5fadaa41de9c2259e4d0cd35a08b758f694ba703
 --- /dev/null
 +++ b/chrome/browser/browseros/server/BUILD.gn
-@@ -0,0 +1,175 @@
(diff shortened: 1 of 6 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1748, opened 2026-07-09
fix(bpatch): stop treating annotate commits as drift

shadowfax92:
## Summary
- anchor committed drift at the newest bpatch-authored apply or annotate commit while keeping applied-store bookkeeping apply-only
- make `bpatch diff` share the exact HEAD-overlay target used by apply, including `store: false` filtering
- preserve legacy `Bpatch-Tree` semantics and harden zero-file, metadata-only, and real-drift convergence paths

## Design
State resolution tracks separate apply and drift anchors. Apply reconstructs the prior raw store target from the recorded store revision, overlays only the real store delta onto HEAD, and records explicit state-only transitions when only store metadata advances. Exact-target bookkeeping is restricted to apply-authored anchors, so hand commits after annotate remain drift.

## Test plan
- [x] `cargo clippy --all-targets -- -D warnings`
- [x] `cargo test --locked` (81 tests)
- [x] `make install`
- [x] ch4 smoke: annotate-authored history is clean; the original six-file store delta applied; immediate repeat converged
- [x] ch4 final read-only check: clean with no drift; newer store revision left unapplied

### DIFF
--- a/packages/browseros/tools/bpatch/src/cli/apply.rs
+++ b/packages/browseros/tools/bpatch/src/cli/apply.rs
@@ -147,7 +147,7 @@ pub struct ApplyConflictFile {
 #[derive(Clone, Debug, Eq, PartialEq, Serialize)]
 #[serde(rename_all = "snake_case")]
 pub enum ApplyDriftSource {
-    /// Drift is committed on top of the applied tree.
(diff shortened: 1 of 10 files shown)


shadowfax92:
@claude review

shadowfax92:
@greptileai review

PR #1750, opened 2026-07-09
fix: package BrowserClaw payloads and executable identity

shadowfax92:
## Summary
- package the BrowserClaw server and shared bundled-extension directory in the static Windows installer manifest while preserving the BrowserOS server block
- select BrowserOS, BrowserClaw, or debug CRX inputs from the existing GN product booleans
- compile POSIX executable constants from the baked product buildflag and add static payload-contract coverage

## Design
The Windows manifest declares both server roots because unmatched globs are intentionally skipped; clean product GN output decides which root ships. Bundled extensions retain the shared `browseros_extensions` path, while GN selects agent plus bug reporter for BrowserOS release, Claw plus bug reporter for BrowserClaw release, and all three when runtime product override is enabled. macOS naming remains driven by BRANDING.

## Merge ordering
This Chromium GN matrix must merge before or at the same time as the parallel Python exact-staging PR. If Python stages only the active two CRXs while the old GN file still lists all three, Ninja fails on the missing inactive CRX input.

## Test plan
- [x] `uv run python -m unittest bos_build.steps.patches.product_payload_manifest_test -v`
- [x] `uv run browseros dev doctor` (379 patches, 34 features)
- [x] extracted patches: 4/4 byte-match and 4/4 apply-check
- [x] shared pool build: `autoninja -C out/Default_browseros_arm64 chrome`
- [x] live Windows Claw server ZIP contains `resources/bin`, `resources/db/migrations`, and `resources/db/migrations/meta`, with no `bin/third_party`

### DIFF
--- a/packages/browseros/bos_build/steps/patches/product_payload_manifest_test.py
+++ b/packages/browseros/bos_build/steps/patches/product_payload_manifest_test.py
@@ -1,0 +1,126 @@
+#!/usr/bin/env python3
+"""Tests for product payload manifest patches."""
+
+import re
+import unittest
+
+from ...lib.paths import get_package_root
+
+
+PATCHES = get_package_root() / "chromium_patches"
+
+
+def _patched_source(relative_path: str) -> str:
+    """Reconstruct the changed source regions from a unified diff."""
+    source_lines: list[str] = []
+    in_hunk = False
+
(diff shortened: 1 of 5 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1751, opened 2026-07-09
fix(linux): package bundled extensions in appimage and deb

shadowfax92:
## Summary
- copy the required browseros_extensions output beside Linux browser executables for both products
- fail before staging when the required extension directory is missing
- preserve optional files, active-product server roots, product naming, and the Chromium resources boundary

## Design
The shared copy_browser_files helper preflights the generated out-dir payload and copies it explicitly, separate from the existing optional directory loop. Both AppImage and deb staging already use this helper, so they inherit the same fatal contract without format-specific changes.

## Test plan
- [x] uv run python -m unittest bos_build.steps.package.linux_test -v
- [x] uv run ruff check bos_build
- [x] uv run python -m unittest discover -s bos_build -t . -p "*_test.py" -v
- [x] release-secrets and vendor-upload unittest suites
- [x] browseros build list/show-plan and product doctor smokes
- [x] context-free diff review: clean

### DIFF
--- a/packages/browseros/bos_build/steps/package/linux.py
+++ b/packages/browseros/bos_build/steps/package/linux.py
@@ -163,8 +163,16 @@ def copy_browser_files(
     Returns:
         True if successful, False otherwise
     """
-    target_dir.mkdir(parents=True, exist_ok=True)
     out_dir = join_paths(ctx.chromium_src, ctx.out_dir)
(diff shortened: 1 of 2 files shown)


shadowfax92:
@claude review

shadowfax92:
@greptileai review

PR #1752, opened 2026-07-09
fix(build): stage exact product extension sets

shadowfax92:
## Summary

- declare exact BrowserOS and BrowserClaw release extension ownership at each product definition
- derive the runtime-override debug union once in `Context` and use it for downloaded and local/hybrid staging
- remove only generated CRXs/manifests and the active out-dir extension tree before staging, preventing sequential build leakage

## Design

Product descriptors own release requirements, while `Context` turns those into the effective build requirements: the active pair for release and the deduplicated registered-product union for debug. Every staging, validation, hybrid, and local-preflight path consumes that one boundary. Cleanup is deliberately narrow so checked-in staging sources and unrelated output trees survive.

## Merge order

This PR must merge **after** the parallel Chromium GN PR from `task/claw-payload-manifests` / remote extraction branch `feat/claw-payload-manifests-patches` (currently commit `c3656e702099ee2c60d3417e086bfce23f093036`), or in the same release train after that commit. This Python change stages only two CRXs for release; the old GN list names all three as sources and would make Ninja fail on the inactive product's missing input.

## Test plan

- [x] `uv run python -m unittest bos_build.steps.extensions.bundled_extensions_test bos_build.products.products_test bos_build.core.context_test -v`
- [x] `uv run browseros product doctor`
- [x] `uv run python -m unittest discover -s bos_build -t . -p '*_test.py' -v` (668 tests)
- [x] Ruff formatting check for all eight touched Python files

### DIFF
--- a/packages/browseros/bos_build/core/context.py
+++ b/packages/browseros/bos_build/core/context.py
@@ -298,6 +298,20 @@ class Context:
         base = self.get_chromium_replace_files_dir()
         return [base / "common", base / "products" / self.product.id]
 
+    @property
+    def required_extension_ids(self) -> tuple[tuple[str, str], ...]:
+        """Return bundled-extension requirements for the current build."""
+        if self.build_type != "debug":
+            return self.product.required_extension_ids
+
+        from ..products import PRODUCTS
+
+        required: Dict[str, str] = {}
+        for product in PRODUCTS.values():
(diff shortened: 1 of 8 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1753, opened 2026-07-09
fix(windows): sign chrome.exe before installer packaging

shadowfax92:
## Summary

- keep Windows `chrome.exe` at Chromium's canonical path through executable signing and mini-installer construction
- fail signing immediately when the primary browser executable is missing
- copy the signed/built executable to `BrowserOS.exe` or `BrowserClaw.exe` only after Windows installer artifacts are created
- make the Windows debug plan build WinSparkle and the unsigned mini-installer before packaging

This contract applies equally to BrowserOS and BrowserClaw releases. Linux and macOS compile-time rename behavior is unchanged.

## Evidence

- [Windows release run 28836949185](https://github.com/browseros-ai/BrowserOS/actions/runs/28836949185) entered `sign_windows` and logged `Binary not found: ...\chrome.exe`, then proceeded with only the product server executable. This confirms the missing-primary-executable signing boundary on a fresh CI output directory.
- That job subsequently failed on signing-provider authentication before mini-installer construction. It therefore does **not** prove that a released installer shipped an unsigned `chrome.exe`; that artifact-level claim remains unverified.
- The source plan orders Windows release builds as `compile` → `sign_windows` or `mini_installer` → `package_windows`. Before this change, compile moved `chrome.exe` to the product name while signing still required `chrome.exe`.

## Design

Windows compile now registers and preserves `chrome.exe`. The signing step treats it as mandatory, signs it before building the mini-installer, and still signs the outer installer afterward. Both signed and unsigned routes converge on `package_windows`, which creates the installer and ZIP first, then uses `shutil.copy2` to overwrite the exact descriptor-derived product executable path without removing `chrome.exe` from Ninja's output graph.

## Test plan

- [x] missing `chrome.exe` is fatal for BrowserOS and BrowserClaw even when their server binaries exist
- [x] Windows compile preserves/registers `chrome.exe`; non-Windows compile keeps the existing rename
- [x] unsigned mini-installer construction leaves product aliasing to packaging
- [x] packaging creates installer outputs before overwriting `BrowserOS.exe` / `BrowserClaw.exe`, while preserving `chrome.exe`
- [x] Windows debug plan includes WinSparkle setup and mini-installer before package
- [x] `cd packages/browseros && uv run python -m unittest discover -s bos_build -t . -p "*_test.py" -v` (667 passed)
- [x] `cd packages/browseros && uv run ruff check bos_build`

### DIFF
--- a/packages/browseros/bos_build/core/planner.py
+++ b/packages/browseros/bos_build/core/planner.py
@@ -228,6 +228,8 @@ def _plan_release(switches: Switches, platform: str) -> List[str]:
 def _plan_debug(switches: Switches, platform: str) -> List[str]:
     steps: List[str] = []
     steps.extend(_provision_steps(switches))
+    if platform == "windows":
+        steps.append("winsparkle_setup")
     if switches.download:
         steps.append("download_resources")
     steps.extend(
@@ -242,6 +244,8 @@ def _plan_debug(switches: Switches, platform: str) -> List[str]:
     )
     if platform == "macos" and switches.sign:
         steps.append("sign_macos")
+    if platform == "windows":
+        steps.append("mini_installer")
     steps.append(f"package_{platform}")
     if switches.upload:
         steps.append("upload")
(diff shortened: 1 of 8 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1777, opened 2026-07-10
fix(claw-server-rust): HTTP parity batch (cancel 404 shape, MCP hygiene, failure logs, env overrides, screenshot fallback flag)

shadowfax92:
## Summary
- **Cancel 404 body shape**: `POST /agents/{agent_id}/cancel` with no active dispatches now returns the TS-shaped 404 body `{"ok":false,"cancelled":0,"reason":"no active dispatches for this agent"}` instead of the generic `{"error":...}` — claw-app parses this body as a `CancelAgentResult` without throwing, so `result.ok`/`result.cancelled` were coming back `undefined` against the Rust server.
- **MCP request hygiene middleware** (parity with `routes/mcp/mcp-request-hygiene.ts`), layered on the `/mcp` nest only: any request carrying `origin` or `sec-fetch-site` → 403 `{"error":"unsupported request"}`; POST/PUT/PATCH/DELETE without an `application/json` content-type → 415 `{"error":"unsupported content type"}`. Carve-out: DELETE with **no** content-type header passes, because rmcp's own session-teardown DELETE sends no body/content-type (proven by the existing integration test); native MCP clients always send `application/json` on writes with bodies. OPTIONS answers 204 preflight (the nested service shadows the router's catch-all preflight route; on TS, hono's cors layer answers OPTIONS before hygiene runs).
- **Request failure logging** (parity with `server.ts` `requestFailureLog`): one structured line per response with status ≥ 400 (`warn` 4xx / `error` 5xx) with method, path, status, duration_ms. Sub-400 stays unlogged on purpose — claw-app polls `/tabs/activity` at 1.5s and `/audit/*` at 3s.
- **Env override parity** (vs `src/env.ts`): `CLAW_SESSION_IDLE_MS`, `CLAW_SESSION_SWEEP_INTERVAL_MS`, `CLAW_SCREENCAST_SCREENSHOT_FALLBACK`, and `BROWSERCLAW_DIR` were already honored with identical names/defaults; the ms readers now also mirror `Number.parseInt(raw, 10)` prefix semantics (`"500ms"` → 500, whitespace tolerated, non-positive/digit-less → default). Parsing is unit-tested for all four vars.
- **Screenshot fallback flag honored**: `ScreenshotPersister` (mcp/hooks.rs) now skips the fallback capture when `screencast_screenshot_fallback` is false, while images explicitly returned by a tool are still persisted — matching `services/screenshots.ts`. A tool-carried image also marks the page's first capture (TS branch-1 parity), so later read-only dispatches no longer trigger an extra CDP capture. `services/screencast.rs` untouched.

## VERIFY items (as requested)
- **(a) Unknown `mcp-session-id` on /mcp**: rmcp returns 404 with a **plain-text** body `"Not Found: Session not found"`, while TS returns 404 JSON `{"error":"unknown mcp-session-id","hint":...}`. Not fixed here: rewriting rmcp's response bodies would require buffering middleware over the streaming service — not trivial, and no consumer parses this body (claw-app never calls /mcp; MCP clients only react to the 404 status by re-initializing).
- **(b) GET /system/version**: Rust reports CARGO_PKG values (`claw-server-rust` / `0.1.0`) vs TS package name/version. Confirmed no consumer parses these: claw-app never calls `/system/*`, and tools/dev only polls `/system/health` (grep over both). Left as-is (rust-identified) per the decision.

## Notes / deliberate deltas
- `NODE_ENV` is deliberately not read by the Rust config (pinned by the existing `ignores_node_env_when_sidecar_flag_is_missing` test): the dev-vs-prod state-dir switch (`~/.browserclaw-dev` vs `~/.browserclaw`) comes from the sidecar `flags.devMode` or the build profile. The dev stack runs `cargo run` (debug ⇒ dev dir) with `NODE_ENV=development`, and release binaries run without `NODE_ENV` ⇒ prod dir — behavior matches the TS server in both real flows.
- **Pre-existing race surfaced while stabilizing tests** (out of scope for this PR, worth a follow-up in `crates/browseros-mcp`): `ensure_session_started` (service.rs) sets `lifecycle.started = true` and *then* awaits the `session_started` hook, so a tools/call racing the `notifications/initialized` processing can hit `"mcp session ... is not registered"`. Reproduced on `main` (3/10 runs under test concurrency). Tests now wait for registration via a shared `wait_for_session_registration` helper.
- Bodyless-DELETE carve-out means a content-type-less `DELETE /mcp` gets 202 here where TS would 415 — required to keep rmcp's real teardown flow working; decided in the task.

## Test plan
- [x] `cargo test -p claw-server-rust` — 38 unit + 12 integration tests green (10 consecutive full-suite runs to rule out the flake)
- [x] `cargo clippy -p claw-server-rust --all-targets` — clean under workspace lints (deny unwrap/expect/unsafe)
- [x] New tests: cancel-404 body; hygiene 403 (origin, sec-fetch-site) / 415 (text/plain) / 204 (OPTIONS) / scope (non-/mcp routes unaffected); env ms + bool + dir parsing matrix; explicit-image persist with fallback disabled (unit) + fallback-off ⇒ no audit screenshot, fallback-on ⇒ audit screenshot served (integration, mock CDP)

### DIFF
--- a/packages/browseros-agent/apps/claw-server-rust/src/config.rs
+++ b/packages/browseros-agent/apps/claw-server-rust/src/config.rs
@@ -216,7 +216,8 @@ fn read_positive_ms(env: &ConfigEnv, key: &str, fallback: u64) -> u64 {
     let Some(raw) = env.get(key) else {
         return fallback;
     };
-    raw.parse::<u64>()
+    raw.trim()
+        .parse::<u64>()
         .ok()
         .filter(|value| *value > 0)
         .unwrap_or(fallback)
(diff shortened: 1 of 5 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1778, opened 2026-07-10
feat(claw-server-rust): session naming via MCP elicitation

shadowfax92:
## Summary
- Ports `mcp/session-naming.ts` to the rust claw-server: when an MCP client advertises the `elicitation` capability, a spawned task (never blocking initialize) asks it to name the browser session, normalizes the answer to a 2-3-word label (`normalizeSmallName` semantics, TS test vectors ported), and titles the agent's tab group `<prefix>/<name>` (e.g. `claude/flight-search`).
- Extends the typed `McpHooks` boundary minimally: `McpSessionStarted` now carries the session's `Peer<RoleServer>` (rmcp's native elicitation API — no raw JSON-RPC); enables the `elicitation` cargo feature on the workspace rmcp dep.
- TS retry table mirrored exactly: 120s timeout per attempt; timeout (user ignored) → no retry; transport failure → one retry after 2s; decline/cancel/empty → silent. Label is in-memory only (the TS `sessionLabel` never reaches audit rows — mirrored).

## Design
`ensure_session_started` (crates/browseros-mcp) hands the live peer to the hook; `ClawMcpHooks::session_started` spawns a `SessionNamer` task gated on form-elicitation support. The task's elicit future is raced against the session's cancellation token, so teardown (idle sweep, disconnect, shutdown) frees it immediately; a post-elicit registry lookup handles the swept-session case silently. Late-name arrival retitles the existing group via a `tab_groups` update; a name that lands while the first group create is in flight is applied by a dedicated post-create title update (independent of the color lock, so a color failure can't drop the name); a name known before the first `tabs new` flows into the create's `title` argument. Elicitation prompt + `requestedSchema` are byte-identical to the TS wire format (manual `ElicitationSchema` instead of schemars derive).

Review notes (workflow code-review, high effort — 8 verified findings, 5 fixed):
- Skipped: cross-session title clobber for same-slug concurrent sessions — a consequence of the rust server's *intentional* slug-pooled tab groups (`agent_ref.rs` ownership comment); TS avoids it only because it never pools. Last-namer-wins within the shared pool.
- Skipped: unserialized parallel `tab_groups create` race — pre-existing in `TabGroupOrchestrator`, untouched surface.
- Skipped: rmcp serializes an extra `"mode":"form"` field in elicit params — verified accepted by the real `@modelcontextprotocol/sdk` client (see manual check); newer MCP revisions include `mode`.

## Test plan
- [x] `cargo test -p claw-server-rust -p browseros-mcp` — green (normalizer TS vectors, retry table incl. timeout-no-retry, schema JSON parity with the TS `requestedSchema`, label slot, elicitation-capability initialize integration test)
- [x] `cargo clippy -p browseros-mcp -p claw-server-rust --all-targets` — clean
- [x] Manual check (headless equivalent of "connect Claude Code"): standalone server + a client built on the same `@modelcontextprotocol/sdk` Claude Code embeds, advertising `{elicitation:{}}` → server sent the naming prompt with the exact TS message/schema → client answered "Flight Search" → server log `mcp session named … title=claude/flight-search`. Browser-side retitle path exercises the same `tab_groups` machinery covered by the mock-CDP tests; visible-in-browser confirmation to be done on the devrust watch stack post-merge.

Note: `tests/routes.rs` flaked ~50% on the base tree before #1777's `wait_for_session_registration` workaround (root cause documented there); with main merged in, 3 consecutive full runs green.

### DIFF
--- a/packages/browseros-agent/Cargo.lock
+++ b/packages/browseros-agent/Cargo.lock
@@ -1497,6 +1497,7 @@ dependencies = [
  "tokio-util",
  "tower-service",
  "tracing",
+ "url",
  "uuid",
 ]
 
--- a/packages/browseros-agent/Cargo.toml
+++ b/packages/browseros-agent/Cargo.toml
@@ -18,7 +18,7 @@ tokio-tungstenite = { version = "0.29", features = ["rustls-tls-webpki-roots"] }
 futures-util = "0.3"
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
-rmcp = { version = "2.1", features = ["server", "transport-io", "transport-streamable-http-server"] }
+rmcp = { version = "2.1", features = ["server", "transport-io", "transport-streamable-http-server", "elicitation"] }
 schemars = "1.0"
 axum = "0.8"
 http = "1"
--- a/packages/browseros-agent/apps/claw-server-rust/src/domain/session.rs
+++ b/packages/browseros-agent/apps/claw-server-rust/src/domain/session.rs
@@ -97,6 +97,9 @@ pub struct Session {
     cancel: CancellationToken,
     replay_handle: Mutex<Option<String>>,
     last_activity: Mutex<Instant>,
+    /// User-facing session name gathered via MCP elicitation. In-memory only,
+    /// mirroring the TS identity map's sessionLabel (never persisted to audit).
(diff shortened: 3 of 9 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1779, opened 2026-07-10
feat(claw-server-rust): push-based CDP screencast for live tab thumbnails

shadowfax92:
## Summary
- Replaces the 1.5s `Page.captureScreenshot` polling internals of `ScreencastService` with a push pipeline: each BrowserOS window's active agent tab gets a `Page.startScreencast` (jpeg / q60 / 1280x800 / everyNthFrame 1); frames land in the same LRU cache (cap 50) and `GET /tabs/activity` stays byte-identical (`screencast: {jpegBase64, capturedAt} | null`), so claw-app needs zero changes.
- Frames are routed by the envelope target session id (string) and acked immediately with the integer cookie (`Page.screencastFrameAck`) — lifting the TS server's single-active-screencast constraint; concurrent screencasts across windows are safe (proven in the mock-CDP e2e).
- The captureScreenshot poll stays as fallback for background agent tabs (CDP never composites them; `clip: None` preserved — the BrowserOS fork resizes tabs otherwise), with the existing 3-failure→5s backoff. `Page.bringToFront` is never called (it would steal cockpit focus).
- New: 15s idle governor (no `/tabs/activity` reads → stop casts + skip polls; resumes within one 1s supervisor tick) and a 5s frameless-restart that revives casts silently killed by CDP reconnects.

## Design
CDP verbs are code-generated from the curated protocol JSON (`page::StartScreencastParams`, `ScreencastFrameAckParams`, `ScreencastFrameEvent`, `ScreencastFrameMetadata`), with thin helpers in a new `browseros_core::screencast` module. In the service, a 1s supervisor tick snapshots agent pages (tab_activity), maps them to windows (`pages.list()`), resolves each relevant window's active tab via root `Browser.getActiveTab{windowId}`, and feeds a pure `plan_tick` function (unit-tested decision matrix) whose plan starts/stops casts through the existing attach path. A single pump task consumes frames and acks them; casts register before `startScreencast` so the first frame is always routable.

## Deviations from the dispatch design (as requested, noted here)
- **Frame delivery**: the pump consumes a new `CdpClient::events_targeted(method)` per-method mpsc channel instead of the shared broadcast stream. The 4096-slot broadcast ring retains each event until overwritten — sustained casting would pin hundreds of MB of frame payloads. The targeted channel frees frames on consumption, and CDP ack gating bounds depth at ~1 frame in flight per cast. `RecvError::Lagged` handling is moot on this path (mpsc, not broadcast).
- **Active-tab resolution**: root `Browser.getActiveTab{windowId}` via `BrowserSession::cdp` + `pages.get_session` (attach) only when the active tab is agent-driven — `get_active_session_for_window` would attach (Page/Runtime.enable) to the user's own tabs.
- **Idle-governor timestamp** lives in `ScreencastService` (`note_read()` called by the handler) rather than a new `AppState` field — same behavior, no state-shape change.
- **Unit tests** exercise a pure `plan_tick` decision function (start/stop/poll matrix, idle, frameless-restart) instead of tokio test-util time mocking — deterministic, no sleeps.
- Screencast eligibility follows the design letter: the window's active agent page is cast regardless of tab_activity status; status `"active"` gates only the poll fallback. Poll fallback keeps its existing quality 50; screencast uses the specced quality 60.
- Ack sends are spawned (review finding): a wedged ack cannot stall routing for other casts; CDP's ack gating preserves per-session ordering.

## Review
Adversarial subagent review, 2 rounds: round 1 → 1 Important (broadcast ring retention) + 3 Minor, all fixed in 8b01f5b65; 3 findings dropped with recorded reasons (2 resolved by the origin/main #1777 merge, 1 observation). Round 2 → clean.

## Test plan
- [x] `cargo test -p claw-server-rust -p browseros-cdp -p browseros-core` — green (59 core + 50 claw unit + 13 integration; includes 12 new plan_tick/governor unit tests, 4 core wire-shape tests, and the new e2e)
- [x] `cargo clippy --workspace --all-targets` — clean (workspace denies unwrap/expect/unsafe)
- [x] Mock-CDP e2e `tabs_activity_serves_pushed_screencast_frames_with_acks`: two windows cast concurrently; frame 2 is only released after a correct integer ack of frame 1 per target session; foreign-session frames are never stored or acked; `/tabs/activity` serves the pushed frames
- [ ] Manual: `bun run dev:claw-rust:watch` from packages/browseros-agent, open the cockpit, watch thumbnails update live; background agent tabs still show polled frames

### DIFF
--- a/packages/browseros-agent/apps/claw-server-rust/src/routes/mod.rs
+++ b/packages/browseros-agent/apps/claw-server-rust/src/routes/mod.rs
@@ -134,6 +134,7 @@ async fn agents_cancel(
 }
 
 async fn tabs_activity(State(state): State<AppState>) -> AppResult<Json<Value>> {
+    state.screencast.note_read();
     let profiles = state.agents.list_profiles().await?;
     let tabs = state.tab_activity.snapshot().await;
     let mut enriched = Vec::with_capacity(tabs.len());
--- a/packages/browseros-agent/apps/claw-server-rust/src/services/screencast.rs
+++ b/packages/browseros-agent/apps/claw-server-rust/src/services/screencast.rs
@@ -1,32 +1,47 @@
 use crate::services::{
     browser::BrowserService,
     now_epoch_ms,
-    tab_activity::{ScreencastFrame, TabActivityService},
+    tab_activity::{ScreencastFrame, TabActivityRecord, TabActivityService},
 };
+use browseros_cdp::CdpEvent;
 use browseros_core::{
-    PageId,
+    BrowserSession, CoreError, PageId, SessionId,
+    screencast::{self, SCREENCAST_FRAME_METHOD, ScreencastOptions},
     screenshot::{ScreenshotCaptureOptions, ScreenshotFormat},
 };
+use serde_json::json;
 use std::{
-    collections::{HashMap, VecDeque},
-    sync::Arc,
+    collections::{HashMap, HashSet, VecDeque},
+    sync::{
+        Arc,
+        atomic::{AtomicI64, Ordering},
+    },
     time::Duration,
 };
 use tokio::{
-    sync::Mutex,
+    sync::{Mutex, broadcast},
     task::JoinHandle,
     time::{MissedTickBehavior, interval},
 };
 use tokio_util::sync::CancellationToken;
-use tracing::warn;
+use tracing::{debug, warn};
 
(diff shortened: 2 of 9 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1781, opened 2026-07-11
feat(browseros-agent): fetch LM Studio models live instead of a stale static list

GriggsRice:
## Summary
- The LM Studio model picker sourced from a manually-run script snapshotting models.dev's generic registry (last refreshed weeks ago), never matching what's actually loaded on a user's machine. Picking a stale suggestion sent a model ID LM Studio didn't recognize, breaking chat and Test Connection.
- Add `useLMStudioProbe`, a live query against LM Studio's native `/api/v0/models` endpoint (reports per-model load state + context length).
- The picker now shows real local models, loaded one sorted first, with accurate context-length badges; the existing context-window auto-fill now uses real values.
- Falls back to manual entry with a status hint when LM Studio is unreachable.

## Test plan
- [x] `bun run typecheck`, `bun run lint`, `bun run fallow` — clean
- [x] `bun test` (app package) — 314 pass
- [x] Manually verified live against a running LM Studio instance via CDP: dropdown shows real loaded models with accurate context-length badges, loaded model sorted first, embedding models excluded

### DIFF
--- a/packages/browseros-agent/apps/app/modules/llm-providers/lmstudio-probe.hooks.test.ts
+++ b/packages/browseros-agent/apps/app/modules/llm-providers/lmstudio-probe.hooks.test.ts
@@ -1,0 +1,96 @@
+import { describe, expect, it } from 'bun:test'
+import {
+  isLMStudioProbeEnabled,
+  parseLMStudioModels,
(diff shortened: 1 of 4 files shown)


GriggsRice:

PR #1795, opened 2026-07-13
fix(claw-onboard): let users skip the onboarding import step

shadowfax92:
## Summary

Fixes the dead end from the Discord thread: a user with no Chrome installed reached the Import step (2 of 3), saw **"No profiles found."**, a red **"Pick a profile."** error, and a **disabled** primary CTA — with no way forward and no way back.

- **Always-enabled ghost "Skip for now"** in the Import step's `picker` and `failed` phases, wired to the existing `onContinue` transition. Enabled with zero sources, with nothing selected, and while detection is still running. Import stays the visually primary action.
- **No skip in `importing`** (it would abandon a running import) or in `imported` (already has Continue).
- **Stop painting the red "Pick a profile." error when there is no profile to pick.** It was self-inflicted, not a user mistake: `OnboardingV2`'s sources effect calls `setValue('selectedSourceId', '', { shouldValidate: true })` whenever the list is empty, so the form validated itself into an error with zero user interaction.
- **`ReadyStep` stops claiming "Logins imported."** after a skip. It takes a required `imported: boolean`; `false` renders "Connect your AI." instead. Body copy is unchanged.

`failed` mattered as much as `picker`: its "Try again" is also `disabled={!isPickerValid}`, so a zero-source or repeatedly-failing user was trapped there too.

## Design

Skip reuses the existing `onContinue` prop (`() => setStep(2)`) rather than adding an `onSkip` prop, a bridge event, or new state — "skip" and "continue" are the *same* transition (advance to Ready without importing anything more). Skip deliberately routes **through** Ready rather than calling `finishOnboarding()`, because Ready carries the "Connect your AI" MCP pointer that every user needs, importer or not. `imported` is derived from existing state (`importPhase === 'imported'`) and is a **required** prop, so no call site can silently keep the old claim.

## Test plan

- [x] `bun run check` — green (lint + typecheck + fallow)
- [x] `bun run test` — green; claw-onboard **73 pass / 0 fail**
- [x] Skip renders **enabled** in `picker` with zero sources while the import CTA is disabled — i.e. an escape provably exists on the exact screen from the report
- [x] Skip renders in `picker` with sources, in `picker` while detecting, and in `failed`
- [x] No skip in `importing` or `imported`
- [x] The "Pick a profile." message is suppressed with zero sources **and** while detecting, and still shown when sources exist
- [x] `ReadyStep` renders both heading variants by the `imported` prop
- [x] Every new test mutation-checked (revert the behavior → the test goes red), so none pass vacuously

## Notes for review

- **Copy call for you, @nsonti1:** on the skip path the `<h1>` reads "Connect your **AI.**" while the primary CTA is the verbatim "Connect your AI" — the heading echoes the button instead of saying something of its own. This is the copy that was specified, so I implemented it as-is; if you'd rather it carried its own information ("You're **set up.**", "Ready to **connect.**"), it's a one-line change in `ReadyStep.tsx`.
- **Known trade-offs, both consequences of "no new state"** (accepted, not accidents): (1) between clicking Import and Chromium echoing `importing`, the phase is still `picker`, so the skip is briefly clickable — the import isn't cancelled, it completes in the background and Ready is where the user was headed anyway; (2) `imported` is derived from live state, so on the imported path `complete()` can flip the heading for a frame on the already-unmounting page.
- **Residual, out of scope:** `importing` has no controls by design. If the Chromium importer never reports a terminal status (e.g. it hangs on the Keychain prompt), the user is stuck there with no affordance. A watchdog that pushes a hung import to `failed` would close the last hole.
- **No click-level tests**, because the monorepo has no DOM test harness at all — every React test is `renderToStaticMarkup`, which strips event handlers, and `useEffect` never runs under SSR. Adding `happy-dom`/testing-library would mean a `preload` in the **root** `bunfig.toml`, injecting `window` into every server and CLI test in the repo. Not worth it for one ghost button; the tests instead pin exactly what static markup can prove — that an *enabled* way forward renders in each stuck state.

### DIFF
--- a/packages/browseros-agent/apps/claw-onboard/src/onboarding/OnboardingV2.tsx
+++ b/packages/browseros-agent/apps/claw-onboard/src/onboarding/OnboardingV2.tsx
@@ -141,7 +141,12 @@ export function OnboardingV2() {
             onContinue={() => setStep(2)}
           />
         )}
-        {step === 2 && <ReadyStep onDone={finishOnboarding} />}
+        {step === 2 && (
+          <ReadyStep
+            imported={importPhase === 'imported'}
+            onDone={finishOnboarding}
+          />
+        )}
       </OnboardingShell>
     </Form>
   )
--- a/packages/browseros-agent/apps/claw-onboard/src/onboarding/steps/ImportStep.test.tsx
+++ b/packages/browseros-agent/apps/claw-onboard/src/onboarding/steps/ImportStep.test.tsx
@@ -14,9 +14,20 @@ import {
   type OnboardingFormValues,
   onboardingFormDefaults,
   onboardingFormResolver,
+  onboardingFormSchema,
 } from '../onboarding-v2.schemas'
 import type { ImportPhase } from '../onboarding-v2.types'
 import { ImportStep } from './ImportStep'
+
+// Take the message from the resolver rather than restating it, so these tests
+// track the production copy instead of a string they hardcoded themselves.
+const unpickedResult = onboardingFormSchema.safeParse({
+  selectedSourceId: '',
+  selectedItems: [],
+})
+const PICK_PROFILE_MESSAGE = unpickedResult.success
+  ? ''
+  : (unpickedResult.error.issues[0]?.message ?? '')
 
 function readyState(
   overrides: Partial<BrowserOSOnboardingState> = {},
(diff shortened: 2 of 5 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1797, opened 2026-07-13
refactor: remove obsolete Claw agent profiles

shadowfax92:
## Summary

- remove the unused stored-agent-profile service, schemas, filesystem storage, legacy MCP registration, and per-profile harness install paths
- keep `/tabs/activity` on live session identity with slug fallback, while preserving shared harness connections and Stop/cancel behavior
- narrow startup URL migration to the shared BrowserClaw entry with crash-safe retries and exact recorded config paths
- leave the Rust port untouched

## Design

Live TypeScript agents are ephemeral MCP sessions, so activity enrichment now reads only the in-memory identity registry. The shared connection catalog and harness-to-agent mapping moved into a small live-only module. Obsolete `agents/*.json` migration and profile-owned integration code were deleted rather than shimmed.

The BrowserClaw URL migration writes a pending marker before relinking harness configs, retries all recorded links after a partial or interrupted run, preserves each manifest `configPath`, and clears the marker only after complete success.

## Test plan

- `bun run check`
- `bun run test`
- focused Claw server activity, identity, connection, cancellation, MCP, harness, and migration tests
- Claw server and Claw app typechecks
- repository-wide searches for removed profile symbols and imports

All commands pass. `bun run check` continues to print the existing non-blocking lint/Fallow inventory and exits successfully.


shadowfax92:
@greptileai review

shadowfax92:
@claude review

Issue #1701, opened 2026-07-08
Browser/Agent crash during mid/long-running tasks


darkanubis0100:
### Issue Type

Agent Issue

### Operating System

Windows

### Description of the bug

# BrowserOS Agent crashes after `Network error` on dynamic pages; likely auto-diff / snapshot / UI serialization issue

## Summary

I am seeing BrowserOS Agent become unstable after using it on a dynamic page (`Pugnatrix` in my case).

The weird part is that the machine itself is not running out of RAM. I have 96 GB RAM and the system memory barely moves before the crash. Even forcing V8 with:

```bash
--js-flags="--max-old-space-size=8192"
```

does not really solve it.

From Crashpad dumps, the crash points to the bundled BrowserOS Assistant extension, not to Bitwarden or a user-installed extension.

Extension ID seen in the dump:

```text
bflpfmnmnokmjhmgnolecpppdbdophmk
```

This matches the bundled Assistant extension ID in:

```text
apps/app/wxt.config.ts
```

```ts
// Extension ID will be bflpfmnmnokmjhmgnolecpppdbdophmk
```

## What happens

My normal flow looks like this:

1. BrowserOS is stable.
2. I open a page like `Pugnatrix`.
3. I ask the Agent to work on the page.
4. The Agent starts normally.
5. After a while, the side panel shows:

```text
Something went wrong - Network error
```

6. If I try to continue / retry the same task, sometimes only the Agent crashes.
7. Other times the whole BrowserOS renderer / browser crashes.

This did not happen to me with other long tasks, for example searching flights. So it feels very page-dependent, not simply “long task = crash”.

## Crash evidence

I am not attaching full `.dmp` files because they can contain private page/session data.

Relevant details from the Crashpad dumps:

```text
Process type: renderer
Exception: 0xe0000008
V8 OOM: out of memory
v8-oom-location: CALL_AND_RETRY_LAST
v8-oom-stack: stringify
Extension: bflpfmnmnokmjhmgnolecpppdbdophmk
JS chunk: chunks/mermaid-*.js
```

After launching with the 8 GB V8 flag, the dump confirmed the flag was applied:

```text
--js-flags=--max-old-space-size=8192
```

But the crash still happened, with memory looking more like V8 heap / cage reservation / fragmentation than actual system RAM exhaustion:

```text
v8-oom-main-cage-last-alloc-status = ran out of reservation
v8-oom-main-cage-size = 4096.00MB
[...]

PR #1802, opened 2026-07-13
fix(app): keep side panel open across worker restarts

shadowfax92:
## Summary

- make routine MV3 worker rehydration restore the side-panel scope in memory without writing Chrome options
- keep explicit settings changes applying the same enabled/path options, with the existing epoch guard
- establish the stored/fallback option once on fresh extension installation so the manifest default and storage default agree
- cover passive hydration, storage failure, install initialization, explicit changes, and stale async reads

Contributes a scoped fix to #1701 based on the trace in discussion #1768.

## Caller map

- Background worker startup calls ensureSidePanelRuntimeStateLoaded, which now restores only the scope cache and session-backed open-window IDs.
- The settings sidePanelScopeChanged message calls setSidePanelPerWindowPreference, which remains the explicit setOptions path.
- OnInstalledReason.INSTALL calls initializeSidePanelOptions once to establish the fresh-install default.
- onOpened/onClosed listeners continue to persist per-window open state and do not configure options.

## Chrome behavior verified

Chrome documents setOptions without a tabId as the default side-panel configuration, while tab-specific options override it: https://developer.chrome.com/docs/extensions/reference/api/sidePanel#method-setOptions

CDP verification refined the original report:

- setOptions({ enabled: false }) closes an open inherited/window panel and emits onClosed
- it does not close a panel with its own tab-specific enabled override
- default and tab-specific options survive an MV3 service-worker restart
- the configured default also survived a complete same-profile browser restart

So this removes the verified startup mutation that can close inherited/window panels; it does not claim to fix every stream interruption tracked by #1701.

## Manual verification

1. Started a clean profile with bun run dev:watch -- --new and confirmed fresh install registered the fallback as enabled: false.
2. Opened an inherited/window panel, kept storage false, and deliberately left Chrome default options enabled to reproduce the dangerous restart state.
3. Closed service-worker target 9DDA... through CDP. Panel target DDB5... stayed present.
4. Woke the extension from the panel. New worker ED3A... appeared while the same DDB5... panel target remained open; options stayed enabled and session open-window state remained intact.
5. Applied enabled: false explicitly. Chrome emitted onClosed and removed the panel target.
6. Stored/applied window scope true in the stable dev profile, stopped the entire browser, and relaunched the same profile. Storage and getOptions both remained true. Restored the profile to false afterward.

## Test plan

- [x] bun test ./apps/app/lib/browseros/toggleSidePanel.test.ts — 13 passed
- [x] cd apps/app && bun run typecheck
- [x] cd apps/app && bun run test — 297 passed
- [x] bun run check from packages/browseros-agent — passed with the existing non-failing warning inventory
- [x] bun run test from packages/browseros-agent — all groups passed


shadowfax92:
@claude review

shadowfax92:
@greptileai review

PR #1803, opened 2026-07-13
fix: bound snapshot diff LCS allocation

shadowfax92:
## Summary

- trim equal snapshot prefixes and suffixes before running the existing LCS diff
- cap the actual padded LCS table at 4,000,000 cells and return a bounded summary above that limit
- render summary-only recovery guidance outside the untrusted page-content boundary
- cover compatibility, allocation guards, exact and asymmetric budget boundaries, edge cases, and MCP rendering

The change keeps the existing LCS algorithm for ordinary diffs. Over-budget results preserve the required `SnapshotDiff` fields and add `lineDiffSkipped` so consumers can distinguish changed-window line counts from a computed line diff.

Related to #1701. This addresses the server-side snapshot-diff stall; the separate extension-renderer OOM is outside this PR.

## Test plan

- [x] `bun run check` from `packages/browseros-agent`
- [x] `bun run test` from `packages/browseros-agent`
- [x] focused browser-core and browser-mcp diff tests
- [x] three-round adversarial review, final round clean


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1804, opened 2026-07-13
fix: default browser tests to headless

shadowfax92:
## Summary
- default BrowserOS browser-backed tests to headless when `BROWSEROS_TEST_HEADLESS` is unset
- preserve explicit `false` as headed mode and cover both real launch argument paths
- align the shared env registry and generated development example with the headless default

## Design
The existing test runtime remains the single policy boundary: it resolves all values except exact `false` to headless, then both browser-backed test paths pass that explicit boolean into the unchanged generic browser spawn helper.

## Test plan
- [x] `bun test ./apps/server/tests/test-runtime.test.ts ./apps/server/tests/browser-launch-args.test.ts`
- [x] `env -u BROWSEROS_TEST_HEADLESS bun run --filter @browseros/server test:root`
- [x] `bun run env:examples:check`
- [x] `bun run check`
- [x] `env -u BROWSEROS_TEST_HEADLESS bun run test`
- [x] explicit shell `true` and `false` override opposite `--env-file` values
- [x] context-free adversarial review: clean

### DIFF
--- a/packages/browseros-agent/.env.development.example
+++ b/packages/browseros-agent/.env.development.example
@@ -86,4 +86,4 @@ LOG_LEVEL=info
 # BROWSEROS_AI_SDK_DEVTOOLS=true
 
 # Headless browser setting for local server tests.
-BROWSEROS_TEST_HEADLESS=false
+BROWSEROS_TEST_HEADLESS=true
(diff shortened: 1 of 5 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1805, opened 2026-07-13
fix(app): avoid transcript stringify OOM

shadowfax92:
## Summary

- replace whole-message-array `JSON.stringify` equality checks with bounded terminal-message change detection
- preserve no-op saves, streaming append/content detection, and the 50-conversation cap without changing stored message shapes
- debounce and deduplicate full-list migration uploads across hook instances, in-flight work, hydration, logout, and account changes

Extension-side fix for #1701. The GraphQL transport still performs its required request-body serialization; this removes the redundant renderer-side transcript equality serialization and upload churn.

## Test plan

- `cd packages/browseros-agent/apps/app && bun run typecheck && bun run test`
- `cd packages/browseros-agent && bun run check && bun run test`


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1807, opened 2026-07-13
fix: make MCP structured content opt-in

shadowfax92:
## Summary

- make browser-tool `structuredContent` a wire-level opt-in: schema-declared tools retain it, `/mcp?structured=1` enables it for machine clients, and ordinary chat-harness endpoints omit it
- keep BrowserClaw's internal structured envelopes available to audit, screenshot, ownership, and grouping effects, then strip them from the returned wire result
- restore grep's bounded `matches` machine contract, preserve actionable trailing refs while clamping, and align spill metadata with the other text tools
- pin every browser tool's exact structured envelope keys so machine-consumer regressions are visible in TypeScript tests

## Design

Browser tools continue producing their existing envelopes. Presentation is decided only at the registration/dispatch boundary: `structuredContent` is included when the tool declares an `outputSchema` (currently only `run`) or when the server endpoint explicitly opts in. The apps/server MCP route defaults browser tools to content-only and accepts `?structured=1`; the CLI uses that query contract. BrowserClaw has only chat-harness clients today, so it strips structured output after all internal effects and has no opt-in plumbing.

No output schemas were added, and full payload mirrors were not restored for snapshot, diff, read, or evaluate.

## Claude Code behavior

Claude Code versions in the affected behavior range surface only `structuredContent` when both result channels are present, while falling back to `content` when structured output is absent. Upstream reports:

- https://github.com/anthropics/claude-code/issues/55677
- https://github.com/anthropics/claude-code/issues/15412
- https://github.com/anthropics/claude-code/issues/9962

## Machine client contract

Machine clients that consume browser-tool envelopes must connect to `/mcp?structured=1`. The BrowserOS CLI now does so. Grep again includes the bounded rendered `matches` array that `find` and batch find steps parse for `[ref=eN]` targets.

## Rollout

The server's strip-by-default behavior and the CLI's `?structured=1` endpoint change must ship in the same release train. Already-installed CLIs will temporarily lose structured browser fields after the server update until their updater under `apps/cli/update` brings them forward.

## Test plan

- [x] `bun run check`
- [x] `bun run test`
- [x] browser-mcp, apps/server, and BrowserClaw focused contract/integration tests
- [x] `gofmt`, `go vet ./...`, `go build ./...`, and `go test ./...` in `apps/cli`

### DIFF
--- a/packages/browseros-agent/apps/claw-server/src/mcp/dispatch.ts
+++ b/packages/browseros-agent/apps/claw-server/src/mcp/dispatch.ts
@@ -218,7 +218,7 @@ async function dispatchToolCall(
   effects: readonly NamedToolEffect[],
 ): Promise<ToolResult> {
   const rejection = runGuards(call)
-  if (rejection) return rejection
+  if (rejection) return wireResult(rejection)
   const connectedCall = call as ConnectedToolCall
 
   if (ARBITRARY_SCRIPT_TOOLS.has(call.tool.name)) {
@@ -238,10 +238,13 @@ async function dispatchToolCall(
     })
   }
   const result = runEffects({ call, ...outcome }, effects)
+  return wireResult(result)
+}
+
+function wireResult(result: ToolResult): ToolResult {
   return {
     content: result.content,
     isError: result.isError,
-    structuredContent: result.structuredContent,
   }
 }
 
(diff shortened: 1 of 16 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1827, opened 2026-07-14
fix(claw-server): paginate listTasks by session via HAVING, not row WHERE

Osamaali313:
## Problem

`listTasks` (`packages/browseros-agent/apps/claw-server/src/services/tasks.ts`) groups tool dispatches by session and paginates on **each session's `max(id)`** (`orderBy desc(max(id))`, `nextCursor = last session's maxId`). But the cursor was applied as a **row-level** filter *before* `GROUP BY`:

```ts
if (typeof query.cursor === 'number') {
  wheres.push(lt(toolDispatches.id, query.cursor))
}
...
  .where(wheres.length > 0 ? and(...wheres) : undefined)
  .groupBy(toolDispatches.sessionId)
  .orderBy(desc(sql`max(${toolDispatches.id})`))
```

A session already returned on an earlier page that still owns dispatch rows with `id < cursor` gets re-grouped on a later page as a **duplicate task**, with truncated `dispatchCount` / `startedAt` / `toolSequence` / `errorCount` (aggregates now computed over only the surviving rows). This happens whenever two sessions' dispatch ids interleave — the norm with parallel agents.

## Reproduction

Two busy interleaved sessions `S1` (ids {1,5}), `S2` (ids {2,6}) plus short `S3`,`S4`, `limit=2`:

- **Before:** page1 `S1(2),S2(2)` → page2 `S4(1),S3(1)` → page3 `S2(1),S1(1)` — **S1 and S2 returned twice, counts truncated 2→1.**
- **After:** page1 `S1(2),S2(2)` → page2 `S4(1),S3(1)` — each session once, full counts, pagination terminates.

## Fix

Move the cursor from a row-level `WHERE` to a group-level `HAVING max(id) < cursor`, so the pagination filter matches the pagination key:

```ts
  .where(wheres.length > 0 ? and(...wheres) : undefined)
  .groupBy(toolDispatches.sessionId)
  .having(
    typeof query.cursor === 'number'
      ? lt(sql`max(${toolDispatches.id})`, query.cursor)
      : undefined,
  )
  .orderBy(desc(sql`max(${toolDispatches.id})`))
```

`lt` and `sql` are already imported; `.having()` is a builder method — no new imports.

## Test

Added a regression test to `tests/services/tasks.test.ts` with interleaved multi-dispatch sessions that walks every page and asserts no session appears twice and counts aren't truncated. It fails on the current code (a session is seen on two pages) and passes after the fix. Full `tasks.test.ts` suite: **15 pass** (`bun test`).

The existing `paginates via cursor` test passes today only because it seeds exactly one dispatch per session, so no session can straddle the cursor — the multi-dispatch interleaved case was uncovered.


Osamaali313:

DaniAkash:
@greptileai

PR #1828, opened 2026-07-14
feat: publish extension releases and both feed channels

shadowfax92:
## Summary
- create or update `extensions/v<version>` GitHub releases after manual CRX builds, with notes, undrafting, and clobbered CRX assets
- preserve the CRX/feed split and keep reusable product callers opted out by default while granting compatible permissions
- add `channel=both` to publish alpha then prod with the same feed arguments

## Design
The workflow mirrors the existing claw-server release lifecycle locally in `release-extensions.yml`, using the built-in GitHub token and targeting the checked-out SHA only when creating a new tag. Feed publishing remains solely in `release-extension-feeds.yml`; its existing args array is reused sequentially for alpha and prod.

## Test plan
- [x] `actionlint .github/workflows/release-extensions.yml .github/workflows/release-extension-feeds.yml .github/workflows/release-browseros.yml .github/workflows/release-browserclaw.yml`
- [x] shell harness verifies new-release, existing-draft, notes, and CRX upload commands
- [x] shell harness verifies unchanged alpha/prod commands and ordered alpha-then-prod behavior for `both`
- [x] adversarial diff review completed with no kept findings

### DIFF
--- a/.github/workflows/release-browserclaw.yml
+++ b/.github/workflows/release-browserclaw.yml
@@ -323,7 +323,7 @@ jobs:
       }}
     uses: ./.github/workflows/release-extensions.yml
     permissions:
-      contents: read
+      contents: write
     with:
       version: ${{ needs.preflight.outputs.extensions_version }}
       extension: browserclaw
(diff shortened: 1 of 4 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1846, opened 2026-07-15
fix(claw-server): default audit created_at in JS so it records on SQLite < 3.42

### LINKED ISSUE #1845: BrowserClaw audit log never records on SQLite < 3.42 (created_at NOT NULL: unixepoch('subsec'))
## Summary
The BrowserClaw cockpit **Audit** page stays empty ("the audit is quiet so far") even when an agent connects via MCP and runs tools. No sessions or dispatches are ever recorded.

## Root cause
The audit DB `created_at` columns (`tool_dispatches`, `agent_session_starts`, `agent_session_ends`) default to `(unixepoch('subsec') * 1000)`. The `subsec` modifier to `unixepoch()` was added in **SQLite 3.42.0**. On older SQLite the whole expression evaluates to **NULL**. The audit writer (`recordToolDispatch`) omits `created_at`, relying on that default, and the column is `NOT NULL`. So every insert fails with:

```
NOT NULL constraint failed: tool_dispatches.created_at
```

The write path is best-effort (`catch` → `logger.warn('audit log write failed', …)`), so it fails silently and the audit stays empty.

Key detail: on macOS, `bun:sqlite` links the **system** SQLite (often < 3.42), while CI runs a newer SQLite. So `apps/claw-server/tests/modules/db/db.test.ts` (which inserts relying on the default) passes in CI, but the shipped desktop build never writes a row.

## Reproduction
On any SQLite < 3.42 (e.g. macOS system SQLite 3.39.5, which `bun:sqlite` uses):
```
sqlite> SELECT unixepoch('subsec') * 1000;    -- (NULL)
sqlite> CREATE TABLE t(created_at integer DEFAULT (unixepoch('subsec')*1000) NOT NULL, x text);
sqlite> INSERT INTO t(x) VALUES('hi');         -- Error: NOT NULL constraint failed: t.created_at
```
Also: `bun test apps/claw-server/tests/modules/db/db.test.ts` on macOS → the insert-based tests fail with the same error.

Server log:
```
{"level":40,"msg":"audit log write failed","tool":"navigate","error":"NOT NULL constraint failed: tool_dispatches.created_at"}
```
and the audit DB has 0 rows in all three tables.

## Expected
Tool dispatches and session start/end events record regardless of the runtime SQLite version; the Audit page populates.

## Environment
- BrowserOS/BrowserClaw 0.47.14.0 (Chromium 148.0.7962.97)
- macOS (Darwin 23.1)
- `bun:sqlite` → SQLite 3.39.5

## Fix
Default `created_at` in JS (`$defaultFn(() => Date.now())`) instead of a SQLite-version-dependent SQL expression. PR incoming.

thomasxm:
## Problem

The BrowserClaw cockpit **Audit** page stays empty ("the audit is quiet so far") even when an agent connects over MCP and runs tools — no sessions or dispatches are ever recorded.

**Root cause.** The audit DB `created_at` columns (`tool_dispatches`, `agent_session_starts`, `agent_session_ends`) default to `(unixepoch('subsec') * 1000)`. The `subsec` modifier to `unixepoch()` was only introduced in **SQLite 3.42.0**; on any older engine the whole expression evaluates to **NULL**. The audit writer (`recordToolDispatch`) omits `created_at` and relies on that default, and the column is `NOT NULL` — so every insert fails with:

```
NOT NULL constraint failed: tool_dispatches.created_at
```

The write path is best-effort (`catch` → `logger.warn('audit log write failed', …)`), so it fails silently and the audit never populates.

**Who is affected.** On macOS, `bun:sqlite` links the *system* SQLite, and current macOS ships an Apple build older than 3.42 (3.39.5 on my machine — the `bun:sqlite` version and the system `sqlite3` CLI report the same build). So this is not a local misconfiguration: **every macOS user on a system SQLite < 3.42 gets a permanently empty audit.** Linux/Windows builds generally use a newer SQLite where `subsec` works, which is also **why CI (Linux) never caught it** — the existing default test in `db.test.ts` passes there while the shipped macOS build never writes a row.

## Reproduction

On any SQLite < 3.42 (e.g. the macOS system SQLite 3.39.5 behind `bun:sqlite`):

```
sqlite> SELECT unixepoch('subsec') * 1000;    -- (NULL)
sqlite> CREATE TABLE t(created_at integer DEFAULT (unixepoch('subsec')*1000) NOT NULL, x text);
sqlite> INSERT INTO t(x) VALUES('hi');         -- Error: NOT NULL constraint failed: t.created_at
```

Equivalently, `bun test apps/claw-server/tests/modules/db/db.test.ts` on macOS fails the insert-based tests with the same error before this change.

## Fix

Default `created_at` in JS via `.$defaultFn(() => Date.now())` on the three audit tables. `Date.now()` is epoch-milliseconds — the same unit and intent as the original SQL default — but it is computed in the runtime, so it no longer depends on the SQLite version. Every insert already goes through Drizzle, so the value is always supplied.

I also regenerated the migration to drop the now-unreliable SQL default (a standard Drizzle row-preserving table rebuild), and added a **version-independent** regression test that asserts the JS default is present — it fails even on a newer SQLite if the SQL-only default is ever reintroduced.

## Tests

- `bun test` (claw-server): **399 pass, 0 fail**
- `bun run typecheck`: clean
- `bunx biome check`: clean
- Before this change, on macOS `bun test`, the insert-based tests fail with `NOT NULL constraint failed`; afterwards all pass.

## Behaviour and compatibility

- No semantic change for consumers: `created_at` was, and remains, epoch-milliseconds. Readers (`listDispatches`, the task deriver) page by `id` and treat `created_at` as milliseconds, so they are unaffected.
- Existing installs: the migration is row-preserving. Installs on SQLite < 3.42 have no rows anyway (the bug prevented writes); installs on a newer SQLite keep their rows.

## Not in this PR

`apps/claw-server-rust` carries the same `unixepoch('subsec')` default in its migrations and also omits `created_at` on insert. Its `rusqlite` typically bundles a newer SQLite, so it is not currently broken, but it shares the same latent version dependency. I have left it out to avoid rewriting already-applied Rust migration checksums — happy to follow up separately if you would like it aligned.

Fixes #1845


thomasxm:

DaniAkash:
@greptileai

PR #1849, opened 2026-07-16
fix(audit): group Agent filter by slug, not per-connection agentId

DaniAkash:
## What was broken

On the audit page, the **Agent** filter dropdown lists each MCP connection as its own entry:

- \`codex-mcp-client\` shows up 15+ times, each with count \`1\`
- \`browserclaw-claude-desktop-wrapper\` shows up 3+ times, each with count \`1\`
- \`claude-code\` shows once

Every reconnection of the same client creates a fresh row instead of incrementing a single \"codex-mcp-client (N)\" entry.

## Root cause

\`agentChipsFor\` in \`packages/browseros-agent/apps/claw-app/screens/audit/audit.helpers.ts\` dedupes tasks by \`t.agentId\`. But \`agentId\` is minted per MCP session/connection; the field that identifies the agent kind is \`slug\`. So the map's keys were effectively unique per session, and the \"Agent\" filter turned into a session list.

## The fix

Group by \`slug\` everywhere in the audit filter path.

**Client**
- \`AgentChip\` drops \`agentId\`, keyed by \`slug\`
- \`agentChipsFor\` maps by \`t.slug\`
- \`AuditFilters.agentId\` -> \`AuditFilters.agentSlug\` (URL search-param key stays \`agent\` for shareable-link compatibility; only the stored value semantics changed)
- \`FilterBar\` prop rename: \`selectedAgentId\` -> \`selectedAgentSlug\`
- \`useTasks\` gains a \`slug?: string\` variable that maps to a \`slug\` query param
- Test fixtures updated to the new field names

**Server**
- \`GET /audit/tasks\` accepts an optional \`slug\` query param
- \`listTasks\` predicates \`eq(toolDispatches.slug, query.slug)\` when provided
- Existing \`agentId\` filter kept (2 lines added, nothing removed)

## Deliberately not touched

- Cockpit's live-agent grid (\`RunningGrid.tsx\`, \`cockpit.helpers.ts\`) still uses \`agentId\`. That's correct there: focus / cancel actions target a specific live session, not an agent type.

## Backwards compat

Old \`?agent=<oldAgentId>\` bookmarks will silently show no results (an agentId won't match any slug). Acceptable because the field was buggy: those bookmarks were already showing a single random session's worth of data mislabelled as the entire agent's history.

## Test plan

- [x] \`bun run typecheck\` (claw-app): pass
- [x] \`bun run typecheck\` (claw-server): pass
- [x] \`bun test screens/audit\`: 9/9 pass
- [x] biome check on changed files: 3 pre-existing warnings on untouched lines in \`services/tasks.ts\`; no new warnings
- [ ] Manual verification: open /audit, open Agent dropdown, confirm each agent type appears exactly once with a correct total count; picking one filters the list to that agent kind

### DIFF
--- a/packages/browseros-agent/apps/claw-app/components/audit/FilterBar.tsx
+++ b/packages/browseros-agent/apps/claw-app/components/audit/FilterBar.tsx
@@ -19,11 +19,11 @@ interface FilterBarProps {
   agentOptions: AgentChip[]
   statusOptions: { status: TaskStatus; count: number }[]
   siteOptions: { site: string; count: number }[]
-  selectedAgentId: string | null
+  selectedAgentSlug: string | null
   selectedStatus: TaskStatus | null
   selectedSite: string | null
   search: string
-  onAgentChange: (agentId: string | null) => void
+  onAgentChange: (slug: string | null) => void
   onStatusChange: (status: TaskStatus | null) => void
   onSiteChange: (site: string | null) => void
   onSearchChange: (q: string) => void
@@ -33,7 +33,7 @@ export function FilterBar({
   agentOptions,
   statusOptions,
   siteOptions,
-  selectedAgentId,
+  selectedAgentSlug,
   selectedStatus,
   selectedSite,
   search,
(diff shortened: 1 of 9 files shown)


DaniAkash:
thought: confirmed and fixed in 00bfb3d96. You were right — the earlier commit dropped the schema line and I missed it because tests mock `audit.data` at the frontend layer so the HTTP roundtrip was never exercised. `slug: z.string().optional()` is now in `listQuerySchema`; the client's slug query param now passes through the zValidator into `listTasks`.

PR #1859, opened 2026-07-17
feat(claw-server): self-healing per-host recipes (behind flag)

DaniAkash:
## What ships

A convention-based self-healing mechanism for BrowserClaw's MCP dispatch loop: a coding-agent MCP client can leave a per-site Markdown note once and read it on every future visit to the same host. The mechanism is a filesystem convention plus one dispatch effect plus one appended instructions block. No engine, no ML, no schema migration, no UI.

Off by default behind `BROWSERCLAW_RECIPES=1` for a careful rollout.

## The two commits

**1. `feat(claw-server): add recipes service + domain-skills-hint effect (behind flag)`**

- `services/recipes.ts`: `hostStemFromUrl` (`hostname` after stripping `www.`, then the first dot-separated label, filtered to `http:` / `https:`), `recipesDirFor`, `listRecipeFiles` (sorted, capped at 10, subdir / non-file safe).
- `mcp/effects/domain-skills-hint.ts`: a new `NamedToolEffect` that fires on `navigate` when the flag is on and annotates the result's `structuredContent` with `domain_skills.{files, workspace_dir}` while also appending a synthetic `content[]` text block for clients that only pass `content` to the LLM. Registered in `BASE_EFFECTS` between `tabs-list-view` and `audit` so the annotation lands in `resultMeta` and the operator sees which recipes each agent was told about.
- `env.ts`: `BROWSERCLAW_RECIPES` flag via a new `readBoolFlagDefaultFalse` helper.
- Tests: 12 unit tests on the service (host-stem edge cases, cap enforcement, symlink following, subdir skipping, multi-agent + multi-host isolation) + 10 effect tests (flag gating, non-navigate passthrough, error passthrough, no-identity guard, `chrome://` no-op, cold host hint, warm host hint, args-url fallback, structuredContent field preservation, multi-agent isolation).

**2. `feat(claw-server): append recipe discipline to MCP instructions when flag is on`**

- `mcp/mcp-prompt.ts`: adds a `Domain recipes:` block that tells the agent to `Read` every `domain_skills.files` before acting on the site and to `Write` new `.md` notes to `workspace_dir` when it figures something out. Refactored to `getBrowserClawMcpInstructions()` so the flag composition happens at call time.
- `mcp/single-server.ts` (+1 -1): switched the McpServer constructor to the getter.
- Tests: 2 unit tests covering flag-off (base only) and flag-on (base + recipes tail) shapes.

## Storage layout

Recipes live at `<browserclawDir>/recipes/<slug>/<host-stem>/<name>.md`, alongside the existing `audit.sqlite` and `replays/`. Slug comes from `identityService` so recipes are scoped per MCP client (`claude-code`, `codex-mcp-client`, etc.). Host stem is deliberately the same reduction upstream libraries use, so a folder is portable in either direction.

## Writes

Coding-agent MCP clients (Claude Code, Codex, Cursor, VS Code, Zed, OpenCode, Antigravity) use their own `Read` / `Write` tools. That is why this PR ships zero new BrowserClaw MCP tools: the write path exists in the client, not the server. Clients that lack filesystem access (Claude Desktop-like) are a no-op today; a dedicated MCP `save_recipe` tool for those clients is a follow-up if needed.

## Full suite

- `bun test` in `apps/claw-server`: **424 pass, 0 fail** (was 400, +24 new tests across 3 files)
- `bun run typecheck`: clean
- `bunx biome check`: clean

## What is deliberately NOT in this PR

- No new MCP tools (`list_recipes`, `read_recipe`, `save_recipe`). Coding-agent clients use their own filesystem tools.
- No `apps/claw-app/` UI. Browsing / deleting recipes is a follow-up.
- No SQLite schema. Filesystem-first.
- No cross-agent shared recipe pool. Recipes are per-slug.
- No auto-heal-on-failure detection.

## Rollback

`unset BROWSERCLAW_RECIPES` (or set to `0` / `false`). The effect is a cheap early-return with no filesystem I/O; the instructions revert to the base block; existing recipe files stay on disk but are never read by the server until the flag flips back on.

## Manual verification

1. `BROWSERCLAW_RECIPES=1 bun run start` inside `apps/claw-server`.
2. Connect Claude Code via the standard MCP config.
3. Call `navigate("https://news.ycombinator.com")`. Confirm the tool result carries `structuredContent.domain_skills.workspace_dir` pointing at the per-slug per-host directory.
4. Ask Claude Code: "save a recipe named `frontpage-selectors.md` at that path with `# HN\n- .titleline > a for top story titles`".
5. Navigate again to the same URL. Confirm `domain_skills.files` includes `frontpage-selectors.md` and the synthetic text block names it.
6. Kill the session, start a new one, repeat step 5. Confirm persistence.
7. Toggle `BROWSERCLAW_RECIPES=0`, restart, navigate. Confirm no `domain_skills` field appears and the instructions no longer include the `Domain recipes:` block.

## Test plan

- [x] `apps/claw-server` unit tests green locally
- [x] Typecheck + biome clean
- [ ] CI green
- [ ] Manual end-to-end verification against a real MCP client before the flag flips on-by-default

## Follow-ups (separate work)

- A cockpit UI to browse and delete recipes (`shadcn` AlertDialog with typed-confirmation for delete, matching the audit-cleanup UX).
- Optional MCP tools for clients that lack filesystem access.
- User-facing docs.
- Interaction check with the audit-cleanup orphan sweep once that lands.

### DIFF
--- a/packages/browseros-agent/apps/claw-server/src/env.ts
+++ b/packages/browseros-agent/apps/claw-server/src/env.ts
@@ -50,19 +50,6 @@ function readBoolFlagDefaultTrue(name: string): boolean {
   return normalised !== '0' && normalised !== 'false'
 }
 
-/**
- * Opt-IN gate for features that stay OFF by default. Only `1` and
- * `true` (case-insensitive) enable. Used for careful rollouts where
- * we want a hard-off default until the feature has soaked.
- */
-function readBoolFlagDefaultFalse(name: string): boolean {
-  // biome-ignore lint/style/noProcessEnv: env.ts is the sanctioned env-reader for the package
-  const raw = process.env[name]
-  if (raw === undefined) return false
-  const normalised = raw.trim().toLowerCase()
-  return normalised === '1' || normalised === 'true'
-}
-
 /** Trims a string value, returning undefined when it is unset or blank. */
 function trimmedString(value: string | undefined): string | undefined {
   const raw = value?.trim()
@@ -101,12 +88,6 @@ export const env = {
   screencastScreenshotFallback: readBoolFlagDefaultTrue(
     'CLAW_SCREENCAST_SCREENSHOT_FALLBACK',
   ),
-  // Per-host recipes: when on, the `navigate` dispatch effect
-  // surfaces per-agent per-host Markdown filenames in the tool result
-  // and the MCP Initialize response tells the agent to read them
-  // first. Off by default for a careful rollout; flip to default-on
-  // after production soak.
-  recipesEnabled: readBoolFlagDefaultFalse('BROWSERCLAW_RECIPES'),
   // Anonymous product analytics (PostHog). Disabled unless a project
   // write key is provided (production builds inject it). `posthogHost`
   // defaults to PostHog US Cloud. `analyticsEnabledByEnv` is an
--- a/packages/browseros-agent/apps/claw-server/src/env.ts
+++ b/packages/browseros-agent/apps/claw-server/src/env.ts
@@ -101,12 +101,11 @@ export const env = {
   screencastScreenshotFallback: readBoolFlagDefaultTrue(
     'CLAW_SCREENCAST_SCREENSHOT_FALLBACK',
   ),
-  // Self-healing recipes: when on, the `navigate` dispatch effect
+  // Per-host recipes: when on, the `navigate` dispatch effect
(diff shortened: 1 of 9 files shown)


DaniAkash:
Converting this branch into an epic. Future work will target `feat/recipes-phase-1` directly, then the whole thing merges to `main` as one PR when complete. Closing this PR now; the branch stays live.

PR #1872, opened 2026-07-17
fix: route BrowserOS catalog extension updates through CDN

shadowfax92:
## Summary
- route every active BrowserOS product-catalog extension through the product CDN update manifest
- select the stable or alpha manifest from `kBrowserOsAlphaFeatures`
- heal already-installed BrowserClaw copies whose CRX manifest omits `update_url`

## Design
`ExtensionManagement::GetEffectiveUpdateURL` is the shared browser-side choke point for installed-extension update routing. Returning the product manifest before policy and manifest fallback keeps default-installed catalog extensions out of the policy override path and makes channel changes effective on the next update check. Non-catalog extensions are unchanged.

## Code trace
`BrowserOSExtensionMaintainer::ForceUpdateCheck` calls `ExtensionUpdater::CheckNow`, which builds downloader tasks using `ExtensionManagement::GetEffectiveUpdateURL`. For the active BrowserClaw catalog ID, stable now returns `kBrowserOSUpdateUrl` and alpha returns `kBrowserOSAlphaUpdateUrl`. An empty URL previously fell through to the Chrome Web Store default in `ExtensionDownloader::SanitizeUpdateURL`; it no longer reaches that fallback for catalog extensions. The empty-URL condition at `extension_updater.cc:404` only filters converted user scripts and was not the Claw failure gate.

## Test plan
- [x] `/Users/shadowfax/.skills/sf-mux/scripts/sfmux pool build -- autoninja -C out/Default_browseros_arm64 chrome`
- [x] extracted `browseros..task/claw-extension-update-url` with `bpatch extract`
- [x] `verify-patches.sh`: 1/1 byte-match and 1/1 apply-check
- [x] reviewed `GetEffectiveUpdateURL` and `UpdatesFromWebstore` consumers; the CDN remains non-webstore and preserves existing off-store behavior

### DIFF
--- a/packages/browseros/chromium_patches/chrome/browser/extensions/extension_management.cc
+++ b/packages/browseros/chromium_patches/chrome/browser/extensions/extension_management.cc
@@ -1,5 +1,5 @@
 diff --git a/chrome/browser/extensions/extension_management.cc b/chrome/browser/extensions/extension_management.cc
-index bea1864156f76..cad92ebeb592f 100644
+index bea1864156f76066fe88de154d06ef5070f6f0ac..9eaef771c6bf35b81c9786c4bb0150a3db05fc2c 100644
 --- a/chrome/browser/extensions/extension_management.cc
 +++ b/chrome/browser/extensions/extension_management.cc
 @@ -25,6 +25,8 @@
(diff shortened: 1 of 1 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1881, opened 2026-07-18
fix(tests): preserve exports in Bun mocks

shadowfax92:
## Summary
- avoid runtime WXT-alias imports when registering Bun module mocks
- preserve real exports for the partial provider-template mock
- make total replacements exhaustive so process-scoped mocks cannot drop exports

## Design
Bun 1.3.6 cannot resolve the WXT alias for a real runtime import in this test context. Partial mocks import the real module by relative path; complete mocks use `satisfies typeof import(...)` to keep the full module contract without a runtime alias import.

## Test plan
- [x] `bun test apps/app/modules/llm-providers/oauth-provider-flow.hooks.test.ts`
- [x] `bun test apps/app/modules/llm-providers/acp-probe.hooks.test.ts`
- [x] `bun run --cwd=apps/app typecheck`
- [x] `bun run test`


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1883, opened 2026-07-18
feat: make recording delivery resilient to tab teardown

shadowfax92:
## Summary
- keep destroyed tab-to-target mappings resolvable for a five-minute recording flush window while closing claims immediately
- deduplicate retried recording batches per target with a bounded 256-entry LRU
- preserve headerless clients and the existing recording wire contracts

## Design
The tab identity map retains a forward-only grace entry after target destruction and resolves live state before grace before CDP fallback. Recording batch IDs are checked, appended, and remembered inside the existing per-target write queue, so concurrent retries cannot double-append and failed appends remain retryable.

## Test plan
- [x] `bun run check`
- [x] `bun run test`
- [x] focused grace, claim-release, duplicate, concurrency, failure-retry, headerless, isolation, and LRU tests


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1884, opened 2026-07-18
feat(claw-app): make recording delivery resilient

shadowfax92:
## Summary

- replace probe-gated recording uploads with attempt-first delivery and bounded per-tab retry queues
- recover from dropped batches with stable batch IDs and fresh rrweb checkpoints
- show honest replay states for incomplete streams and tabs without a FullSnapshot

## Design

Recording POST outcomes now drive relay state. Transient failures retain batches in a 10 MB in-memory queue and retry every five seconds while preserving per-tab FIFO order. A confirmed 404 enters the ten-minute legacy interval, while every affected tab remains marked gapped so later recovery requests a FullSnapshot. The player starts at the first usable checkpoint and surfaces any omitted prefix.

## Test plan

- `bun run check`
- `bun run test`
- `bun run --filter @browseros/claw-app build`
- verified generated Chrome MV3 permissions and host permissions remain unchanged
- two adversarial review rounds; final round clean

### DIFF
--- a/packages/browseros-agent/apps/claw-app/entrypoints/background.ts
+++ b/packages/browseros-agent/apps/claw-app/entrypoints/background.ts
@@ -14,6 +14,7 @@ export default defineBackground(() => {
     resolveServerBaseUrl: resolveBrowserOSServerBaseUrl,
   })
   const requestResnapshot = (tabId: number) => {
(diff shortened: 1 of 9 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1887, opened 2026-07-18
test: lock Rust recording reliability parity

shadowfax92:
## Summary
- align the Rust destroyed-target grace constant with the TypeScript `GRACE_MS` contract and preserve the tab-id non-reuse invariant
- port the missing retry-dedupe matrix for target isolation, absent headers, concurrent retries, post-catalog failure recovery, and true LRU eviction at 257 entries
- verify mixed-case `X-Recording-Batch-Id` retries return the existing successful `accepted: 0` response without duplicate events

## Design
`21815a767` already contained both reliability mechanisms, so this keeps the shipped Rust implementation intact and adds the exact cross-port naming and regression contracts. Dedupe stays in the per-target serialized append boundary and remains in-memory; destroyed-target grace remains append-resolution-only while claims close immediately.

## Test plan
- [x] `cargo fmt --check`
- [x] `cargo build -p claw-server-rust`
- [x] `cargo clippy -p claw-server-rust`
- [x] `bun run check`
- [x] `bun run test`

### DIFF
--- a/packages/browseros-agent/apps/claw-server-rust/src/services/recordings.rs
+++ b/packages/browseros-agent/apps/claw-server-rust/src/services/recordings.rs
@@ -170,6 +170,7 @@ impl RecordingStore {
             self.release_append_handle(target_id).await;
             result?;
             if let Some(batch_id) = batch_id {
+                // Remember only committed batches so a failed append remains retryable.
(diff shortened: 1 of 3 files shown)


shadowfax92:
@claude review

shadowfax92:
@greptileai review

PR #1890, opened 2026-07-18
refactor(claw-server-rust): single shutdown owner and owned group locks

shadowfax92:
## Summary
- replace the consumable shutdown oneshot with an idempotent `ShutdownHandle` and route HTTP, signals, and stdio completion through one `AppRuntime` teardown
- retain and cancel browser, screencast, idle-sweeper, recording-retention, and HTTP boot-heal tasks, with bounded named joins
- move per-conversation tab-group operation locks from process-global state into each `PageOwnership` instance

## Design
`AppState` remains the cloneable service graph, while `AppRuntime` owns everything that is running. The shutdown route keeps its response-producing session drain so `drainedSessions` remains wire-compatible, then requests the shared cancellation token; after Axum finishes that response, the runtime performs audit/recording cleanup, stops browser services, and joins its tasks. The weak operation-lock registry now sits beside the durable group state it protects, so in-process app/test instances are isolated.

## Test plan
- [x] `cargo build -p claw-server-rust`
- [x] Ctrl-C smoke exits 0 with the drained-session log
- [x] `POST /system/shutdown` returns HTTP 200 with the unchanged 35-byte body, then exits 0

## Known remaining
- Tab-group effect work remains deliberately fire-and-forget (`drop(spawn_tab_group_work(...))`); task ownership for that short-lived workflow is a separate pass.


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1891, opened 2026-07-18
fix(claw-app): keep rrweb replay mounted in Strict Mode

shadowfax92:
## Summary
- make every `PlayerCanvas` Effect setup own a matching rrweb `Replayer` cleanup under React Strict Mode
- add a real `ReplayViewport` regression test for Strict Mode setup, target replacement, and unmount
- preserve the FullSnapshot gate, incomplete-recording UX, and both server replay contracts

## Design
Remove the redundant `lastEventsRef` equality guard and let the Effect dependency list control real event-array replacements. The regression test mocks only rrweb's imperative `Replayer`, renders the public viewport under `StrictMode`, and tracks live instances through setup, cleanup, replacement, and unmount.

## Test plan
- [x] focused Strict Mode test failed before the source fix and passes after it
- [x] neighboring replay tests
- [x] claw-app tests, typecheck, and lint
- [x] TypeScript recording and audit replay route tests
- [x] targeted and full claw-server-rust suites
- [x] repository `bun run check` and `bun run test`
- Optional GUI smoke not run: the only live dev runtime belonged to a separate checkout, so it could not exercise this branch

### DIFF
--- a/packages/browseros-agent/apps/claw-app/screens/replay/ReplayViewport.test.tsx
+++ b/packages/browseros-agent/apps/claw-app/screens/replay/ReplayViewport.test.tsx
@@ -1,0 +1,172 @@
+import { afterEach, beforeEach, describe, expect, it, mock } from 'bun:test'
+import { parseHTML } from 'linkedom'
+import { act, StrictMode } from 'react'
+import type { Root } from 'react-dom/client'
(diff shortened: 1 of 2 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1892, opened 2026-07-18
feat: add canonical BrowserClaw API contract

shadowfax92:
## Summary

- Add a split, implementation-neutral OpenAPI contract with route and schema files plus deterministic generated TypeScript and Rust DTO/client packages.
- Add handwritten canonical adapters to both Claw servers while preserving legacy routes, and verify shared behavior with the same 30-case contract suite.
- Migrate claw-app and the recorder to the neutral client, including exact tab/session recording attribution, stable retry IDs, reassociation gap recovery, and removal of the app dependency on the TypeScript server.

## Design

OpenAPI is the single wire-contract source of truth. The root document stays a small index over per-route and per-domain schema files. Generation produces only client/DTO artifacts; each server keeps a handwritten adapter over its production services. Canonical resources use sessions, tabs, and binary artifact endpoints while compatibility routes remain available for older installations.

## Test plan

- [x] Generate twice with no diff and run codegen:claw-api:check
- [x] Typecheck claw-api, claw-server, and claw-app
- [x] Run full claw-server and claw-app test suites
- [x] Build claw-app
- [x] Run cargo test for claw-api and claw-server-rust
- [x] Run clippy with warnings denied and cargo fmt check
- [x] Run the 30-case cross-server contract suite (108 assertions)
- [x] Verify canonical output has no legacy public identities, embedded JPEG payloads, generated server scaffold, or claw-app server dependency


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1896, opened 2026-07-18
chore: restore comments on claw servers and API contract surfaces

shadowfax92:
## Summary
- Restores the commentary that #1892 (canonical BrowserClaw API contract) dropped in the server rewrites, and documents the new hand-written surfaces that shipped bare — comments-only, zero behavior change.
- claw-server (TS): rewritten `server.ts` header for today's two-generation composition, canonical error-envelope/onError rationale, `SessionState` + dependency-seam null semantics, recording batch-pinning gotcha, `profileId` asymmetry, cache-control decisions, `BodyInit` copy gotcha.
- claw-server-rust: `routes::api_v1` module doc, mirrored route semantics (profile filter post-pagination, association pinning, label fallback chain), `RequestId`/`CanonicalError` error-dialect docs, `#[serde(skip)]` legacy-wire rationale, shutdown-owner note, headers for `examples/contract-server.rs` and the canonical/fixture test crates.
- contracts + codegen: purpose docs for the cross-server suite (`cases.ts` / `server-adapters.ts` / `cross-server.test.ts`, incl. the TS↔Rust fixture-lockstep constraint and the no-`:null` wire rule) and for `scripts/codegen/claw-api.ts` (spec location, emitted trees, regen + `--check` drift gate).

## Design
Single curated pass, one subsystem per commit, each verified with its own toolchain. Removed comments were treated as raw material, not restored verbatim — several described deleted behavior (e.g. the old `AppType` hono-rpc story in `server.ts`). Every factual claim was traced to the implementation before being written down; an adversarial context-free review then re-verified all claims and its four accuracy findings were fixed. Generated trees (`crates/claw-api/src/generated`, `packages/claw-api/src/generated`) are untouched — if richer generated-client docs are wanted, the right lever is OpenAPI `description` fields flowing through codegen (deliberately not done here).

## Test plan
- [x] claw-server: `bun run typecheck` + `bun run test` (61 files, 0 fail)
- [x] claw-server-rust: `cargo clippy --all-targets -- -D warnings` + `cargo test` (156 tests pass)
- [x] contract suite: `bun run test:claw-api-contract` (30/30) and `bun run codegen:claw-api:check` (output current)
- [x] `git diff main...HEAD` audited comments-only (230 insertions, 3 deletions, all comment/blank lines)

### DIFF
--- a/packages/browseros-agent/apps/claw-server-rust/examples/contract-server.rs
+++ b/packages/browseros-agent/apps/claw-server-rust/examples/contract-server.rs
@@ -1,3 +1,10 @@
+//! Contract-testable claw-server-rust: the full router over real app
+//! state seeded with one live and one ended session, no browser or CDP
+//! required. The cross-server contract suite
+//! (`contracts/claw-api/tests`) builds and spawns this binary to run
+//! the shared cases against the Rust implementation; it can also be run
+//! by hand: `cargo run --example contract-server <port> <data-dir>`.
+
 use axum::Router;
 use browseros_core::TargetId;
 use claw_server_rust::{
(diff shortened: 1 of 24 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1898, opened 2026-07-18
fix(browseros-core): deserialize backendDOMNodeId so refs mint

shadowfax92:
## Summary

- map CDP's exact `backendDOMNodeId` wire key into `AxNode.backend_dom_node_id`, restoring refs for semantic interactive elements
- map `Runtime.evaluate`'s `objectId` into the cursor-augmentation response, restoring refs and `[cursor=pointer]` for non-ARIA click targets
- parse a realistic `Accessibility.getFullAXTree` JSON fixture through the public render pipeline and assert button/textbox refs survive interactive mode

## Verified root cause

The original diagnosis was correct for semantic refs: `#[serde(rename_all = "camelCase")]` mapped `backend_dom_node_id` to `backendDomNodeId`, but CDP sends `backendDOMNodeId`. Before changing production code, the new deserialization test failed with `left: None`, `right: Some(101)`.

The live Rust-server smoke found one additional wire bug in the same path. Cursor scanning successfully located the synthetic pointer target, but `RemoteValue.object_id` had no `objectId` mapping, so `DOM.describeNode` was never reached. Its focused CDP-shaped test also failed `None` versus the expected object handle before the explicit rename.

No ref semantics, role tables, interactive filtering, or output formatting changed.

## Snapshot before / after

Before, the fixture's full snapshot had controls but no handles, and interactive mode reduced to its first ancestor line:

```text
- main "Checkout"
  - paragraph "Review your order"
  - button "Place order"
  - textbox "Promo code": "SAVE20"

# mode="interactive"
- main "Checkout"
```

After, the fixture and a live `claw-server-rust` smoke retain real controls; the live page also exercises cursor augmentation:

```text
- heading "AX smoke" [level=1]
- button "Run smoke" [ref=e1]
  - textbox "Query" [ref=e2]
- generic [ref=e3] [cursor=pointer]
```

## Audit

Manual CDP response structs in `browseros-core` were checked against the pinned protocol JSON. `backendDOMNodeId` is the only represented name with a preserved `DOM`/`URL`/`CSS`/`HTML` initialism. The live smoke additionally caught the missing ordinary camelCase `objectId` mapping described above; both corrections have direct deserialization assertions.

## Test plan

- [x] pre-fix `cdp_ax_tree_deserializes_backend_dom_node_ids` fails `None` vs `Some(101)`
- [x] pre-fix `runtime_remote_value_deserializes_object_id` fails `None` vs the CDP object handle
- [x] `cargo test -p browseros-core -p browseros-mcp -p claw-server-rust`
- [x] `cargo clippy -p browseros-core -p browseros-mcp -p claw-server-rust --all-targets -- -D warnings`
- [x] `bun run check` (exit 0; existing unrelated Biome/Fallow warnings remain)
- [x] `bun run test`
- [x] live `claw-server-rust` stdio MCP smoke against BrowserOS CDP 9421; background test page closed afterward

### DIFF
--- a/packages/browseros-agent/crates/browseros-core/src/observer.rs
+++ b/packages/browseros-agent/crates/browseros-core/src/observer.rs
@@ -379,6 +379,7 @@ struct RuntimeEvalResult {
 #[derive(Debug, Deserialize)]
 struct RemoteValue {
     value: Option<Value>,
+    #[serde(rename = "objectId")]
     object_id: Option<String>,
 }
 
@@ -673,7 +674,7 @@ fn name_of(node: &AxNode) -> String {
 
 #[cfg(test)]
 mod tests {
-    use super::resolve_ref_entry;
+    use super::{RemoteValue, resolve_ref_entry};
     use crate::{
         BrowserSession, BrowserSessionHooks, CoreError, ProtocolSession,
         connection::CdpConnection,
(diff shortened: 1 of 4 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1899, opened 2026-07-19
test(claw-mcp): cross-server real-browser MCP contract suite

shadowfax92:
## Summary

A Tier-3 integration suite for the BrowserClaw stack: deterministic local fixture pages + a **spawned real BrowserOS** + the full MCP tool matrix, driven against **both** claw servers (`apps/claw-server` TS and `apps/claw-server-rust`) over their `/mcp` endpoints and compared **semantically**.

It exists because no lower tier could see the `backendDOMNodeId` serde break that silently zeroed Rust ref-minting (#1898): unit tests build structs directly, the mock-CDP tests speak the author's casing, and the REST contract suite (`contracts/claw-api/`) sits above this layer. This one puts a live browser in the loop.

Lives at `packages/browseros-agent/contracts/claw-mcp/`, following the `contracts/claw-api/` mold. Gated on `BROWSEROS_BINARY` — unset, it skips cleanly and `bun run test` / `bun run check` stay green anywhere.

**Verified live** against a dev BrowserClaw: **205/205 cross-server tests pass**, smoke tier ~30s.

## Case counts per tool

| Tool group | Cases | Highlights |
|---|---:|---|
| transport / session [7] | 4 | tools/list (17 tools) · Origin→403 · unknown session id → reject · `DELETE /mcp` ends session + audit |
| tabs · tab_groups · windows | 17 | list/active/new/close · group create→verify/update/ungroup/close · window lifecycle |
| navigate · snapshot · diff | 16 | fresh-snapshot refs · **interactive drops paragraphs (REGRESSION)** · depth truncation · state markers · >15k-token spill · url-change diff |
| act (all kinds) | 20 | click/type/fill/press/hover/focus/check/select/scroll/drag/dialog · covered-click · unknown+stale ref texts |
| grep · read · evaluate · run | 20 | **grep over=ax keeps `[ref=eN]` (REGRESSION #2)** · console format · spill files · SDK end-to-end · exception→ok:false |
| screenshot · pdf · wait · upload/download | 16 | JPEG/PNG/WebP magic bytes · size/fullPage dims · `%PDF` · condition waits · file round-trips |
| name_session + cross-cutting [1-6] | 9 | two-session ownership · nonce fence on hostile page · auto-context · audit tie-in · cancellation · browser-down |
| **Total** | **102** | × 2 servers = 204 + parity gate |

Semantic comparison (not byte-wise): same `(role, name)` interactive sets carrying `[ref=`, same success/error classes, same spill/truncation behavior, same guard texts. A code-checked divergence registry (`tests/divergences.ts` + `DIVERGENCES.md`) lists expected cross-server deltas; the suite fails only on a **new** divergence.

## ⚠️ Cross-server disagreements found (migration gold)

The suite is the first thing to drive both servers against a real browser, and it surfaced **real behavioral gaps** — these are the migration to-do list:

**Bugs one server has and the other doesn't:**
- **`act kind=check`/`uncheck` — broken on Rust** (`CDP error: Invalid parameters`); works on TS.
- **`act kind=scroll` (page wheel scroll) — broken on Rust** (`Input.dispatchMouseEvent` times out ~60s); works on TS.
- **`windows set_visibility` — broken on Rust** (`Invalid parameters`); on TS it "works" but **recreates the window** and returns a new id, killing the old one.

**Bugs BOTH servers share (parity holds, but both should be fixed):**
- **`act kind=focus` by ref fails on both** (`CDP error: Document needs to be requested first`) — `focusElement` pushes backend node ids without a prior `DOM.getDocument`, which a snapshot never primes. A real click is the working focus path.
- A **checked checkbox never renders `[checked]`** from a live CDP AX tree on either server.
- **`evaluate` does not enforce its `timeout`** on a page-context wait.
- **A browser that dies mid-session** surfaces `CDP not connected` on every tool, **not** the polished `start BrowserClaw` guard (which only fires at boot).

**Intended (Rust-only) differences, plus surface differences:**
- Rust-only params: `snapshot mode`/`depth`, `act dialog_accept`/`dialog_dismiss`, `read format=console`.
- Rust `tabs new` embeds a `[Page N snapshot]`; TS returns only `opened page N`.
- Rust `act` appends a `[page N console]` summary on logged errors; TS embeds the diff only.
- Ownership-guard wording (TS names the owner; Rust does not), covered-click blocker naming, and `DELETE /mcp` content-type hygiene differ.

Full table with ids in [`contracts/claw-mcp/DIVERGENCES.md`](../blob/feat/claw-mcp-contract-suite/packages/browseros-agent/contracts/claw-mcp/DIVERGENCES.md).

## Also in this PR

- **Capture mode** (`CLAW_MCP_CAPTURE_DIR`) dumps raw CDP payloads and refreshes offline browseros-core serde fixtures; a new `captured_cdp_fixture.rs` guards their structure (backend node ids present, interactive pages mint refs) so the ref-minting regression can't pass unit tests again.
- Root scripts `test:claw-mcp-smoke` / `test:claw-mcp-contract`; the ungated unit tests + a gated smoke step join the `all` suite so it can't bit-rot.
- Nightly workflow step (after the signed build, mounting the just-built DMG) runs the full matrix.

## Test plan

- [x] `BROWSEROS_BINARY=… bun run test:claw-mcp-contract` — 205/205 pass, both servers
- [x] `BROWSEROS_BINARY=… bun run test:claw-mcp-smoke` — ~30s (<60s budget)
- [x] `bun run check` (lint/typecheck/fallow) green; `biome ci .` green
- [x] Unset `BROWSEROS_BINARY` → suite skips cleanly, tree green
- [x] `cargo test -p browseros-core --test captured_cdp_fixture` green
- [x] Nightly workflow YAML validates

### DIFF
--- a/.github/workflows/nightly-browserclaw.yml
+++ b/.github/workflows/nightly-browserclaw.yml
@@ -252,6 +252,48 @@ jobs:
             --notes-file "$notes_file" \
             "${files[@]}"
 
+      # Runs after the artifact is uploaded and the prerelease is published,
+      # so a red suite pings Slack (below) without blocking the nightly
+      # build. Mounts the just-built DMG and drives both claw servers'
+      # /mcp against that real browser.
+      - name: Run claw-mcp cross-server contract suite
+        env:
+          BROWSEROS_REPO: ${{ steps.build_inputs.outputs.browseros_repo }}
+          VERSION: ${{ steps.version.outputs.version }}
+        run: |
+          set -euo pipefail
+          shopt -s nullglob
+
+          if ! command -v cargo >/dev/null 2>&1; then
+            echo "::error::cargo is required to build claw-server-rust for the claw-mcp contract suite; install the Rust toolchain on this runner"
+            exit 1
+          fi
+
+          agent_root="$BROWSEROS_REPO/packages/browseros-agent"
+          dmgs=("$BROWSEROS_REPO"/packages/browseros/releases/"$VERSION"/BrowserClaw_*.dmg)
+          if [ "${#dmgs[@]}" -eq 0 ]; then
+            echo "::error::No BrowserClaw DMG found to test"
+            exit 1
+          fi
+
+          mount_point="$(mktemp -d)"
+          cleanup() { hdiutil detach "$mount_point" -quiet || true; }
+          trap cleanup EXIT
+          hdiutil attach "${dmgs[0]}" -nobrowse -readonly -mountpoint "$mount_point"
+
+          app="$(/usr/bin/find "$mount_point" -maxdepth 1 -name '*.app' -print -quit)"
+          exe="$(/usr/bin/find "$app/Contents/MacOS" -maxdepth 1 -type f -perm +111 -print -quit)"
+          if [ -z "$exe" ]; then
+            echo "::error::Could not resolve the BrowserClaw executable inside the DMG"
+            exit 1
+          fi
+          echo "Testing against $exe"
+
+          cd "$agent_root"
+          BROWSEROS_BINARY="$exe" \
+          BROWSEROS_TEST_HEADLESS=true \
+            bun contracts/claw-mcp/tests/run.ts
+
       - name: Slack failure ping
         if: ${{ failure() }}
         env:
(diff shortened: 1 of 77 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1905, opened 2026-07-20
fix(server): forward configured baseUrl for Anthropic, OpenAI, Google, OpenRouter, Azure factories

### LINKED ISSUE #1870: Anthropic provider ignores configured Base URL in agent server
### Issue type
Bug / regression

### Description
The AI settings UI allows configuring a custom `Base URL` for the Anthropic provider, but the agent server ignores the saved value.

`createAnthropicFactory()` currently constructs the SDK client with only the API key:

```ts
return createAnthropic({ apiKey: config.apiKey })
```

As a result, Anthropic-compatible providers such as MiniMax are tested against Anthropic's default endpoint instead of the configured endpoint.

This appears to be a regression or incomplete migration of #149 / `browseros-ai/BrowserOS-agent#197`: the form now accepts a custom URL, but the current monorepo provider factory does not pass it to `createAnthropic()`.

### Steps to reproduce
1. Add an **Anthropic** provider.
2. Set Base URL to `https://api.minimax.io/anthropic`.
3. Set model to `MiniMax-M3` and enter a valid MiniMax Coding Plan key.
4. Save and click **Test**.

### Actual behavior
The test fails with:

```text
[anthropic] No output generated. Check the stream for errors.
```

The configured `baseUrl` is not supplied to the Anthropic SDK.

### Expected behavior
When a Base URL is configured, BrowserOS should initialize the provider as:

```ts
createAnthropic({
  apiKey: config.apiKey,
  ...(config.baseUrl && { baseURL: config.baseUrl }),
})
```

The default Anthropic endpoint should remain unchanged when no custom Base URL is supplied.

### Affected code
`packages/browseros-agent/apps/server/src/agent/provider-factory.ts`

### Proposed regression coverage
Add a provider-factory test asserting that a custom Anthropic `baseUrl` is forwarded as the SDK's `baseURL` option, while an absent value preserves the SDK default.

DaniAkash:
Fixes #1870.

## Summary

The AI settings dialog accepts a Base URL for the Anthropic, OpenAI, Google, OpenRouter, and Azure providers and round-trips the value through the persisted `LlmProviderConfig`, but `provider-factory.ts` never forwarded it to the underlying SDK. Custom endpoints (MiniMax's Anthropic-compatible API, LiteLLM, self-hosted OpenAI/Google/OpenRouter gateways, custom Azure OpenAI gateways) therefore silently resolved to each SDK's default endpoint and failed with a confusing "no output generated" error.

This PR threads `baseURL: config.baseUrl` through each of the five factories when the field is set, and keeps the SDK default when it is unset.

## Verified against each SDK's docs

| Package | Option | Source |
|---|---|---|
| `@ai-sdk/anthropic` | `baseURL: string` (default `https://api.anthropic.com/v1`) | https://ai-sdk.dev/providers/ai-sdk-providers/anthropic |
| `@ai-sdk/openai` | `baseURL: string` (default `https://api.openai.com/v1`) | https://ai-sdk.dev/providers/ai-sdk-providers/openai |
| `@ai-sdk/google` | `baseURL: string` (default `https://generativelanguage.googleapis.com/v1beta`) | https://ai-sdk.dev/providers/ai-sdk-providers/google |
| `@ai-sdk/azure` | `baseURL: string`; mutually exclusive with `resourceName` (SDK ignores resourceName when baseURL is set). Matches the UI copy "Overrides resource name if set". | https://ai-sdk.dev/providers/ai-sdk-providers/azure |
| `@openrouter/ai-sdk-provider@2.9.1` | `baseURL?: string` (legacy `baseUrl?` alias marked `@deprecated`) | shipped `.d.ts` in this project's `node_modules` |

## OpenAI Responses vs Chat Completions

`@ai-sdk/openai` defaults to the Responses API. Many OpenAI-shape proxies (MiniMax, LiteLLM, self-hosted gateways) only speak Chat Completions. The factory now forwards the user's `baseUrl` regardless; the OpenAI branch of the provider dialog gains a hint next to the Base URL field:

> If your custom endpoint doesn't work with OpenAI, try the OpenAI Compatible provider template instead.

That puts the choice in the user's hands: point at a Responses-compatible endpoint and it works direct; hit a Chat-Completions-only proxy and the visible next step is the existing "OpenAI Compatible" template.

## Files changed

- `packages/browseros-agent/apps/server/src/agent/provider-factory.ts` — five factories updated; Azure now accepts either `resourceName` or `baseUrl`.
- `packages/browseros-agent/apps/app/screens/ai-settings/NewProviderDialog.tsx` — inline `FormDescription` on the OpenAI Base URL field.
- `packages/browseros-agent/apps/server/tests/agent/provider-factory-baseurl.test.ts` — new suite. 13 assertions across the five providers.

## Behavior preservation

- Users who never set Base URL see zero change: each factory omits the SDK option entirely when `config.baseUrl` is falsy.
- Empty-string Base URL (form serialization for cleared fields) treated same as unset (Anthropic test explicitly covers this).
- Azure's `resourceName`-only path is unchanged.
- OpenAI users see one new line of description text on the Base URL field.

## Test plan

- [x] `bun test tests/agent/provider-factory-baseurl.test.ts` in `apps/server` → 13/13 pass.
- [x] `bun test tests/agent` in `apps/server` → 314/314 pass.
- [x] `bun test screens/ai-settings/NewProviderDialog.test` in `apps/app` → 7/7 pass.
- [x] `bunx tsc --noEmit` clean in both `apps/server` and `apps/app` (app requires `NODE_OPTIONS='--max-old-space-size=8192'` due to a pre-existing repo-scale memory ceiling; unrelated to this PR).
- [x] `bunx @biomejs/biome check` clean.
- [ ] Manual smoke: configure an Anthropic provider with `https://api.minimax.io/anthropic` + a MiniMax key, click Test, confirm response comes through the custom endpoint.

### DIFF
--- a/packages/browseros-agent/apps/app/screens/ai-settings/NewProviderDialog.tsx
+++ b/packages/browseros-agent/apps/app/screens/ai-settings/NewProviderDialog.tsx
@@ -889,6 +889,12 @@ export const NewProviderDialog: FC<NewProviderDialogProps> = ({
               <FormControl>
                 <Input placeholder="https://api.openai.com/v1" {...field} />
               </FormControl>
+              {watchedType === 'openai' && (
+                <FormDescription>
+                  If your custom endpoint doesn't work with OpenAI, try the
+                  OpenAI Compatible provider template instead.
+                </FormDescription>
+              )}
               <FormMessage />
             </FormItem>
           )}
--- a/packages/browseros-agent/apps/server/src/agent/provider-factory.ts
+++ b/packages/browseros-agent/apps/server/src/agent/provider-factory.ts
@@ -278,18 +278,21 @@ function createAnthropicFactory(
   })
 }
 
-// OpenAI-specific: `baseUrl` is intentionally NOT forwarded here.
-// The SDK's default transport is the Responses API and almost every
-// OpenAI-shape proxy in the wild (MiniMax, LiteLLM, self-hosted
-// gateways) speaks Chat Completions instead, so silently threading
-// baseURL would swap one silent failure for another. The /mcp UI
-// shows a hint on the OpenAI Base URL field pointing users at the
-// "OpenAI Compatible" provider template for that case.
+// The SDK defaults to the Responses API; many OpenAI-shape proxies
+// only speak Chat Completions. The `NewProviderDialog` shows a hint
+// on the OpenAI Base URL field pointing users at the "OpenAI
+// Compatible" provider template for that case, so a proxy that fails
+// here has a documented next step. Users who point at a Responses-
+// compatible custom endpoint get the direct-forward behavior they
+// asked for.
 function createOpenAIFactory(
   config: ResolvedAgentConfig,
 ): (modelId: string) => unknown {
   if (!config.apiKey) throw new Error('OpenAI provider requires apiKey')
-  return createOpenAI({ apiKey: config.apiKey })
+  return createOpenAI({
+    apiKey: config.apiKey,
+    ...(config.baseUrl && { baseURL: config.baseUrl }),
+  })
 }
 
 function createGoogleFactory(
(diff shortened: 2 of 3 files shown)


DaniAkash:
@greptileai

PR #1908, opened 2026-07-20
fix(claw): make retained group collapse idempotent

shadowfax92:
## Summary
- confirm Chromium tab-group absence after a failed retained collapse and clear the matching cached identity
- preserve present or unknown group state for warning and later retry
- protect replacement groups with atomic compare-and-clear ownership semantics

## Design
Retained collapse keeps its optimistic update under the per-conversation operation lock. Only a failed update triggers the existing group-list probe; a valid response confirming absence clears the attempted group identity atomically and succeeds, while present or unknown results retain state and fail. Later retained sweeps see no cached group and perform no CDP work.

## Test plan
- [x] cargo fmt -p claw-server-rust --check
- [x] cargo clippy -p claw-server-rust --all-targets -- -D warnings
- [x] cargo test -p claw-server-rust mcp::effects::tab_groups::tests
- [x] cargo test -p claw-server-rust sessions::tests
- [x] cargo test --workspace
- [x] bun run check
- [x] bun run test

### DIFF
--- a/packages/browseros-agent/apps/claw-server-rust/src/mcp/effects/tab_groups.rs
+++ b/packages/browseros-agent/apps/claw-server-rust/src/mcp/effects/tab_groups.rs
@@ -221,7 +221,7 @@ async fn ensure_agent_tab_group_unlocked(
     }
 }
 
-/// Collapses the durable group when its session enters retention.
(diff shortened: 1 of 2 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1909, opened 2026-07-20
fix(browseros-core): treat select as semantic input

shadowfax92:
## Summary
- remove physical click hit-testing from Rust semantic select while preserving ref resolution and scrolling
- keep click and hover blocker protection intact, including the same covered combobox regression
- add real-browser cross-server coverage for styled native select success, bubbling change, and missing-option behavior

## Design
Rust select now follows the same semantic DOM path as TypeScript: resolve the target, scroll it into view, and execute the existing option-selection function. Pointer actions retain their existing center hit-test and mouse-dispatch path; geometry classification is unchanged.

## Test plan
- [x] `cargo clippy -p browseros-core -p browseros-mcp --all-targets -- -D warnings`
- [x] `cargo test -p browseros-core -p browseros-mcp`
- [x] `BROWSEROS_BINARY=/Applications/BrowserClaw.app/Contents/MacOS/BrowserClaw bun run test:claw-mcp-smoke`
- [x] `bun run check`
- [x] `bun run test`

### DIFF
--- a/packages/browseros-agent/contracts/claw-mcp/fixtures/pages/form.html
+++ b/packages/browseros-agent/contracts/claw-mcp/fixtures/pages/form.html
@@ -3,6 +3,26 @@
   <head>
     <meta charset="utf-8" />
     <title>Form fixture</title>
+    <style>
+      .styled-select {
+        position: relative;
+        display: inline-block;
+      }
+      .styled-select select {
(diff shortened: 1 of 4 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1910, opened 2026-07-20
fix: make recording delivery bounded and recoverable

shadowfax92:
## Summary

- batch rrweb NDJSON by UTF-8 bytes while preserving atomic lines, per-tab ordering, and retry-stable batch ids
- publish an optional recording ingest capability with a conservative old-server fallback, bounded relay queues, terminal 413 handling, gap recovery, and resnapshot throttling
- enforce the same explicit 4 MiB ceiling in TypeScript and Rust with canonical 413 responses and cross-server parity coverage

## Design

The recorder targets the conservative 2 MiB fallback so upgraded extensions remain compatible with older sidecars. Individually large snapshots stay atomic and are checked against the optional server capability. Both servers advertise and enforce a 4 MiB canonical ceiling. Oversized batches are removed from the queue, marked as replay gaps, and followed by bounded recovery instead of becoming permanent retry heads.

The TypeScript boundary streams request bytes through a bounded middleware and drains rejected uploads without retaining them, avoiding Bun client backpressure after 413. Rust configures Axum's extractor limit explicitly and maps extractor rejections into the canonical error envelope.

## Verification

- `bun run codegen:claw-api:check`
- `bun run --filter @browseros/claw-app test`
- `bun run --filter @browseros/claw-server test`
- `bun run test:claw-api-contract`
- `cargo fmt --all --check`
- `cargo clippy --workspace --all-targets -- -D warnings`
- `cargo test --workspace`
- `bun run check`
- `bun run test`


shadowfax92:
@claude review

shadowfax92:
@greptileai review

PR #1911, opened 2026-07-20
fix(claw-server-rust): enforce live tab incarnations

shadowfax92:
## Summary

- reconcile every Rust tab activity row against the live page and exact target incarnation before canonical or recording consumers can use it
- pin MCP activity, screencast frames, capture work, backoff, and audit fallback to `(pageId, targetId)` and clean stale lifecycle state
- make PageManager refresh publication race-safe and keep transient disconnects non-destructive and bounded

## Design

`TabActivityService` now snapshots stored trails without holding its mutex over CDP work, validates live page plus target identity, refreshes URL/title, and prunes only the exact record version it checked. Successful close effects remove the exact incarnation promptly while central reconciliation covers external closes, crashes, disconnects, and page reuse.

Screencast state and capture are target-bound. BrowserSession checks the expected target before and after capture, stale failure/frame state is garbage-collected, and canonical preview lookup cannot cross a reused page handle.

## Test plan

- `bun run fmt:rust`
- `bun run lint:rust`
- `bun run test:rust`
- `bun run test:claw-api-contract` (34 pass)
- `bun run check`
- `bun run test`
- `git diff --check`


shadowfax92:
@claude review

shadowfax92:
@greptileai review

PR #1917, opened 2026-07-20
fix(app): make custom model IDs discoverable in the provider picker

shadowfax92:
## Problem

> when you select LM Studio it doesnt show up with the model list just 3 pre defined ones

The Model combobox renders a bundled models.dev snapshot as if it were the provider's live catalog. LM Studio has exactly three entries there, so anyone running their own loaded models reads that list as "unsupported" and gives up.

Free-form entry already worked — but nothing said so. The trigger read "Select a model...", the search read "Search models...", and the custom row rendered as a bare ID indistinguishable from a catalog entry. The one hint that existed (`CommandEmpty`'s "Press Enter to use …") was unreachable dead code: the custom row always mounts, so cmdk's item count is never zero.

Expanding the preset list would not fix this — no static catalog can know which models a local endpoint has loaded.

## Changes

- **Placeholders** say what is actually possible: `Select or paste a model ID` on the trigger, `Search or paste a model ID...` in the search box.
- **Persistent popover footer** spells out the gesture: *Model not listed? Type or paste its exact ID, then press Enter.*
- **Free-form row** reads `Use "<id>"` with a `CUSTOM` badge, pinned above the catalog.
- **In-place hint** under the field for LM Studio / Ollama / OpenAI-compatible only — the catalogs that are knowingly incomplete. Providers with an authoritative catalog stay uncluttered.
- Picker row derivation moved to `model-picker.helpers.ts` (`getModelPickerRows`) with unit tests driven by the real Fuse config over the shipped catalogs.

## Bugs fixed along the way

- **Enter ignored the highlighted item.** The `onKeyDown` interception called `stopPropagation()`, so cmdk never got to select. Typing `qwen`, arrowing down to `qwen/qwen3-coder-30b`, and pressing Enter saved the literal string `qwen`. Removing the interception lets cmdk select normally — and because cmdk re-selects the first item on every keystroke, the free-form row still wins Enter while typing, even when a fuzzy near-match is listed. It also restores cmdk's IME-composition guard, which the hand-rolled handler lacked.
- **Pasted IDs were stored untrimmed**, so `  openai/gpt-oss-120b  ` reached the provider verbatim and came back rejected.
- **A pasted ID with a stray space selected the wrong model.** Trimming what got committed while leaving the Fuse query on the raw text made the two halves of the picker disagree: `gpt-5 ` scored better against `gpt-5.5` than against the exact `gpt-5`, so Enter silently saved a different model and logged it as a deliberate pick. The same root cause emptied the list for `  o3  ` — a direct catalog hit once trimmed — leaving Enter with nothing to commit and no way out but deleting the whitespace by hand. Both derivations now run on the trimmed ID.
- **The context window carried over onto custom models.** The autofill effect only wrote when the catalog could size the model, so selecting `gpt-5.5` and then pasting an 8k local model saved a 1M-token window. Custom models now take the 128k default. This one predates the branch, but free-form entry was a hidden affordance before and is the promoted path now, so it is newly reachable by design.

The whitespace fix is also what makes dropping the empty state sound: an exact catalog ID scores 0 and ranks itself first, and anything else is by definition custom, so there is always at least one row.

## Verification

`bun run check` clean on the touched files; `bun run test` → 2222 pass, 0 fail. Two adversarial review passes; every finding above the cosmetic line is fixed in-branch.

Driven against the running extension over CDP (`scripts/dev/inspect-ui.ts`):

| Flow | Result |
| --- | --- |
| LM Studio → type unlisted ID → Enter | commits the custom ID, picker closes |
| …with a fuzzy near-match listed (`qwen/qwen3-coder-30b` vs `…-30b-instruct-mlx`) | custom row stays selected, custom ID wins |
| Type → ArrowDown → Enter | commits the highlighted catalog model + autofills context window (was broken) |
| Mouse-click a catalog model | commits, context window → 262144 |
| Paste `"  openai/gpt-oss-120b  "` | stored trimmed |
| Paste `"gpt-5 "` | commits `gpt-5` @ 400K (was `gpt-5.5`) |
| Paste `"  o3  "` | commits `o3` (was an empty list, Enter a no-op) |
| `gpt-5.5` (1.05M) → paste custom ID | context window resets to 128K (was 1,050,000) |
| Anthropic (authoritative catalog) | no hint rendered, 18 models + footer |

### DIFF
--- a/packages/browseros-agent/apps/app/screens/ai-settings/NewProviderDialog.tsx
+++ b/packages/browseros-agent/apps/app/screens/ai-settings/NewProviderDialog.tsx
@@ -96,6 +96,9 @@ function showsStandardModelField(type: ProviderType): boolean {
   return !isAcpProviderType(type)
 }
 
+/** Window assumed for any model the bundled catalog cannot size. */
+const DEFAULT_CONTEXT_WINDOW = 128000
+
 function defaultReasoningEffort(type?: ProviderType) {
   return type === 'chatgpt-pro' ? 'medium' : 'high'
 }
@@ -211,7 +214,7 @@ export const NewProviderDialog: FC<NewProviderDialogProps> = ({
       modelId: initialValues?.modelId || '',
       apiKey: initialValues?.apiKey || '',
       supportsImages: initialValues?.supportsImages ?? false,
-      contextWindow: initialValues?.contextWindow || 128000,
+      contextWindow: initialValues?.contextWindow || DEFAULT_CONTEXT_WINDOW,
       temperature: initialValues?.temperature ?? 0.2,
       resourceName: initialValues?.resourceName || '',
       accessKeyId: initialValues?.accessKeyId || '',
@@ -325,16 +328,17 @@ export const NewProviderDialog: FC<NewProviderDialogProps> = ({
 
   useEffect(() => {
     if (initialValues?.id) return
+    if (!watchedModelId) return
 
-    if (watchedModelId) {
-      const contextLength = getModelContextLength(
-        watchedType as ProviderType,
-        watchedModelId,
-      )
-      if (contextLength) {
-        form.setValue('contextWindow', contextLength)
-      }
-    }
+    // A custom model has no catalog entry, so fall back to the default rather
+    // than keeping whatever the previously selected model left behind: picking
+    // gpt-5.5 and then pasting an 8k local model would otherwise save a
+    // 1M-token window and overflow it on the first long chat.
+    const contextLength = getModelContextLength(
+      watchedType as ProviderType,
+      watchedModelId,
+    )
+    form.setValue('contextWindow', contextLength || DEFAULT_CONTEXT_WINDOW)
   }, [watchedModelId, watchedType, form, initialValues?.id])
 
   useEffect(() => {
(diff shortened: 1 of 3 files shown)


shadowfax92:
Thanks — taking the `useCallback` suggestion as a deliberate no.

`apps/app/CLAUDE.md` is explicit: *"Avoid `useCallback` and `useMemo` unless they solve a measured or obvious render problem."* `commitModelId` is handed to at most a few dozen `CommandItem`s inside a popover that is open only while the user is picking, and cmdk re-renders the list on every keystroke regardless — memoizing the callback would not prevent a single re-render, so it would be cost without benefit.

On the flagged assumption — *"worth a quick read-through to confirm the cmdk selection-reset assumption holds across the versions in use"* — it is verified against the pinned cmdk 1.1.1 rather than assumed. `setState("search", …)` schedules `selectFirstItem` unconditionally (`dist/index.mjs`), and with `shouldFilter={false}` the sort is a no-op, so DOM order is preserved and the free-form row (rendered first) is what Enter commits. Confirmed live over CDP, including the adversarial case where a fuzzy near-match is still listed.

Note the summary above predates the last two commits. Two further bugs were found and fixed in-branch since: a pasted `gpt-5 ` committed `gpt-5.5` because the Fuse query still ran on the untrimmed text, and the context window carried over onto custom models. Both are covered in the PR description and by tests.

PR #1918, opened 2026-07-20
chore: disable model invocation for Claude skills

shadowfax92:
## Summary
- mark all three `.claude/skills/*/SKILL.md` files as unavailable for model invocation
- keep `disable-model-invocation: true` as the final YAML frontmatter key

## Design
This is a frontmatter-only change: each skill gets the same one-line key immediately before the closing delimiter, with no changes to skill content.

## Test plan
- [x] `rg -c '^disable-model-invocation: true$' .claude/skills/*/SKILL.md` returns 1 for every skill\n- [x] `git diff --check origin/main..HEAD`\n- [x] branch diff is exactly 3 insertions across 3 files


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1946, opened 2026-07-21
refactor(claw-server): organize HTTP and MCP APIs

shadowfax92:
## Summary

- centralize the TypeScript claw-server HTTP method/path table under `src/api/http`
- split canonical resource behavior into narrow handler factories with production bindings beside each resource
- move the MCP transport and dispatch pipeline mechanically under `src/api/mcp`
- add a layout regression test and remove the unused agent display route helper

## Design

`server.ts` keeps global middleware and mounts the HTTP and MCP registries. HTTP handlers preserve the generated claw API contract while depending only on their resource operations. Existing services, stores, lifecycle, and generated contracts remain unchanged.

## Test plan

- [x] `bun run --filter @browseros/claw-server test`
- [x] `bun run --filter @browseros/claw-server typecheck`
- [x] `bun run test:claw-api-contract` — 46 TypeScript/Rust parity cases
- [x] `bunx biome check --config-path biome.json --changed --since origin/main` — 63 changed files clean
- [x] `bun run test:claw-mcp-contract` — exits 0 and skips by design when `BROWSEROS_BINARY` is unset

Package-wide `bun run --filter @browseros/claw-server lint` remains nonzero on the unchanged `tests/lib/migrate-mcp-config-paths.test.ts` unsafe optional-chaining diagnostic and existing non-null warnings. This branch changes none of those files; all changed files are clean.


shadowfax92:
@claude review

shadowfax92:
@greptileai review

PR #1949, opened 2026-07-21
fix(claw): handle CORS preflight before route dispatch

shadowfax92:
## Summary
- answer trusted and origin-less OPTIONS requests before exact Axum method routing returns 405
- preserve recording-ingest origin rejection and normal non-OPTIONS method semantics
- enforce the trusted recording preflight contract against both TypeScript and Rust servers

## Design
The canonical Rust HTTP router now has a minimal OPTIONS-only middleware inside the existing request-context layer. Request context remains responsible for origin trust, CORS headers, request IDs, and failure logging; non-OPTIONS requests continue through the existing route tree unchanged.

## Test plan
- [x] cargo test -p claw-server-rust --test canonical_routes -- --test-threads=1
- [x] cargo fmt --all --check
- [x] cargo test -p claw-server-rust
- [x] cargo clippy -p claw-server-rust --all-targets -- -D warnings
- [x] bun run test:claw-api-contract
- [x] live trusted/hostile preflight, system GET, and unsupported-method curl checks

### DIFF
--- a/packages/browseros-agent/apps/claw-server-rust/src/api/http/mod.rs
+++ b/packages/browseros-agent/apps/claw-server-rust/src/api/http/mod.rs
@@ -80,6 +80,7 @@ pub fn router(state: AppState) -> Router<AppState> {
                 .layer(middleware::from_fn(mcp_request_hygiene)),
         )
         .fallback(route_fallback)
+        .layer(middleware::from_fn(options_preflight))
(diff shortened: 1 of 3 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1950, opened 2026-07-21
fix(claw): make recording API client receiver-safe

shadowfax92:
## Summary
- generate a deterministic, type-only OpenAPI operation surface and add `@browseros/claw-api-client` with all 17 canonical operations
- invoke configured fetch lexically, preserve exact NDJSON recording requests, keep error responses readable, and avoid cloning successful large bodies
- migrate claw-app hooks and the durable recorder relay to the package without changing outbox, retry, or replay parsing behavior

## Design
The canonical OpenAPI document remains the route authority. Codegen now commits a type-only `openapi-typescript` tree alongside the existing DTO trees, while the handwritten `openapi-fetch` facade owns transport semantics and stable DTO returns. claw-app retains only BrowserOS port/base-URL resolution and client caching; route strings and raw replay fetches move behind the package.

## Test plan
- [x] `bun test scripts/codegen/claw-api.test.ts contracts/claw-api/tests/schema.test.ts contracts/claw-api/tests/fixtures.test.ts`
- [x] `bun run --filter @browseros/claw-api-client test` and typecheck
- [x] 29 focused claw-app API, replay, audit, and recorder tests
- [x] `bun run codegen:claw-api:check`
- [x] claw-app typecheck and development build
- [x] `bun run test:claw-api-contract`
- [x] `cargo test -p claw-server-rust --test canonical_routes -- --test-threads=1` (23 passed)
- [x] disposable live BrowserClaw proof: `GET /api/v1/system` 200, `POST /api/v1/recordings/events` 200, outbox drained without deletion, recording has data (1,853 bytes, one tab, four events)
- [x] receiver mutation proof: reintroducing an object receiver makes the recorder regression test fail before either request is sent
- [x] full claw-app suite baseline classified: the same 16 unrelated `@/…` resolver failures reproduce on untouched main; all touched test files pass


shadowfax92:
@greptileai review

shadowfax92:
@claude review

shadowfax92:
Review triage: 5 discussion points · ACCEPT 0 · DISCUSS 0 · DECLINE 5 · unresolved threads 0.

1. The root barrel is intentional here: the approved seeded design explicitly specifies `src/index.ts` as the stable facade and requires ordinary claw-app consumers to import `@browseros/claw-api-client`. Splitting it would override that fixed design decision.
2. The 2xx parse-error scenario does not enter `unwrap` as an error result in openapi-fetch 0.17. Its success branch awaits parsing and rejects directly; only non-OK responses return `{ error, response }`. Cloning successful replay/blob bodies would reintroduce the buffering problem covered by the no-success-clone test.
3. The credential fallback is reached. openapi-fetch puts the client option on its synthesized `Request`, then calls the injected fetch with that Request and an undefined second argument on this runtime. Keeping both sites preserves Request semantics and explicitly supplies credentials at the custom-fetch boundary.
4. A client instance is allocation-only and holds no connection. Adding relay cache state would change the deliberately preserved durable-outbox/retry control flow for no measured benefit.
5. The explicit `{}` branch deliberately omits the optional query field without depending on third-party undefined serialization. That is clearer at a wire-contract boundary than passing `refresh: undefined`.

No code changes from these points. Greptile also reports 5/5 confidence with no blocking issues.

PR #1951, opened 2026-07-21
fix: cap automatic action settling at 50ms

shadowfax92:
## Summary
- cap automatic post-action settling at 50ms total before diff or snapshot capture
- document the intentional responsiveness tradeoff and direct slower asynchronous changes to the explicit `wait` tool
- pin the automatic default and total-budget behavior in the focused settle test

## Design
Keep the existing single-deadline settle pipeline unchanged and reduce only its automatic default budget. This preserves one total cap across navigation detection, load waiting, DOM quiet, and CDP calls without tuning unrelated timeouts.

## Test plan
- [x] `cargo fmt --package browseros-core --package browseros-mcp --check`
- [x] `cargo test --package browseros-core --package browseros-mcp`


shadowfax92:
@claude review

shadowfax92:
@greptileai review

PR #1958, opened 2026-07-22
docs(core): explain snapshot reference semantics

shadowfax92:
## Summary
- preserve the full snapshot as the next diff baseline even when returned output is filtered
- document synthesized document scope and capture-local stale-ref recovery metadata
- define the URL-change full-snapshot result shape

## Design
The comments sit on the producer boundaries that own each invariant, avoiding duplicate explanations in MCP consumers.

## Test plan
- [x] `cargo test -p browseros-core`
- [x] `git diff --check`
- [x] independent Codex review, one wording correction, final PASS


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1959, opened 2026-07-22
feat(claw-server): add release-safe version contract

shadowfax92:
## Summary
- add a terminal `--version` action that prints only the Cargo package version and never enters BrowserClaw startup
- set the checked-in Rust package to `0.0.13` and share that metadata across CLI, HTTP system info, and MCP initialization
- continue Rust releases on `claw-server/v` and verify every stamped matrix binary before packaging

## Design
Clap projects validated arguments into an explicit `Version` or `Run` action, keeping optional config confined to parsing while run mode always receives a concrete path. The release workflow retains manifest and lockfile stamping, then executes the target-specific binary and compares its output with the requested release version before staging.

## Test plan
- [x] `cargo test -p claw-server-rust --locked`
- [x] `cargo clippy -p claw-server-rust --locked --all-targets -- -D warnings`
- [x] focused Bun release helper and workflow tests
- [x] `bun run typecheck`
- [x] `actionlint` and `shellcheck` on the changed release files
- [x] context-free adversarial diff review: clean

### DIFF
--- a/.github/workflows/release-claw-server-rust.yml
+++ b/.github/workflows/release-claw-server-rust.yml
@@ -6,7 +6,7 @@ name: "Release: BrowserClaw Server (Rust)"
 on:
   push:
     tags:
-      - "claw-server-rust/v*"
+      - "claw-server/v*"
   workflow_dispatch:
     inputs:
       version:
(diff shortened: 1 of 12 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

shadowfax92:
Addressed Claude's test-fidelity note: the release-helper fixture now passes `--release-ref HEAD`, matching the workflow invocation. Focused release tests remain green.

PR #1948, opened 2026-07-21
chore(claw): retire the TypeScript server

shadowfax92:
## Summary

- archive the latest TypeScript Claw server as an inert top-level reference and make Rust the only active workspace, dev, dogfood, CI, and MCP implementation
- remove two-server parity while retaining canonical OpenAPI, codegen, schema, fixture, and Rust-only MCP conformance coverage
- split onboarding release ownership and migrate BrowserClaw bundles and OTA to the existing Rust artifact identities and compatibility binary name
- merge landed API and session-visibility work into Rust and refresh the archive from the matching final TypeScript source

## Verification

- `cargo fmt --all -- --check`
- `cargo clippy --workspace --all-targets --locked -- -D warnings`
- `cargo test --workspace --locked`
- `bun run codegen:claw-api:check`
- `bun run test:claw-api-contract`
- 102/102 live Rust MCP conformance cases against BrowserClaw
- release workflow tests, release-secret tests, actionlint, Biome, and frozen Bun install
- 724 Python build tests, Ruff, product doctor, and BrowserClaw release plan
- dev and dogfood Go test suites

## Review

Two context-free review rounds completed. Round one found and fixed missing OTA R2 credentials, the missing retained Claw API CI script, and onboarding reflection Git identity. Round two was clean.

PR #1977, opened 2026-07-22
fix(bos-build): scope universal skew guard to the product's server bundles

shadowfax92:
## Summary
- The merge_universal skew guard scanned **every** directory under `resources/binaries/`, so orphaned families — dirs left behind when a bundle is retired or renamed — failed the merge forever on persistent runners: no download/copy op manages them anymore, so nothing can ever re-align their per-arch metadata.
- BrowserClaw 0.48.0 universal (run [29882827339](https://github.com/browseros-ai/BrowserOS/actions/runs/29882827339), job 88809062699) hit exactly this: the guard tripped on `browseros_claw_server` — the TypeScript server family retired in #1948 — with arm64 `0.0.11` vs x64 `0.0.12`, while the actual shipping family (`browseros_claw_server_rust`) was fresh and aligned at `0.0.13` on both arches. Pure validation false positive; the merge never started.
- The guard now iterates `server_bundles_for_product(ctx.product.id)` and checks each bundle's `local_resources_root` — exactly the families that ship in the merging product's app.

## Design
Registry-scoped, not filesystem-scoped: the `ServerBundle` registry (`bos_build/products/server_binaries.py`) already drives download destinations, signing, and OTA, so it is the single source of truth for what enters an app. Product scoping is deliberate (`server_bundles_for_product`, not `all_server_bundles`): nightlies stage `version: "local"` arm64 bundles daily, and a browseros merge must not be blocked by claw-family skew (or vice versa) — only bundles whose `product_ids` include the merging product ship in that app. Skip-silently semantics for missing dirs/metadata, the error message shape, and genuine own-family skew detection (e.g. in `--from` resumes) are unchanged. Rejected: allowlists (second source of truth that rots the same way), parsing `download_resources.yaml` (wrong ownership boundary), pruning orphan dirs in the download step (risky vs nightly local-staging flows), special-casing `"local"` versions (that skew is genuine when the guard sees it).

## Test plan
- [x] Regression for run 29882827339: orphaned `browseros_claw_server` dir with mismatched versions is ignored for product browserclaw
- [x] Own-family skew still raises with family name, both versions, both generatedAt timestamps
- [x] Cross-product isolation: browseros ignores claw-family skew (incl. the nightly `"local"` case) but still raises on `browseros_server`
- [x] Matching versions pass; missing metadata on either arch side skips; missing binaries dir is a no-op
- [x] From `packages/browseros`: `uv run python -m unittest discover -s bos_build -t . -p "*_test.py"` — 726 tests OK · `uv run ruff check bos_build` — clean

### DIFF
--- a/packages/browseros/bos_build/steps/compile/universal.py
+++ b/packages/browseros/bos_build/steps/compile/universal.py
@@ -15,6 +15,7 @@ from typing import Optional
 
 from ...core.context import Context
 from ...core.step import Step, ValidationError, step
+from ...products.server_binaries import server_bundles_for_product
 from ..package.merge import merge_architectures
 from ..storage.download import ARTIFACT_METADATA_NAME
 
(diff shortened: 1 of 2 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1983, opened 2026-07-22
fix(claw): classify recovered session status

shadowfax92:
## Summary
- classify completed sessions as failed only when their final three non-cancelled tool calls all error
- preserve live and cancellation precedence, historical error counts, and the canonical API status consumed by Audit and Cockpit
- reclassify existing persisted session rows on upgrade so recovered historical work is corrected immediately

## Design
The Rust audit task projection now derives status from the ordered dispatch tail plus the first session end event. A one-time data migration applies the same rule to existing task rows, keeping server-side status filters and every UI consumer on one canonical value.

## Test plan
- [x] `cargo test -p claw-server-rust db::audit_log::tests --lib`
- [x] `cargo test -p claw-server-rust db --lib`
- [x] `bun run fmt:rust`
- [x] `bun run lint:rust`
- [x] `bun run test:rust`
- [x] `bun run codegen:claw-api:check`
- [x] `bun run test:claw-api-contract`
- [x] `bun run test`

## Baseline note
- `bun run check` currently stops on two pre-existing `noUnsafeOptionalChaining` errors in `apps/server/tests/lib/agents/acpx-runtime.test.ts`; this branch does not modify that file.

### DIFF
--- a/packages/browseros-agent/apps/claw-server-rust/src/db/audit_log.rs
+++ b/packages/browseros-agent/apps/claw-server-rust/src/db/audit_log.rs
@@ -24,6 +24,7 @@ use url::Url;
 pub use crate::db::entities::tool_dispatches::Model as ToolDispatchRow;
 
 const ARGS_JSON_MAX: usize = 4096;
+const SUSTAINED_ERROR_TAIL: usize = 3;
 
 #[derive(Clone)]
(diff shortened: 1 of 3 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1985, opened 2026-07-22
refactor(claw): move runtime file publisher into services

shadowfax92:
## Summary
- move the `runtime.json` publisher under the Rust Claw service layer and remove its root boundary exception
- preserve the exact filename, JSON bytes, atomic replacement, best-effort warning, and canonical URL selection
- register publication with `AppRuntime` so startup remains non-blocking while shutdown owns the task lifetime

## Design
Listener binding and advertised URL selection remain in `main.rs`, where the actual bound address and optional Chrome proxy port are available. The stateless service module owns only the runtime-file format and filesystem operation, while `AppRuntime::spawn_task` tracks the best-effort publication without awaiting it before `axum::serve`.

## Test plan
- [x] `bun run fmt:rust`
- [x] `cargo test -p claw-server-rust runtime_file`
- [x] `cargo test -p claw-server-rust --test dependency_boundaries`
- [x] `cargo test -p claw-server-rust --locked`
- [x] `bun run lint:rust`
- [x] `bun run test:rust`

### DIFF
--- a/packages/browseros-agent/apps/claw-server-rust/src/lib.rs
+++ b/packages/browseros-agent/apps/claw-server-rust/src/lib.rs
@@ -8,7 +8,6 @@ pub mod error;
 pub mod identity;
 pub mod ids;
 pub mod runtime;
-pub mod runtime_file;
 pub mod services;
 pub mod storage;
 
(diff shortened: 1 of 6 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1987, opened 2026-07-22
fix(claw-app): scale Audit tab navigation

shadowfax92:
## Summary
- replace the dense Session/Tab row with one compact, accessible Audit view selector
- preserve aggregate and per-browser-tab counts, screenshots, and timeline filtering
- keep session-only content inline and bound long option lists for narrow viewports

## Design
`TaskViewNavigation` composes the existing Base UI Select and renders the same `TabView` content built by `TaskDetailPage`. Multi-view sessions get one responsive selector; zero- and one-view sessions avoid unnecessary navigation.

## Test plan
- [x] focused selector tests: zero, one, invalid default, selection, and 21-option overflow
- [x] Audit grouping/filter helper tests
- [x] claw-app typecheck, lint, development build, and production build
- [x] live visual verification at wide and 386px viewports with session-only, one-tab, and 12-tab sessions
- [x] mouse and keyboard selection, count semantics, and horizontal-overflow checks

## Repository baseline
- `bun run check` retains two unchanged unsafe-optional-chaining errors in `apps/server/tests/lib/agents/acpx-runtime.test.ts`.
- `bun run test` retains the current-main claw-app `@/...` alias resolution failures; the new focused selector and existing grouping/filter suites pass directly.


shadowfax92:
@greptileai review

shadowfax92:
@claude review

shadowfax92:
Addressed review feedback in `a4fe7085f`:

- Removed the now-unreferenced `tabs-auto-hide` component and its test; repository-wide search showed the Audit page was its only production consumer.
- Removed the no-op `pointerdown` from the selector mock test.
- Kept the `value !== null` guard. The installed Base UI signature is `(value: string | null, ...)`; removing it fails claw-app typecheck with TS2322.

Re-verified 20 focused tests, claw-app typecheck, lint, and `git diff --check`.

PR #1989, opened 2026-07-23
refactor(agent): clarify harness integration ownership

shadowfax92:
## Summary
- establish a single shared harness catalog boundary for identity, install fingerprints, and nested MCP/skill surface metadata
- move MCP-only types, config-path resolution, management, planning, emission, and state I/O behind a coherent `mcp` module
- document the crate's MCP and skill responsibilities while preserving the existing flat public facade and persisted formats

## Design
This is more than a folder shuffle: MCP-only resolver behavior and the mostly-MCP root type collection move into the MCP domain, while `catalog.rs` retains only shared identity and immutable facts for the seven harnesses. Both MCP and skills consume that one catalog. `lib.rs` deliberately re-exports the existing flat API, and the crate-level `Error` remains the intentionally shared compatibility envelope because both domains already return its public variants.

## Test plan
- [x] `cargo test --locked -p harness-integrations`
- [x] `cargo clippy --locked -p harness-integrations --all-targets -- -D warnings`
- [x] `cargo test --locked -p claw-server-rust --test connections`
- [x] `cargo check --workspace --all-targets`
- [x] adversarial local review: clean

### DIFF
--- a/packages/browseros-agent/crates/harness-integrations/README.md
+++ b/packages/browseros-agent/crates/harness-integrations/README.md
@@ -1,18 +1,37 @@
 # harness-integrations
 
-`harness-integrations` registers and deregisters MCP servers in the real configuration files used by AI coding agents. A workspace `manifest.json` records which server entry the library wrote to each agent and the exact configuration path used.
+`harness-integrations` manages BrowserClaw integrations with AI coding harnesses. It has two runtime responsibilities:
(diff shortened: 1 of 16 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1990, opened 2026-07-23
feat: cockpit idle saved-stats band with durable efficiency stats

shadowfax92:
## Summary
- Shows a kinetic budget-bar saved-stats band on the idle ready Cockpit (All time / 30 days / 7 days), summarizing estimated tokens saved vs a screenshot-first agent, human time saved, sessions, and tool calls — hidden entirely when no measured history exists.
- Persists one immutable `session_efficiency_stats` projection per eligible ended session (migration m0007), finalized asynchronously after the audit end commits, with idempotent inserts, startup reconciliation for missed sessions, and an ordered shutdown drain.
- Exposes `GET /api/v1/cockpit/stats` through the generated OpenAPI contract (all three windows in one response, signed raw savings) and emits one privacy-safe raw `agent_session_efficiency_computed` analytics event per winning insert via a new bounded signed-integer allowlist kind.

## Design
Audit dispatch rows and the durable session end remain the single measurement source. A completion hook fires at the durable boundary (before any subsequent await) and queues background finalization through a mutex-gated task tracker, so teardown latency is unchanged and no finalizer can spawn after the shutdown drain. Efficiency-estimator v1 models one 1,920x1,080 screenshot (1,536 tokens via the shared patch estimator) per audited dispatch; raw savings stay signed through storage, API, and analytics — only the Cockpit presentation clamps to zero. The React band is pure presentation over generated DTOs, polls only while the resolved Cockpit is idle, and uses the shared Base UI tabs primitive for accessible window switching.

## Test plan
- [x] `bun run --filter @browseros/claw-app test|typecheck|lint|build` — component states (active/onboarding/idle/loading/error/unmeasured/zero-window/negative-savings), keyboard tabs, reduced motion
- [x] `cargo test -p claw-server-rust` (269 tests) — formula, migration fresh+upgrade, insert races, reconciliation partial-cancel, shutdown ordering, analytics allowlist rejections, route shapes
- [x] `bun run codegen:claw-api:check` + contract/fixture/client tests (signed + zero-window fixture round-trips TS and Rust)
- [x] `cargo clippy --workspace --all-targets --locked -- -D warnings` and `cargo fmt --all --check`


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #1996, opened 2026-07-23
feat(code-mode): collapse to one audited, ownership-aware script tool

DaniAkash:
## What

Builds the code-mode-primary interface on the Rust backend: the agent drives the browser by writing one JavaScript program against a `browser` SDK that runs server-side, instead of paying a model turn per action across a dozen granular tools. This is the whole substrate: a single script tool, at tool parity, with every in-script primitive ownership-checked and audited, the granular tools collapsed off the surface, and the cockpit (audit, replay, tab groups, ownership) working through the script tool exactly as it did through the granular ones.

## Landed here

**Inner-call ownership and audit.** A script's primitives used to reach straight into the shared browser session, skipping the guards and audit the pipeline runs per tool. A host-injected `InnerCallHook` (dependency inversion, since the tool crate cannot call up into the server) now authorizes each primitive against page ownership and records it as a child audit row.

**Audit hierarchy.** A nullable `parentDispatchId` links a primitive to the script dispatch it ran inside, via a named migration, and is surfaced on the audit contract (`dispatchKey` + `parentDispatchId`) so the cockpit can group and replay a script's primitives instead of seeing a black box. Session dispatches order by creation time, so a script reads as run-then-its-primitives rather than by write id.

**Tool parity.** The `browser` SDK reaches full parity with the granular tool set. Rather than reimplement each tool, a `tool:` bridge method dispatches through the real handler, so read, grep, wait, screenshot, evaluate, download, pdf, upload, tabGroups, and windows reuse that tool's validation and output. The bridge carries the full context, so the ownership and audit hook wraps these too. `read`/`grep` return text (not a wrapped object), and the SDK contract is documented exactly so the agent does not probe return shapes and churn tabs.

**Surface collapse.** The host advertises only the script tool plus `name_session`; the granular tools are filtered out of the listed set, so the agent sees only code mode. The granular tools stay callable by name (for the bridge's dispatch and the server's own end-to-end tests) and remain the functions the bridge dispatches to, so no capability is lost. No flag: this branch is the staging mechanism.

**Multi-agent tab ownership.** Pages a script creates are claimed and grouped for the agent exactly as `tabs new` would, so a script's tabs get the agent's ownership, tab group, and the cockpit ownership window. New pages open in the background by default so a working agent does not steal the user's focus. `browser.pages.list()` returns the same tri-bucket ownership view the granular tabs-list produces: every open tab is tagged `ownership` "mine" | "user" | "other-agent" (with the owning agent's label), so a script can tell its own tabs from the user's and other agents' and act only on its own. This is classification, not a hard restriction: authorization still allows the user's tabs when the user asks the agent to work on one, and rejects another agent's tabs.

**Cockpit fidelity for code mode.** Per-step screenshots are captured for a script's page-targeting primitives, code-mode sessions report liveness and tab activity so they do not appear stuck, and a script's tabs open the replay window, so audit, replay, and the live view render a script the same way they render granular tool runs.

**Operating guide** rewritten for code mode: one tool that runs JavaScript against the browser SDK, the SDK surface, the compose-the-whole-flow-in-one-script loop, batching independent work with `Promise.all`, and splitting a large workload across subagents (each gets isolated tabs). The installed BrowserClaw skill is a short pointer that steers agents to this tool and defers the how-to to the tool description.

**Persistence seam.** Traversal-safe storage for reusable `.js` a script saves per host under `helpers/<host>/<name>.js`, the layer the self-healing work builds on.

## Tests

`browseros-mcp` 52 and `claw-server-rust` 235 lib tests pass; clippy and rustfmt clean on the changed crates; the claw-mcp contract suite expects the collapsed catalog. Coverage includes: the bridge invokes the hook with the right page and a rejection blocks the primitive before dispatch; the host hook rejects a foreign page and writes a linked child row; a tool-backed primitive routes through the handler and is hook-wrapped; `pages.list` routes through the hook and is annotated with the tri-bucket ownership view; the host classifies pages into the three buckets; the collapsed surface exposes only the script tool; code-mode tabs open the replay window; helper storage round-trips and rejects unsafe segments; schema and migration-count tests updated.

## Scope notes and follow-ups

- Page authorization rejects peer-owned pages and, by design, allows the user's own (unclaimed) pages, so an agent can work on a user tab when the user asks. Raw CDP (no page argument) is not ownership-checked yet.
- A formal typed `.d.ts` delivered as an MCP resource, the wider hostile-script and end-to-end sweep, the cockpit frontend rendering of the group (separate app), renaming the tool to `execute`, fully blocking granular calls plus migrating the tests that drive them by name, and a `setTimeout`/`setInterval` shim in the script runtime (agents reach for it habitually) are follow-ups.

Targets the integration branch for testing.


DaniAkash:
@greptileai

PR #2001, opened 2026-07-23
fix(agent): reconcile aliased skill targets by identity

shadowfax92:
## Summary
- reconcile managed BrowserClaw skills by physical skill-root identity so harness aliases share one safe installation
- migrate compatible schema-v1 records while preserving foreign destinations, including retargeted legacy aliases
- point BrowserClaw dev and dogfood launchers at the checked-in Rust app resource root

## Design
Catalog paths remain lexical and harness-native. The reconciler plans desired, recorded, and cleanup identities before mutation by canonicalizing the deepest existing ancestor of each parent skill root, then appending missing components and the un-followed final skill name. Only records already stored at physical identity retain manifest-only ownership; migrated aliases require the ownership marker before controlling an existing destination.

## Test plan
- [x] `go test ./...` in `tools/dev`
- [x] `go test ./...` in `tools/dogfood`
- [x] `git diff --check`

### DIFF
--- a/packages/browseros-agent/apps/claw-server-rust/tests/connections.rs
+++ b/packages/browseros-agent/apps/claw-server-rust/tests/connections.rs
@@ -166,6 +166,7 @@ async fn run_connections_case() -> anyhow::Result<()> {
         fs::read_to_string(shared_skill.join("SKILL.md"))?,
         "managed skill v1\n"
     );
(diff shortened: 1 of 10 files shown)


shadowfax92:
@claude review

shadowfax92:
@greptileai review

PR #2005, opened 2026-07-23
feat(claw): move audit observation off the tool path

shadowfax92:
## Summary
- add a post-effect `ToolObserver` stage so audit sees the final wire result without changing it
- queue bounded, persistence-ready audit events through one process-wide worker with ordered writes and batched task projection
- coalesce session previews under real concurrency limits and drain accepted work at session close and runtime shutdown

## Design
Audit now submits an owned `AuditEvent` after ordered effects and returns once the bounded queue admits it. A single `AuditWorker` serializes database appends, batches each dirty session's task projection, and runs at most one preview plus the latest pending preview per session under a global cap. Session teardown flushes accepted audit work before cancellation reconciliation and terminal persistence, while runtime shutdown applies bounded session and audit drain budgets before releasing browser resources.

## Test plan
- [x] adversarial review of the full branch diff, followed by a clean second pass after fixes


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #2010, opened 2026-07-23
fix(claw): use session duration for human time saved

shadowfax92:
## Summary
- define human time saved as the elapsed duration of each completed, measured session
- preserve token estimates, eligibility, aggregation windows, session counts, and tool-call counts
- backfill reconstructable version-3 projections from their matching activity-card task durations

## Design
The efficiency projection now uses the same lifecycle start, first-dispatch fallback, end timestamp, and nonnegative elapsed-duration semantics as activity cards. Formula version 4 distinguishes the new meaning. Migration 11 updates only version-3 rows whose task matches on session ID, end timestamp, and dispatch count, leaving retained or reused history untouched.

## Test plan
- [x] Open a consistent copy of the development database with the new migrator and verify both measured rows total `701245ms` at version 4
- [x] Query the copied database through `/api/v1/cockpit/stats` and verify `humanTimeSavedMs: 701245`, 2 sessions, and 37 tool calls


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #2016, opened 2026-07-23
fix(claw): exclude red from automatic tab-group colors

shadowfax92:
## Summary
- Remove red from BrowserClaw's deterministic automatic tab-group palette.
- Preserve red in the shared color enum and explicit MCP/CDP tab-group updates.
- Lock the eight-color automatic palette order with a focused regression test.

## Design
Automatic assignment continues hashing each agent slug into a stable ordered palette; only the red entry is removed. Explicitly requested red groups remain supported because the session color representation, MCP enum, and CDP definitions are unchanged.

## Test plan
- [x] `cargo test -p claw-server-rust services::browser::tab_groups::tests`


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #2019, opened 2026-07-24
feat: follow replay activity across browser tabs

shadowfax92:
## Summary

- Replace rrweb-owned tab-local playback with one BrowserClaw-owned global session clock and a deterministic completion-minus-duration camera schedule.
- Keep unrelated tab DOM streams isolated behind a bounded stage with one active player and at most one paused standby.
- Preserve global position across explicit tab inspection, while keeping no-visual actions in the timeline and existing same-tab stitching and gap warnings intact.

## Design

This implements the approved bounded double-buffered architecture. A pure session plan projects absolute global time into independent tab tracks, and a pure camera reducer coalesces parallel candidates with a 10-second real-time dwell that is independent of playback speed. rrweb is only a visual sink: promotion is generation-guarded and presentation metadata changes only after the target player is ready, with throttled 250 ms drift correction rather than per-frame seeks. The change is frontend-only.

## Test plan

- [x] Run the full `@browseros/claw-app` test suite.
- [x] Run `@browseros/claw-app` typecheck.
- [x] Run targeted Biome checks for the replay screen and replay API model.
- [ ] Open a replay with sequential and overlapping activity across at least two recorded tabs; confirm the camera follows activity without rapid intermediate jumps.
- [ ] Exercise 1x, 2x, and 4x; confirm a long-lived tab remains visible for about 10 wall-clock seconds at each speed, while a short track switches when its visuals end.
- [ ] Scrub globally and select timeline actions; confirm the replay pauses and resolves the expected tab, URL, caption, and warning.
- [ ] Click both the followed tab chip and another tab chip; confirm inspection starts at local zero, then `Resume session` restores the exact global position and remains paused.
- [ ] Confirm same-tab navigation stays one track, no-visual actions remain in the timeline without taking the viewport, completion restarts from zero, and rapid switches never mount more than two players.

### DIFF
--- a/packages/browseros-agent/apps/claw-app/modules/api/replay.hooks.ts
+++ b/packages/browseros-agent/apps/claw-app/modules/api/replay.hooks.ts
@@ -25,8 +25,18 @@ export type ReplayVerb =
 export type ReplayKind = 'action' | 'block' | 'done'
 
 export interface ReplayFrame {
-  /** Seconds into the session. */
+  /**
+   * Dispatch completion in seconds from session start. The timeline and audit
+   * highlight use completion because that is the timestamp persisted with the
+   * finished dispatch.
+   */
   t: number
+  /**
+   * Approximate dispatch start in session seconds, derived from completion
+   * minus duration. The automatic camera uses this separate clock so it can
+   * show a tab while work is happening without rewriting the audit timestamp.
+   */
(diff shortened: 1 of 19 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #2028, opened 2026-07-24
feat(app): permanent BrowserClaw banner on MCP settings

shadowfax92:
## Summary

- Adds a **permanent** (non-dismissible) BrowserClaw banner to the BrowserOS MCP settings page, telling anyone wiring up the MCP server that BrowserClaw is the stronger MCP surface for their agents, and sending them to `browserclaw.ai`.
- New component rather than a reuse of `components/promo/BrowserClawPromoBanner`: that one gates on `browserClawPromoDismissedStorage`, a single **global** WXT key. Honouring it would hide this banner from everyone who dismissed the unrelated new-tab promo — the opposite of permanent. This one has no dismiss control and no persisted visibility at all.
- Adds `BROWSERCLAW_MCP_BANNER_CLICKED_EVENT` (`settings.browserclaw_mcp_banner.clicked`). No dismiss event, since there is no dismiss.

## Design

`BrowserClawMcpBanner` is a stateless FC colocated in `screens/mcp-settings/`, matching `McpPromoBanner` (also single-consumer, also colocated) rather than the shared `components/promo/` folder. The card shell, typography and outline accent-orange CTA are copied verbatim from the two shipped banners so the three read as one family; the one intentional divergence is the missing dismiss `<button>`. The URL lives as a module-level const, the same way `MCPServerHeader.tsx` in this very folder holds its `DOCS_URL`.

It sits directly **under** `<MCPServerHeader />` rather than above it — `ai-settings/McpPromoBanner`'s "Set up" CTA navigates users to this page to do exactly one thing, and burying the server URL / copy / restart controls under a promo would fight that.

Copy is deliberately MCP-specific, since a user can land here having just read the generic BrowserClaw promo on the AI settings page. Every claim is traceable:

| Line | Source |
|---|---|
| "A browser built for AI agents" | README "The browser for AI agents"; `browserclaw/SKILL.md` "The user's dedicated browser for agents" |
| "a bigger MCP toolset" | `contracts/claw-mcp/README.md` — 17 tools (`tabs`, `tab_groups`, `windows`, `navigate`, `snapshot`, `diff`, `act`, `grep`, `read`, `evaluate`, `run`, `screenshot`, `pdf`, `wait`, `upload`, `download`, `name_session`) |
| "your real logins" | README "using the accounts you're already signed into" |
| "session replay" | README "replay any session like a video" |

## What it looks like

Rendered on the real page (light and dark), directly below the server header:

> **For better MCP support, use BrowserClaw**
> A browser built for AI agents — a bigger MCP toolset, your real logins, and session replay
> `[ browserclaw.ai → ]`

BrowserClaw logo on the left, outline accent-orange CTA on the right, no dismiss X.

## Review fixes (second commit)

An adversarial review of the first commit found two real test gaps. Both were reproduced before being fixed:

- **The analytics assertion was tautological.** The test mocked `@/lib/constants/analyticsEvents` and then asserted against its own literal, so the real constant was never loaded. Setting the shipped value to `'BROKEN.value.here'` left the suite green. It now imports the constant via a relative specifier — which the `@/`-keyed mock doesn't intercept — so the assertion sees the real value. Against the same sabotage it now fails correctly. (The mock can't just be dropped: `@/*` doesn't resolve under `bun test` from `packages/browseros-agent`, which is why every test here mocks its `@/` imports.)
- **Nothing guarded that the page mounts the banner.** Deleting `<BrowserClawMcpBanner />` left all 13 tests passing. Added `MCPSettingsPage.test.tsx`, which asserts the banner renders *between* the server header and integrations, so the placement decision is guarded too, not just presence. Mutation-checked both ways.
- `chrome.tabs.create` is now awaited in a try/catch reporting via `sentry.captureException`, per the app rule on capturing runtime errors; a test forces the rejection path.
- Dropped two "does not repeat sibling copy" assertions — they hardcoded strings from files the test doesn't import, so a reword elsewhere would have silently turned the guard into a no-op. The constraint lives in the component's doc comment instead.

**Knowingly deferred:** the card shell and the 4-clause accent-orange CTA className are now duplicated across three banner files (this one, `McpPromoBanner`, `BrowserClawPromoBanner`). Extracting a shared `PromoBannerCard` would mean editing two already-shipped banners on unrelated screens, which is outside the scope of adding one banner to one page. Worth doing before a fourth copy appears.

## Test plan

- [x] `bun run check` (lint + typecheck + fallow) — green
- [x] `bun run test` — green
- [x] `cd apps/app && bun run typecheck` — green
- [x] `cd apps/app && bun run test` — green
- [x] Unit tests cover: the banner renders its copy; renders exactly one `<button>` and no dismiss control; the CTA tracks the real event constant and calls `chrome.tabs.create({ url: 'https://browserclaw.ai' })`; a failed tab-open reports to Sentry; and `MCPSettingsPage` mounts the banner in the right slot
- [x] Verified live via the CDP inspector on `#/settings/mcp`: banner renders in the right slot, one button, logo present, correct copy, correct in dark theme — and clicking the CTA really opened `browserclaw.ai` (which resolves to the BrowserClaw page)

### DIFF
--- a/packages/browseros-agent/apps/app/lib/constants/analyticsEvents.ts
+++ b/packages/browseros-agent/apps/app/lib/constants/analyticsEvents.ts
@@ -65,16 +65,16 @@ export const MCP_PROMO_BANNER_CLICKED_EVENT =
   'settings.mcp_promo_banner.clicked'
 
 /** @public */
+export const BROWSERCLAW_MCP_BANNER_CLICKED_EVENT =
+  'settings.browserclaw_mcp_banner.clicked'
+
+/** @public */
 export const BROWSERCLAW_PROMO_BANNER_CLICKED_EVENT =
   'ui.browserclaw_promo_banner.clicked'
 
 /** @public */
 export const BROWSERCLAW_PROMO_BANNER_DISMISSED_EVENT =
   'ui.browserclaw_promo_banner.dismissed'
-
-/** @public */
-export const BROWSERCLAW_MCP_BANNER_CLICKED_EVENT =
-  'settings.browserclaw_mcp_banner.clicked'
 
 /** @public */
 export const MCP_EXTERNAL_ACCESS_ENABLED_EVENT =
--- a/packages/browseros-agent/apps/app/lib/constants/analyticsEvents.ts
+++ b/packages/browseros-agent/apps/app/lib/constants/analyticsEvents.ts
@@ -73,6 +73,10 @@ export const BROWSERCLAW_PROMO_BANNER_DISMISSED_EVENT =
   'ui.browserclaw_promo_banner.dismissed'
 
 /** @public */
+export const BROWSERCLAW_MCP_BANNER_CLICKED_EVENT =
+  'settings.browserclaw_mcp_banner.clicked'
+
+/** @public */
 export const MCP_EXTERNAL_ACCESS_ENABLED_EVENT =
   'settings.mcp_external_access.enabled'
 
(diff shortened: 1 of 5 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #2030, opened 2026-07-25
feat: add bpatch ls command

shadowfax92:
## Summary
- Add top-level `bpatch ls` for machine-level listing of the configured patch store and registered Chromium checkout aliases.
- Dispatch `ls` before checkout/store discovery so it works from any directory, including the BrowserOS repo.
- Document human/JSON output, empty config behavior, `--store` display override, and `-C/--checkout` rejection.

## Design
`bpatch ls` reads `~/.config/bpatch/config.toml` through the existing config model and combines it with the optional global `--store` override in a small serializable report. The command is early-dispatched next to `alias` and `init`, avoiding Chromium checkout preflight and store validation.

## Test plan
- [x] `cargo fmt --all --check && cargo test --locked` from `packages/browseros/tools/bpatch`
- [x] Local adversarial review loop: 1 finding fixed, second pass clean

### DIFF
--- a/packages/browseros/tools/bpatch/README.md
+++ b/packages/browseros/tools/bpatch/README.md
@@ -31,6 +31,17 @@ ch2 = "/Users/shadowfax/ch2-src"
 
 Without a checkout selector, `bpatch` discovers the checkout from the current directory as before.
 
+List the machine-level paths from anywhere:
+
+```console
+$ bpatch ls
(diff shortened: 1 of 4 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #2045, opened 2026-07-27
fix(claw-app): make the session-history poll opt-in (stops audit request flood)

DaniAkash:
## Problem

The `/audit` page fires a never-stopping stream of `GET /api/v1/sessions?limit=100` requests. It keeps requesting long after the list has loaded instead of stopping at the end of the data.

## Root cause

`useSessions` is the only infinite query in the app, and it baked `refetchInterval: 3000` into the shared factory. On an infinite query, each interval tick refetches **every loaded page**, so any paginated consumer inherits an unbounded, continuous request stream. The audit view was the visible victim.

The interval was also a *silent* default: the cockpit onboarding probe's own comment already states "elsewhere in the app react-query's default no-polling behaviour is unchanged" — the author assumed the factory did not poll. The baked-in interval contradicted that intended design and forced polling on every non-opting consumer.

I verified the behaviour is client-side and visibility-gated (zero requests fired while the tab was backgrounded, which rules out a render or query-key loop and pins it on the interval). The server pagination itself terminates correctly.

## Fix: make polling opt-in per consumer

Drop the `refetchInterval` default from the shared factory so an infinite query no longer polls unless a consumer asks for it:

- **RecentActivity** (single-page homepage feed) opts in explicitly with `refetchInterval: 3000`. It never paginates, so a tick is one request, not an all-pages refetch.
- **The cockpit onboarding probe** already opts in with its own conditional interval; unchanged.
- **The audit view** simply does not opt in, which is the correct default for a historical review surface. Freshness still comes from refetch on window focus, on remount, and from the cleanup mutation's invalidation.

This is the whole codebase: `useSessions` is the only `createInfiniteQuery`. The remaining `refetchInterval`s are on regular single-result queries (live sessions, screenshots, storage), where a bounded poll is a legitimate default and is left as-is.

## Validation

- `typecheck` passes.
- `biome check` clean.
- Audit and cockpit test suites pass.

### DIFF
--- a/packages/browseros-agent/apps/claw-app/components/cockpit/RecentActivity.tsx
+++ b/packages/browseros-agent/apps/claw-app/components/cockpit/RecentActivity.tsx
@@ -30,6 +30,7 @@ const HOME_TASK_LIMIT = 12
 export function RecentActivity() {
   const query = useSessions({
     variables: { limit: HOME_TASK_LIMIT },
+    // Homepage feed: poll so new sessions surface without a manual refresh.
     refetchInterval: 3000,
   })
   const tasks = (query.data?.pages ?? [])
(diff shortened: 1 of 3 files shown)


DaniAkash:
@greptileai

PR #2052, opened 2026-07-28
feat(claw-app): replace first-run demo with a how-to video bento

DaniAkash:
## What

Auto-connect now wires up agents at launch, so the first-run cockpit screen no longer needs to lead with "install the MCP." This swaps the single setup demo for a **video-first how-to bento** that teaches how to put BrowserClaw to work with your agents.

## Changes

- **Video bento**: a mixed-size grid of poster tiles (one featured, a stacked rail, a wide tile) that open a shared lightbox player. Desktop is a 3-column bento; mobile collapses to a single column of 16:9 tiles.
- **Lite-YouTube facade**: tiles render as posters and exactly one `youtube-nocookie` iframe mounts, only when a tile is clicked, so nothing streams (and no cookies are set) until the reader asks. Content lives in a small manifest; posters derive from the video thumbnail with an optional per-video `poster` override.
- **Retires the obsolete pieces**: the single CDN demo video and the numbered `01/02/03` step strip. The strip is replaced by a **connected-agents line** read from the live connection list, so the auto-connect result is visible.
- **Hero reframed** from setup to learning.
- **CSP**: an explicit `extension_pages` policy allows the nocookie embed via `frame-src`, keeping `script-src`/`object-src` at `'self'`; `img-src` and `connect-src` stay unlisted so poster URLs and the local API keep loading.

## Placeholder content

The line-up currently repeats two BrowserOS videos to fill the five tiles. Real how-to clips and branded poster URLs swap into the manifest later without touching layout.

## Validation

- `typecheck`, `biome`, and the cockpit test suites pass (65 tests). The bento and lightbox render under SSR.
- Note: I could not screenshot the running build from the worktree (the fresh WXT dev server does not serve the newtab over plain HTTP the way the primary dev instance does). Visual proportion tuning (row height, poster crop) may want a pass once viewed in a real newtab.


DaniAkash:
@greptileai

PR #2053, opened 2026-07-28
fix(app): don't save assistant history in incognito mode (#1189)

### LINKED ISSUE #1189: Assistant history being saved in incognito mode
### Issue Type

Browser Issue

### Operating System

macOS

### Description of the bug

Assistant history is being saved even when the browser is in incognito mode, which is unexpected behavior. It would also be nice to have a feature for temporary assistant chats, that are never saved to history.

### Steps to Reproduce

1. Open a window in incognito mode.
2. Use the assistant.
3. Close the window.
4. Check history in a normal window and see that the incognito assistant history was saved.

### Screenshots / Videos

_No response_

### BrowserOS Version

146.0.7821.31 (Official Build) (arm64)

### Additional Context

_No response_

DaniAkash:
## Problem

Closes #1189. `chrome.storage.local` is a single store shared between an extension's normal and incognito contexts (it isn't partitioned by incognito), and there was no incognito check anywhere. So an assistant chat started in an incognito window was persisted to the same store (and to the cloud when signed in) and showed up in normal-window history after the incognito window closed.

## Fix

Gate every durable history write on a `shouldPersistHistory(isIncognito)` check, resolved once per session via `chrome.extension.inIncognitoContext` (synchronous, and true for an extension page opened in an incognito window, so the gate can't race an async lookup). When incognito:

- the turn-end conversation save is skipped (both the local `conversations` store and the cloud upload);
- execution-history writes are skipped. The active task is still tracked in memory so the live in-session view keeps working; it just never becomes durable.

Nothing else changes. A normal-window chat saves exactly as before.

## UI

While the assistant runs in an incognito window, the chat shows a small notice pinned above the input: "Incognito mode: this chat is not saved to history."

## Scope and follow-up

The `shouldPersistHistory(isIncognito, isTemporaryChat)` helper leaves a seam for a future user-facing "temporary chat" toggle (never-saved chats in a normal window), which the maintainer noted fits alongside this. That toggle is not built here.

## Testing

Unit tests cover `shouldPersistHistory`. Type check, lint, and the extension build all pass. Runtime QA needs the extension allowed in incognito: open the assistant in an incognito window, run a chat, close it, and confirm nothing appears in normal-window history or the cloud, and that the notice shows; then confirm a normal-window chat still saves as before.

### DIFF
--- a/packages/browseros-agent/apps/app/lib/browseros/incognito.ts
+++ b/packages/browseros-agent/apps/app/lib/browseros/incognito.ts
@@ -1,0 +1,10 @@
+/**
+ * Whether the current extension page runs in an incognito context. Assistant
+ * history is never persisted there (#1189). `chrome.extension.inIncognitoContext`
+ * is synchronous and reads true for an extension page (side panel, newtab)
+ * opened in an incognito window, so the gate can be applied without racing an
+ * async lookup.
+ */
+export function isIncognitoContext(): boolean {
+  return chrome.extension?.inIncognitoContext ?? false
+}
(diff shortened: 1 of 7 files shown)


DaniAkash:
@greptileai

PR #2054, opened 2026-07-28
fix: preserve pinned CDP protocol bytes on checkout

shadowfax92:
## Summary
- mark the byte-pinned Chromium CDP protocol input as non-text in Git attributes
- prevent Windows checkout filtering from rewriting LF bytes to CRLF
- preserve the existing protocol source, SHA pin, build script, release workflow, and v0.0.18 release state

## Design
The CDP build script intentionally hashes the raw protocol bytes before parsing. A file-specific `-text` rule is the repository-level expression of that contract, preventing Git end-of-line conversion for every checkout without weakening the exact-byte guard.

## Test plan
- [x] simulated Windows checkout filtering hashes `protocol.json` to `84478a2eb6c7d088387aebdc0f5c738e3a87078c6c7bb7e72909dc2bf0de92b9`
- [x] `git check-attr text` reports `unset` for `protocol.json`
- [x] `cargo test --locked -p browseros-cdp`
- [x] `cargo check --locked -p claw-server-rust --bin browseros-claw-server-rs`


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #2060, opened 2026-07-29
fix(ci): recover Windows depot_tools cache safely

shadowfax92:
## Summary
- repair restored depot_tools only when every tracked difference is CR-at-EOL-only and the caller explicitly opts into disposable CI recovery
- fail closed on substantive changes, trailing whitespace, invalid repositories, or hidden index flags; preserve untracked files and default developer checkouts
- invalidate the incompatible checkout cache with schema v2 and document the observed Windows failure signature
- add Linux real-Git coverage and a native Windows Git Bash to cmd batch regression that proves the blocked update recovers and ends clean

## Design
The source provisioner owns depot_tools readiness immediately before gclient. The reusable build workflow explicitly enables a default-off repair at checkout and sync. Classification validates the repository and index, compares staged and unstaged data with only CR-at-EOL ignored, resets committed tracked content only after that proof, and verifies cleanliness. The immutable checkout cache moves to v2 with no v1 fallback so the first post-migration run is cold and deterministic.

## Test plan
- [x] 814 bos_build tests pass (2 platform-only skips on macOS)
- [x] 9 release-secret tests pass
- [x] 47 vendor_uploads tests pass
- [x] Ruff, actionlint, CLI smoke, and diff checks pass
- [x] Three adversarial review rounds end clean after fixing the one verified Important finding
- [ ] GitHub Windows Git bootstrap job runs the native Git-for-Windows CRLF recovery regression

### DIFF
--- a/.github/workflows/build-browseros.yml
+++ b/.github/workflows/build-browseros.yml
@@ -153,7 +153,7 @@ jobs:
 
           {
             echo "version=$version"
-            echo "cache_key=chromium-src-${{ inputs.platform }}-${{ inputs.arch }}-v1-$version"
+            echo "cache_key=chromium-src-${{ inputs.platform }}-${{ inputs.arch }}-v2-$version"
           } >> "$GITHUB_OUTPUT"
           {
             echo "CHROMIUM_ROOT=$parent/chromium"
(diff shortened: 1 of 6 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #2067, opened 2026-07-30
fix(claw): stop recommending isolation windows

shadowfax92:
## Summary
- Keep BrowserClaw parallel work scoped to independent tabs
- Remove the prompt instruction that recommends creating a separate window for isolation
- Add a regression assertion for tabs-only parallel-work guidance

## Design
This is a narrow prompt-only change: retain the existing five-tab parallelism limit, remove only the separate-window sentence, and update the colocated test to guard the intended policy.

## Test plan
- [x] cargo fmt --all -- --check
- [x] cargo test -p claw-server-rust


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #2095, opened 2026-08-05
chore(browserclaw): support BrowserOS neo aliases

shadowfax92:
## Summary
- Preserve the existing default-browser guidance for the BrowserClaw skill.
- Treat requests to use neo or BrowserOS as requests for BrowserOS neo over MCP.

## Design
Append one explicit alias-routing sentence to the skill frontmatter description, keeping the existing description and body unchanged. This makes the aliases visible during skill selection while retaining the stronger default-browser behavior.

## Test plan
- [x] `cargo check -p claw-server-rust`
- [x] `git diff --check`


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #2119, opened 2026-08-05
fix: retain agent tabs and groups for one hour

shadowfax92:
## Summary
- Retain finished-session agent tabs for one hour instead of three minutes.
- Align retained tab-group closure to the same one-hour policy.
- Document that retention lets users inspect agent-opened tabs while browser-side work settles.

## Design
The existing session retention setting is now the single source of truth for both retained groups and orphaned-tab cleanup. Its default is one hour, and CLAW_SESSION_RETENTION_MS continues to override the policy for both paths. The cleanup sweep remains at its independent 60-second polling interval.

## Test plan
- [x] cargo fmt --all --check
- [x] cargo check -p claw-server-rust
- [x] cargo clippy --workspace --all-targets --locked -- -D warnings

Unit tests were intentionally not run per workspace instructions.


shadowfax92:
@claude review

shadowfax92:
@greptileai review

PR #2127, opened 2026-08-06
fix(release): skip unavailable default publish platforms

shadowfax92:
## Summary
- publish every available platform when `--platform` is omitted
- warn with all missing platform metadata while keeping explicit platform requests strict
- validate and build candidates only from metadata that was actually loaded

## Design
`PublishModule` now preserves whether platform selection was omitted, partitions the default selection into loaded and missing platforms, and limits validation and promotion to loaded metadata. Explicit selections retain the prior fail-closed behavior.

## Test plan
- [x] `uv run python -m unittest bos_build.release.publish_test bos_build.cli.release_test -v`
- [x] `uv run python -m unittest discover -s bos_build -t . -p "*_test.py" -v`
- [x] CI-equivalent CLI smoke and product doctor
- [x] `uv run pyright bos_build/release/publish.py`

### DIFF
--- a/packages/browseros/bos_build/cli/release_test.py
+++ b/packages/browseros/bos_build/cli/release_test.py
@@ -734,12 +734,30 @@ class PublishCliIntegrityTest(unittest.TestCase):
         ):
             yield
 
-    def test_missing_requested_platform_makes_actual_cli_exit_nonzero(self):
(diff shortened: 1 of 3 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #2138, opened 2026-08-06
fix(claw-server): preserve useful browser pages

shadowfax92:
## Summary

- Tell Claw agents to leave pages open when users may want to inspect them.
- Remove "close when done" guidance from the shared BrowserOS MCP instructions and `run` tool.
- Stop the bulk-flow example from closing every page it opens.

## Design

The one-hour reconciliation sweep remains the cleanup backstop. This change removes the separate, immediate agent guidance that bypassed that grace period. Close APIs remain available, but their descriptions are capability-only.

## Test plan

- `cargo fmt --all --check`
- `cargo check -p browseros-mcp -p claw-server-rust`
- `git diff --check`
- Unit tests not run per repository instructions.


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #2141, opened 2026-08-06
docs(browserclaw): restore self-contained skill guide

shadowfax92:
## Summary
- restore the self-contained BrowserOS neo operating guide installed across supported agent harnesses
- keep tool choice task-based, with run available for multi-step flows instead of being the default
- mirror the balanced tool guidance in MCP server instructions and document why the installed skill must remain self-contained

## Design
The managed skill carries the shared-tab rules, snapshot-act-verify loop, output guidance, and failure behavior so hosts do not depend on server instructions being surfaced. Tool descriptions remain authoritative for exact contracts, while the MCP prompt uses the same balanced act/run/evaluate guidance.

## Test plan
- [x] cargo fmt --all --check
- [x] cargo test --workspace --locked
- [x] cargo clippy --workspace --all-targets --locked -- -D warnings


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #2147, opened 2026-08-06
fix(server): accept legacy chat requests

shadowfax92:
## Summary
- restore BrowserOS chat when an OTA-updated server receives the pre-refactor extension payload without `target`
- normalize only valid untargeted LLM requests to an explicit BrowserOS target
- keep malformed targets and legacy ACP-provider requests rejected so ACP authorization remains explicit
- run `claw-onboard` tests from that app's project root so Bun resolves its TypeScript path aliases

## Design
The BrowserOS branch of `ChatRequestSchema` accepts an optional target on input and transforms every successful parse into the canonical required target shape. The target uses `providerId`, falling back to the validated provider type, while the ACP schema and trusted-origin checks remain unchanged.

The shared per-file Bun runner now accepts an optional child-process working directory. Only `claw-onboard` opts into it, matching the app root already used by its TypeScript and Vite builds.

## Test plan
- [x] `bun test apps/server/tests/api/chat-request-schema.test.ts` (9 passed)
- [x] `bun run --filter @browseros/server typecheck`
- [x] `bun run --cwd apps/claw-onboard test`
- [x] `bun run check`
- [x] `bun run test`

### DIFF
--- a/packages/browseros-agent/apps/claw-onboard/package.json
+++ b/packages/browseros-agent/apps/claw-onboard/package.json
@@ -9,7 +9,7 @@
     "build": "tsc --noEmit && vite build",
     "build:chromium": "tsc --noEmit && vite build --mode chromium && bun scripts/verify-chromium-build.ts",
     "preview": "vite preview",
-    "test": "bun run ../../scripts/run-bun-test.ts ./apps/claw-onboard",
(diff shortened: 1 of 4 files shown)


shadowfax92:
@claude review

shadowfax92:
@greptileai review

PR #2164, opened 2026-08-09
feat(claw-app): restyle the MCP page and calm the audit page's live signals

felarof99:
## Summary
- **MCP page** restyled to the supplied design: cobalt endpoint block at radius 9, title-case 12px sans labels replacing mono-uppercase, hairline agent rows with no status dot, white Claude Desktop callout.
- **Audit page** live signals calmed: the pulsing Live chip becomes a static `#2FE08C` pill, the per-agent colour dot comes off the hover card, and the filter dropdowns are opaque white instead of translucent-over-blur (they were reading as a gradient).
- **Ledger body text 11px → 12px**, with the tool-chain track widened 256 → 288px so the four-link chain still fits without mid-word clipping.

## Design
No new design tokens. Every colour in the mock already existed as a `--cyanotype-*` var from the recent-activity redesign, so this consumes that vocabulary rather than minting a third one. The state pills share one geometry across the table, the filter bar, the hover card, and `StatusBadge`; only Live carries a saturated fill.

Behaviour is untouched — connect/disconnect, the pending spinner, the error strip, and the install dialog all work as before. Cockpit live indicators deliberately keep pulsing.

## Test plan
- [x] `bun run test` — full suite, 136 files, 0 failures
- [x] `bun run lint` and claw-app typecheck clean
- [x] Both screens rendered and screenshotted against the mock
- [x] New coverage: static Live pill emits no `pulse-dot`; hover card renders no per-agent dot

### DIFF
--- a/packages/browseros-agent/apps/claw-app/components/audit/AuditHoverPreview.test.tsx
+++ b/packages/browseros-agent/apps/claw-app/components/audit/AuditHoverPreview.test.tsx
@@ -1,0 +1,47 @@
+import { describe, expect, it } from 'bun:test'
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query'
+import { renderToStaticMarkup } from 'react-dom/server'
+import type { TaskSummary } from '@/modules/api/audit.hooks'
+import { AuditHoverPreview } from './AuditHoverPreview'
+
+const task: TaskSummary = {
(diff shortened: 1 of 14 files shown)


felarof99:
@greptileai review

felarof99:
@claude review

PR #2280, opened 2026-08-13
fix(ci): isolate macOS Chromium builds in APFS workspaces

shadowfax92:
## Summary
- add an owned macOS Chromium APFS workspace helper that validates a pristine base, proves COW clone support, clones the full gclient root, records base identity, and safely cleans owned workspaces
- wire signed macOS release and both signed nightly workflows to build against the disposable clone src and clean it under `if: always()`
- add workflow and mocked shell tests plus runbook notes for the new lifecycle

## Verification
- `cd packages/browseros && uv run python -m unittest discover -s bos_build -t . -p "*_test.py"`
- `cd packages/browseros && uv run ruff check bos_build`
- `python3 -m unittest discover -s tools/release_secrets -t . -p "*_test.py"`
- `bash -n .github/scripts/macos-chromium-workspace.sh .github/scripts/macos-signing-keychain.sh`
- `git diff --check`

## Note
- `actionlint` still flags the repo-existing `queue: max` concurrency key in these workflows; this PR leaves that scoped behavior unchanged because existing tests assert it.

### DIFF
--- a/.github/scripts/macos-chromium-workspace.sh
+++ b/.github/scripts/macos-chromium-workspace.sh
@@ -323,8 +323,7 @@ setup_workspace() {
   [ "$(basename "$base_src")" = "src" ] || die "Persistent Chromium path must point to src: $base_src"
   base_root="$(resolve_existing_dir "$base_src/..")" \
     || die "Could not resolve Chromium gclient root from $base_src"
-  version_file="${version_file:-$(default_version_file)}"
-  verify_base "$version_file"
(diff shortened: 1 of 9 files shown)


shadowfax92:
Addressed the Greptile finding: the APFS helper now fails before cloning when top-level src or any nested git repo under the gclient root has tracked or untracked changes. Added mocked coverage for untracked base state and dirty nested gclient repos.

PR #2289, opened 2026-08-14
feat(server): give /mcp clients a server-minted session identity handle

DaniAkash:
## What

Every `/mcp` browser tool now carries an optional `session` string argument, giving an agent a server-minted identity for its own calls under the modern stateless transport. Same mechanism the claw-server uses, applied here for identity only.

- The server mints a UUID when the caller omits `session` and returns it in the result's `structuredContent`, so the agent can thread it back on later calls.
- A supplied `session` is echoed back unchanged.
- The handle is stripped from the arguments before the tool runs, so tool logic and validation are untouched, including the one tool that uses a strict schema.
- It is gated behind an opt-in registration flag, so only the `/mcp` surface exposes it. The internal AI-SDK agent shares the tool schemas and is unaffected.
- The handle is attributed in the per-tool execution metric.

Identity only: no isolation, no per-session state store, no change to browser behavior. Each agent still drives the full shared browser.

## Tests

- New browser-mcp test: the schema surfaces the optional `session`; a call without one mints and returns a handle (and the tool still runs with the key stripped); a supplied handle is echoed; the handle reaches the execution metric.
- Updated `/mcp` service tests: the ordinary tool envelope is still stripped by default with only the session handle added, and `run` retains its output plus the handle.

`tsc`, `biome check`, and the browser-mcp plus `/mcp` service, route, and register test files pass.

### DIFF
--- a/packages/browseros-agent/apps/server/src/api/services/mcp/mcp-server.ts
+++ b/packages/browseros-agent/apps/server/src/api/services/mcp/mcp-server.ts
@@ -45,6 +45,7 @@ export function createMcpServer(deps: McpServiceDeps) {
     instructions: MCP_INSTRUCTIONS,
     registration: {
       includeStructuredContent: deps.includeStructuredContent ?? false,
+      sessionIdentity: true,
       logger,
       onToolExecutionStart: () => deps.activity?.beginMcpToolExecution(),
       onToolExecutionEnd: () => deps.activity?.endMcpToolExecution(),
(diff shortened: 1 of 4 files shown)


DaniAkash:
Addressed in 8e451cdb9. The session handle no longer goes to the aggregated `tool_executed` event (whose rollup keeps only tool name, source, and success). It is now attributed on the per-call log context instead, which is the correct home for a per-session value and avoids putting an unbounded-cardinality UUID into an aggregated metric. The returned handle in `structuredContent` is unchanged.

PR #2306, opened 2026-08-14
fix(claw): reconcile historical live session status

felarof99:
## Summary

- Reconcile stored Live history rows against the authoritative connected-session registry in HTTP list and detail responses.
- Apply the Done filter after reconciliation while preserving storage cursors across empty filtered pages.
- Add an HTTP regression covering connected and disconnected rows, unchanged persistence, and cursor pagination.

## Design

The live registry remains authoritative for whether a historical row is currently Live. A stored Live row absent from that registry is presented as Done at the HTTP read-model boundary; storage classification and dedicated live APIs remain unchanged.

## Test plan

- [x] bunx bun@1.3.6 run check
- [x] bunx bun@1.3.6 run test
- [x] cargo test
- [x] cargo test --test routes historical_sessions_reconcile_stored_live_status -- --exact


felarof99:
@greptileai review

felarof99:
@claude review

PR #2344, opened 2026-08-17
fix(chat): surface the full upstream error body in the chat error card

DaniAkash:
## Problem
Chat uses the Vercel AI SDK `useChat`. When the gateway fails a turn (most visibly: credits exhausted, thrown by the worker in front of the default provider), the user only saw a short, generic sentence. The full error the server actually received was discarded, so there was no way to see what really happened.

## Root cause
The error-envelope pipeline already un-masks AI SDK stream errors and classifies them, but it built the envelope's `details` from the short constructed `error.message` and capped it at 500 characters. The gateway's real, actionable specifics live in the response body (OpenRouter-style gateways hide them in `metadata.raw`), which is captured on the custom fetch as `APICallError.responseBody` / `data.raw` but never forwarded. On top of that, the card only revealed `details` behind a toggle, and hid it entirely when it matched the message.

## Change
- Classifier: forward the full upstream response body (or structured `data.raw`) into the envelope `details`, pretty-printed as JSON. Secret redaction is unchanged; the raw body gets a larger-but-bounded cap so a pathological body cannot flood the card.
- Error card: render that detail always-expanded with a copy control instead of behind a toggle, so the real reason is visible without a click. The side panel and new tab share this component, so both surfaces are covered.

No wire or schema change: this reuses the envelope's existing `details` field, which was defined for exactly this.

## Tests
- Classifier: full response body is preserved and pretty-printed in `details`; falls back to `data.raw`; secrets stay redacted and the raw body is bounded.
- Card: the raw detail renders always-expanded (no toggle), and the full JSON is shown pretty-printed with a copy control.
- Type-check and lint clean across both packages.

## Follow-up
If credits exhaustion is ever returned as HTTP 200 with the error embedded in the SSE stream (rather than a 4xx the custom fetch already intercepts), the stream body would need inspecting too. The current 429 handling indicates the HTTP-error path, so that is a separate, verifiable follow-up.

### DIFF
--- a/packages/browseros-agent/apps/app/screens/sidepanel/index/ChatError.test.tsx
+++ b/packages/browseros-agent/apps/app/screens/sidepanel/index/ChatError.test.tsx
@@ -130,7 +130,7 @@ describe('ChatError envelope handling', () => {
     expect(html).toContain('Try again')
   })
 
-  it('offers a details disclosure when raw upstream text is present', () => {
+  it('always shows the raw upstream detail with no toggle', () => {
     const html = renderError(
       envelopeError({
         message: 'The model request failed.',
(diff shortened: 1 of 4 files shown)


DaniAkash:
#2308 ("fix(agent): propagate LLM errors to the chat UI with real detail") already lands the error-envelope pipeline that classifies gateway failures (credits exhausted, provider errors) into the chat error card, so what this PR set out to do is already on main. The extra always-expanded raw upstream JSON on top of that classified card is not worth the added UI noise. Closing as redundant.

PR #2415, opened 2026-08-21
feat(ai-settings): auto-discover models for OpenAI-compatible providers

### LINKED ISSUE #2411: [Feature Request] Auto-discover available models for OpenAI Compatible providers (fetch GET /v1/models instead of manual ID entry)
**What problem does this solve? (describe the problem you're experiencing)**

For all first-class providers (OpenAI, Anthropic, Google, OpenRouter, etc.), adding a provider in Settings → AI shows a searchable dropdown of models populated from the bundled catalog. But when using the **OpenAI Compatible** provider type, the Model field is a free-form text input and the user must know and paste the **exact model ID** from the provider's docs/UI.

This is a recurring friction point because OpenAI-compatible endpoints (vLLM, LM Studio, llama.cpp server, Ollama's `/v1` compatibility layer, OpenRouter, Together, Groq, etc.) each name models differently, and a typo or stale ID only surfaces later as a failed chat/test request. The capability gap is artificial: nearly every OpenAI-compatible server implements `GET {baseUrl}/models`, which returns the exact list of model IDs the endpoint serves.

**How are you working around this today? (your current solution or workaround)**

Copying model IDs by hand from the provider's documentation or a separate `curl {baseUrl}/models` call, then pasting them into the free-form Model field and re-testing the connection until the ID is accepted.

**What's your proposed solution? (how should BrowserOS address this?)**

Once the user has entered a Base URL (and API key if required) for an OpenAI Compatible provider, BrowserOS should query the endpoint's `GET /v1/models` and populate the model picker dropdown with the returned IDs — same UX as the curated providers, with manual entry kept as a fallback:

1. Server-side: add a route next to the existing `POST /test-provider` (e.g. `GET /models?baseUrl=...&apiKey=...`) in the agent server (`apps/server/src/api/routes/provider.ts`) so the user's API key is never exposed to a third party and CORS is handled in one place.
2. Client-side: in the add/edit provider dialog, once Base URL (+ key) are valid, fetch the list and feed it into the existing model picker (`getModelsForProvider` / `NewProviderDialog.tsx`), with a manual-refresh action and free-form override when the endpoint errors, returns an empty list, or doesn't implement `/models`.
3. Map `data[].id` (and `context_length` when present, e.g. LM Studio/Ollama) into `ModelInfo` so context-window and capability defaults stay consistent with catalog-based providers.
4. Optionally apply the same discovery to the **Ollama** and **LM Studio** provider types, which currently also rely on free-form entry but expose the same `/v1/models` endpoint.

**Additional context (optional)**

- Example response shape: `{"object": "list", "data": [{"id": "qwen/qwen3-coder-30b", "object": "model", ...}]}`
- The free-form entry path should remain available, since some gateways require model IDs that aren't listed (e.g. routed/aliased models).
- `@ai-sdk/openai-compatible` is already a server dependency; a plain `fetch` of `/models` in the server route is sufficient and provider-agnostic.

espetro:
Closes #2411

## Problem

For OpenAI-compatible endpoints (including Ollama and LM Studio), the Add Provider dialog showed a bundled models.dev catalog that is only ever a stale sample of what the user's endpoint actually serves. Users with local or proxied models had to paste exact model IDs by hand, and short catalogs (e.g. LM Studio's three entries) read as the complete list, leading users to conclude their loaded models were unsupported.

## Implementation

**Server (apps/server)**

- New `listProviderModels` in `lib/clients/llm/list-models.ts`: normalizes the base URL (strips trailing slash and trailing `/v1`), fetches `GET {base}/v1/models` with optional Bearer auth, 15s timeout, and maps `{ id, context_length }` entries. Non-string ids are filtered out.
- New route factory `createListModelRoutes` in `api/routes/provider.ts`, mounted at `POST /list-models` in `api/routes/index.ts` (separate from `/test-provider` to avoid the ugly `/test-provider/models` path).
- Soft-fail: on any upstream error the route returns 200 with `{ models: [], message }`, so the client always keeps its free-form model entry UX.

**Client (apps/app)**

- New `listModels` in `lib/llm-providers/listModels.ts`: posts to the agent server and maps results to `ModelInfo`, defaulting `contextLength` to 128000 when absent. Any failure returns an empty array (no throw).
- `NewProviderDialog`: a `useDiscoveredModels` hook auto-fetches (600ms debounce) whenever the provider type serves user-loaded models (`openai-compatible`, `ollama`, `lmstudio`) and a base URL is present. Discovered models replace the stale catalog in the picker; the free-form "paste exact ID" path is retained as fallback. A small refresh button next to the Model label re-triggers discovery. The "lists only common models" hint is suppressed once discovery returns results.
- Ollama and LM Studio get this for free (same `USER_LOADED_MODEL_PROVIDERS` set).

## Test plan

- `apps/server/src/lib/clients/llm/list-models.test.ts`: happy path, missing `data`, non-string ids, 401/500 throws, URL normalization (`/v1`, `/v1/`, trailing slash, bare host), Bearer header forwarding. All 15 tests pass.
- `apps/app/lib/llm-providers/listModels.test.ts`: mapping plus contextLength default, network failure returns `[]`, non-200 returns `[]`. Passes.
- `biome check` clean on all touched files; `tsc --noEmit` clean for touched files (pre-existing generated-graphql errors unchanged); full server and app suites show only pre-existing failures (verified identical on clean `main`).

## Out of scope

- Capability detection from `/models` metadata beyond `context_length` (not consistently present across providers).
- Persistent caching of discovered model lists.

### DIFF
--- a/packages/browseros-agent/apps/app/lib/llm-providers/listModels.test.ts
+++ b/packages/browseros-agent/apps/app/lib/llm-providers/listModels.test.ts
@@ -1,0 +1,65 @@
+import { afterEach, beforeEach, describe, expect, it } from 'bun:test'
+import { listModels } from './listModels'
+
+let originalFetch: typeof globalThis.fetch
+
+beforeEach(() => {
+  originalFetch = globalThis.fetch
+})
+
+afterEach(() => {
+  globalThis.fetch = originalFetch
+})
+
+function mockFetchJson(body: unknown, ok = true) {
+  globalThis.fetch = (async () =>
+    ({
+      ok,
+      json: async () => body,
+    }) as unknown as Response) as unknown as typeof globalThis.fetch
+}
+
+describe('listModels — success', () => {
+  it('maps modelId and contextLength', async () => {
+    mockFetchJson({
+      models: [{ modelId: 'llama3.2', contextLength: 131072 }],
+    })
+    const models = await listModels(
+      { type: 'ollama', baseUrl: 'http://localhost:11434' },
+      'http://127.0.0.1:9200',
+    )
+    expect(models).toEqual([{ modelId: 'llama3.2', contextLength: 131072 }])
+  })
+
+  it('defaults contextLength to 128000 when missing', async () => {
+    mockFetchJson({ models: [{ modelId: 'qwen2.5' }] })
+    const models = await listModels(
+      { type: 'lmstudio', baseUrl: 'http://localhost:1234' },
+      'http://127.0.0.1:9200',
+    )
+    expect(models).toEqual([{ modelId: 'qwen2.5', contextLength: 128000 }])
+  })
+})
+
+describe('listModels — soft failures return []', () => {
+  it('network failure → []', async () => {
+    globalThis.fetch = (async () => {
+      throw new TypeError('Failed to fetch')
+    }) as unknown as typeof globalThis.fetch
+
+    const models = await listModels(
+      { type: 'openai-compatible', baseUrl: 'http://x' },
+      'http://127.0.0.1:9200',
+    )
+    expect(models).toEqual([])
+  })
+
+  it('non-200 response → []', async () => {
+    mockFetchJson({ models: [{ modelId: 'm' }] }, false)
+    const models = await listModels(
+      { type: 'openai-compatible', baseUrl: 'http://x' },
+      'http://127.0.0.1:9200',
+    )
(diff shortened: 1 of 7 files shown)


espetro:

espetro:
@shadowfax92 looking forward to your hearing about your thoughts on enabling model auto-discovery for OpenAI-compatible providers

PR #2422, opened 2026-08-22
fix(cli): fall back to compatible status endpoint

tandede:
## What changed

The CLI `status` command now preserves the richer `/status` response when that endpoint exists, but falls back to BrowserOS-compatible health endpoints when `/status` returns HTTP 404.

Only a missing endpoint triggers the fallback. Authentication and server failures such as HTTP 500 are still returned to the caller instead of being hidden.

Regression tests cover the primary `/status` path, the `/system/health` compatibility path, and non-404 failures.

## Verification

- `go test ./...`
- `go vet ./...`
- `go build ./...`


tandede:

PR #2442, opened 2026-08-26
perf(deps): share dependencies across checkouts via the bun global store

DaniAkash:
> Draft. Types, lint and unit tests are green, and both extension bundles build, but this has found four distinct resolution problems so far and one bundling path is still unverified. See **Known cost** before merging.

## What

Turns on Bun's global virtual store, so each package is installed once into a shared location and symlinked in rather than copied per checkout.

```toml
[install]
linker = "isolated"
globalStore = true
```

**`node_modules` drops from 2098M to 55M per checkout, a 97% reduction.**

## What it took

The store places dependencies outside the project. Anything that resolves a sibling package **by name from its own location** stops working, because there is no longer a project tree above it to walk up into. That single cause produced four separate failures, each surfacing in a different layer.

### 1. React types

`lucide-react` resolves to the store, TypeScript follows the symlink to the real path, and `@types/react` is no longer reachable from there. `LucideProps extends Omit<SVGProps<SVGSVGElement>, …>` loses its base:

```
checkpoint.tsx(37,3): error TS2339: Property 'className' does not exist on type 'LucideProps'.
```

Seven errors, four of that shape plus three implicit-anys from the same failure in Radix and shadcn typings. Not an undeclared dependency: `@types/react` is declared in both apps. Fixed by pinning the specifier in `paths`, which applies program-wide including inside dependency `.d.ts` files.

Two alternatives were measured and rejected: `publicHoistPattern` hoists to the project root, which is precisely the location unreachable from a store path, and left all 7; `preserveSymlinks: true` made it far worse at 575 errors by changing resolution for workspace packages too.

`apps/claw-app` had no `paths` block and inherited one through `extends`. A `paths` block **replaces** rather than merges, so adding one dropped the eight generated wxt aliases and took the count to 473. They are restated there alongside the redirect.

### 2. The bun pin

The restated aliases needed `baseUrl` to resolve under the previously pinned Bun 1.3.6, but **TypeScript 7 removed `baseUrl`** (`TS5102`). Bun 1.3.14 resolves relative `paths` entries without it, so the pin is raised and the option dropped.

This also closes a real divergence: 1.3.6 ignores `globalStore` entirely, so CI was installing a materially different tree from local development, and green locally did not imply green in CI. The macOS builder keeps a literal pin because it has no workspace checkout to read the version file from, so it is raised in step.

### 3. Codegen plugins

`typecheck` depends on `codegen`, and codegen failed once the store was active in CI:

```
Unable to find template plugin matching 'schema-ast'
```

The plugin is not missing. `@graphql-codegen/schema-ast` is declared in `apps/app`, but the CLI resolves plugins relative to **itself**, and the store links the CLI in from outside the project. `pluginLoader` resolves them from the config file instead. The `client` preset was unaffected because presets resolve by a different path.

### 4. Phantom dependencies

The wxt dev server failed to start: esbuild could not resolve `lodash-es` from `@lobehub/icons`, nor `unist-util-visit` from `@lobehub/ui`. Both are genuine phantom dependencies, imported but never declared, in packages that are not ours to fix.

`trustedDependencies` keeps a package project-local rather than shared, so those two install as real directories and their undeclared imports resolve as before. Everything else still comes from the store. This costs `node_modules` 32M to 55M.

## Verification

Full typecheck in three configurations, each after wiping and reinstalling every `node_modules`, all producing an identical result:

| Configuration | Errors |
|---|---|
| before this change | 0 |
| store active, local development | 0 |
| store ignored, as CI ran before the bun bump | 0 |

A cold run, immediately after wiping `node_modules`, occasionally emits one spurious `TS2589` attributed to `@browseros/server`. It does not reproduce on a warm run (0 across three consecutive runs), does not reproduce when that package is typechecked on its own, and does not appear in CI. It looks like ordering or resource sensitivity in the parallel run rather than anything this change introduces.

Earlier revisions of this branch reported a differential of 39 findings. That number was an artifact: codegen had never succeeded on either side, so every one of them was a missing `@/generated/graphql/*`. With codegen working the real result is zero, and both extension bundles build: `apps/app` at 43.49 MB, `apps/claw-app` at 2.41 MB.

## Known cost

Four resolution problems, found reactively, each only when something exercised that path. Three were invisible to CI and the fourth needed a dev server to surface.

The uncomfortable part is the shape rather than the count: **phantom dependencies cannot be enumerated ahead of time.** Any dependency importing something it never declared will break, and it is only discoverable by running the code. Both wxt bundles were checked proactively; `build:server:test` could not be verified locally because it fails on missing production secrets, not on resolution, so that path remains unconfirmed.

Per Bun's documentation, packages with applied patches, entries in `trustedDependencies`, and workspace or file dependencies stay per-project by design, so native postinstall packages are unaffected. Tooling that scans `node_modules` without following symlinks may behave differently.

## Scope

Independent of #2446, which shares the cargo build directory. No overlapping files, either can land first.


DaniAkash:
Closing this as - vite + wxt is unable to support global store symlinks

PR #2449, opened 2026-08-27
feat(app-onboard): show ACP agents on the agent setup step

DaniAkash:
Three changes to the agent setup step.

## Coding agent row

Adds **OpenClaw** and **Hermes**, and a caption underneath: *or any other ACP compatible agent*. The row can only ever show a handful of what actually works, so the caption carries the breadth the icons cannot.

```
OR A CODING AGENT HARNESS YOU ALREADY USE
[Claude][Codex][Copilot][Gemini][Qwen][OpenClaw][H]
or any other ACP compatible agent
```

## Wording

The group title becomes **"Or a coding agent harness you already use"**, and the body copy above it follows. Harness is the accurate word here: what the user connects is the tool that drives a model, not the model itself. That distinction is exactly what this row is for, since the row above it is the models.

## Why the marks are inline SVG

The app renders Hermes from a PNG import. This app cannot do that. The Chromium target emits only `app.js`, `app.css` and the `icon/` directory enumerated in its `BUILD.gn`, enforced three ways: `assetsInlineLimit: 0`, an `assetFileNames` hook that throws on any non-CSS asset, and a `data:` URL guard in `verify-chromium-build`. A PNG would need a new entry in that `BUILD.gn` patch plus the verifier allowlist, which is a lot of machinery for one row icon.

So both marks are single-colour inline SVGs, matching the currentColor chips already in the row:

- **OpenClaw** is the silhouette of the mark the app ships. The eyes are cut out with `evenodd` rather than painted over so the shape survives any chip background, and the claws are kept clear of the body so it does not read as one blob at 23px.
- **Hermes** has no single-colour form (its mark is a detailed illustration), so its chip falls back to the same monogram the app already uses for Hermes in its agent picker.

## Verification

- `@browseros/app-onboard` tests: 6 pass, 0 fail. Two new ones cover the agent labels and the monogram fallback.
- `typecheck` and `lint` clean.
- `build:chromium` passes and `verify-chromium-build` reports exactly the allowlisted resources, so the inline SVGs add no new assets and no `data:` URLs.
- Rendered and checked visually on the real step.

## Follow-up worth considering

OpenClaw's mark here is derived from the one the app ships, which is itself a stand-in rather than an official logo. If a real vector mark turns up for either agent, both are a one-component swap.

### DIFF
--- a/packages/browseros-agent/apps/app-onboard/src/onboarding/components/AgentBrandMarks.tsx
+++ b/packages/browseros-agent/apps/app-onboard/src/onboarding/components/AgentBrandMarks.tsx
@@ -1,0 +1,53 @@
+/**
+ * Marks for agents that @lobehub/icons does not carry.
+ *
+ * These are inline single-colour SVGs rather than imported image files: the
+ * Chromium build emits exactly app.js, app.css and the icon/ directory listed
+ * in that target's BUILD.gn, so any Vite-processed asset either throws during
+ * the build or trips the data: URL guard in verify-chromium-build.
+ */
+
+interface MarkProps {
+  size?: number
+}
+
+/**
+ * OpenClaw. Silhouette of the mark the app ships, flattened to one colour.
+ *
+ * The eyes are cut out of the body with evenodd rather than painted over, so
+ * the shape stays legible against any chip background, and the claws are kept
+ * clear of the body so the outline does not read as one blob at chip size.
+ */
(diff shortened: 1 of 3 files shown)


DaniAkash:
Landed directly on `epic/app-onboard` instead. The branch was a single commit on top of the epic tip, so the epic fast-forwarded to `5152c1e54` with no merge commit and the change is identical. Closing this in favour of that.

PR #2469, opened 2026-08-30
fix(release): bound macOS notarization waits

shadowfax92:
## Summary
- Bound every macOS app and DMG `notarytool submit --wait` call to two hours.
- Apply the same policy to keychain-profile and direct-credential fallback paths.
- Add command-contract tests for all four submission paths.

## Design
A small shared notarization policy owns the client-side wait bound. Apple continues processing after the client exits, but the serialized Mac runner now fails with a diagnosable notarization error instead of silently consuming the workflow's 20-hour limit. This does not alter the already-running 0.50.0 job because that run is frozen to its original source SHA.

## Test plan
- [x] `uv run python -m unittest discover -s . -p '*_test.py'` (1,189 passed, 2 skipped)
- [x] Confirm local `notarytool submit --help` supports `--wait --timeout <duration>` and leaves server-side processing running after client timeout.

### DIFF
--- a/packages/browseros/bos_build/lib/notarization.py
+++ b/packages/browseros/bos_build/lib/notarization.py
@@ -1,0 +1,14 @@
+"""Shared client-side wait policy for Apple notarization submissions."""
+
+NOTARYTOOL_WAIT_TIMEOUT = "2h"
+
+
+def notarytool_wait_args() -> list[str]:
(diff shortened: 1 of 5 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #2478, opened 2026-08-31
fix: repair tracked appcast CDN metadata

shadowfax92:
## Summary
- add a dry-run-by-default workflow that republishes selected tracked browser/server appcasts through validation, backup, and monotonic version guards
- support the registered BrowserClaw title migration for an existing empty placeholder without allowing versionless creation or release erasure
- document and test the repair path, including malformed-shell fail-closed cases

## Design
The workflow checks out the repository default branch, exposes only registered appcast choices, shares the server appcast writer concurrency group, and deliberately offers no downgrade override. FeedPublisher recognizes a versionless placeholder only when both tracked and live XML are the exact empty RSS 2.0 shell; populated or malformed live objects remain on the strict version/repair path.

## Test plan
- [x] uv run python -m unittest discover -s bos_build -t . -p '*_test.py' (1,193 passed, 2 skipped)
- [x] uv run ruff check bos_build
- [x] actionlint on repair-update-feed.yml and bos-build-tests.yml
- [x] git diff --check
- [x] preflight freshly downloaded live appcast-claw-win-arm64.xml and appcast-server.xml against tracked bytes with the committed publisher code

### DIFF
--- a/.github/workflows/bos-build-tests.yml
+++ b/.github/workflows/bos-build-tests.yml
@@ -18,6 +18,7 @@ on:
       - ".github/workflows/nightly.yml"
       - ".github/workflows/nightly-macos-product.yml"
       - ".github/workflows/publish-server-ota.yml"
+      - ".github/workflows/repair-update-feed.yml"
       - ".github/workflows/release-browseros.yml"
       - ".github/workflows/release-browserclaw.yml"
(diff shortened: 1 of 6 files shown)


shadowfax92:
@claude review

shadowfax92:
@greptileai review

PR #2517, opened 2026-09-02
fix(server): steer ACP agents to browseros, not a co-installed browseros-neo

DaniAkash:
## Problem

The server spawns ACP agents (Claude / Codex) to browse and hands them the local `browseros` MCP toolset to use. When the separate `browseros-neo` product is also installed on the same machine, those agents keep using `browseros-neo` instead: it installs its skill and MCP server into the user's global agent config (`~/.claude/skills`, `~/.claude.json`, `~/.codex/config.toml`, ...), which the spawned agents inherit, and its skill instructs agents to default to it and not fall back.

## Fix

Steer the agents at the instruction level. The `browseros` skill that the server already appends to Claude's system prompt (`sessionOptions.systemPrompt.append`) and injects as Codex's `developer_instructions` now tells the agent to use only the `browseros` MCP server and to ignore a co-installed `browseros-neo` skill or MCP, while continuing to use the rest of the user's own tools and skills.

This is deliberately an instruction-level change, not config isolation. Redirecting the agent's config directory would also strip the user's own MCP servers and skills, which defeats the point of bringing your own Claude Code / Codex. `browseros-neo` sits in the same shared config as the user's own entries, so the exclusion has to be by identity (name), not by directory. The system prompt / developer-instructions channel targets only `browseros-neo` and leaves everything else intact.

## Scope

- One file: `apps/server/resources/skills/browseros/SKILL.md`, which is the single source injected into both agents.
- No change to config directories, cwd, or MCP wiring, so the user's own configuration is untouched.

## Verification

Loader validation and skill-staging build check pass; frontmatter and existing guidance are unchanged. Manual check to follow: run a server-driven browse task on Claude and Codex with `browseros-neo` installed and confirm the agent uses `browseros` and never calls `browseros-neo`.


DaniAkash:
@greptileai

PR #2530, opened 2026-09-02
feat: unify product installation identity

shadowfax92:
## Summary
- persist one installation ID per product in `~/.browseros/installation.json` or `~/.browserclaw/installation.json`
- share that identity across native Chromium metrics and the BrowserOS/BrowserClaw sidecars without profile-scoped prefs
- route native analytics by the runtime product and hide bookmark-bar tab groups by default in both products

## Design
Product-state root resolution is centralized at each process boundary, and every metrics surface reads the same minimal installation document. First-run creation uses an exclusive atomic publish so concurrent Chromium and sidecar startups converge on one UUID. Chromium owns metrics through a process-scoped service; sidecar config identity fields remain empty and are ignored while the product-specific PostHog tokens stay build/runtime selected.

## Test plan
- [x] `autoninja -C out/Default_browseros_arm64 chrome`
- [x] BrowserOS server focused tests, full server library suite, and typecheck
- [x] BrowserClaw analytics and telemetry tests plus clippy with warnings denied
- [x] `go test ./...` for `tools/dev`
- [x] `bun run check`
- [x] Confirm the broad workspace test failures are confined to eight unchanged app tests with the existing `@/...` resolver issue


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #2531, opened 2026-09-02
feat: agents never steal focus in BrowserClaw

shadowfax92:
## Summary
- New profile pref `browseros.automation_never_steals_focus` (default on for BrowserClaw, off for BrowserOS). A tab with a DevTools client attached can no longer switch the user's active tab or raise the window: `Browser::ActivateContents` becomes a no-op for it, and tabs or popups its pages open after an agent click land in the background (`Browser::AddNewContents`, mirroring upstream's actor gate).
- `Browser.createTab` defaults to background; an explicit `background:false` only selects the tab within its window. Under the pref, `activateTab` stops raising the window, `activateWindow` is a no-op, and `createWindow` / `setWindowVisibility(activate)` show windows inactive.
- The MCP tools drop `background` from `tabs new` and `browser.pages.newPage()`: agents always open pages in the background. The field is still accepted and ignored so clients holding the old schema keep working.

## Design
Production traces showed two systematic focus stealers: agents passing `background:false` (14 of 63 `tabs new` calls), and page-opened tabs and popups triggered by agent clicks, which count as trusted gestures. Every path funnels through two cross-platform Chromium functions plus the BrowserOS DevTools handler, so the gates live there and work identically on macOS, Linux and Windows. "Agent-driven" is approximated by `DevToolsAgentHost::IsDebuggerAttached`; claw-server sessions never detach, so this stays true for any tab an agent touched this browser session, which is acceptable because every such tab was opened by an agent (recorded in a comment at the gate). Full investigation: `.llm/260902-0807_agent_focus_steal/findings.html`.

## Test plan
- [x] `autoninja -C out/Default_browseros_arm64 chrome` green on the task branch (`task/agent_focus_steal` in chromium-3)
- [x] Patches byte-match `git diff BASE_COMMIT..tip --full-index` and apply cleanly onto the bare base tree (5/5)
- [x] `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test -p browseros-mcp`, `cargo test -p claw-server-rust`
- [ ] Manual, with a BrowserClaw build: `tabs new` never switches the active tab; an agent click on a `target=_blank` link opens a background tab; a popup from an agent tab does not raise the app; Ctrl+T and cockpit Watch unchanged
- [ ] `bun run test:claw-mcp-contract` against the new build (conformance cases updated; they need a local browser)

https://claude.ai/code/session_01TwxGoD3ZW8cVTwhyi4z3Q6

### DIFF
--- a/packages/browseros-agent/contracts/claw-mcp/tests/cases-snapshot-concurrency.ts
+++ b/packages/browseros-agent/contracts/claw-mcp/tests/cases-snapshot-concurrency.ts
@@ -452,7 +452,6 @@ export const snapshotConcurrencyCases: ContractCase[] = [
         const result = await ctx.mcp.callTool('tabs', {
           action: 'new',
           url: mixedFrameUrl(ctx),
-          background: false,
         })
         const text = expectOk(result, 'tabs new mixed-frame auto-context')
         page = parsePageId(result)
--- a/packages/browseros-agent/contracts/claw-mcp/tests/cases-tabs.ts
+++ b/packages/browseros-agent/contracts/claw-mcp/tests/cases-tabs.ts
@@ -21,13 +21,12 @@ function parseGroupId(text: string): string {
 
 export const tabsCases: ContractCase[] = [
   {
-    name: 'tabs: new foreground page opens with auto-context',
+    name: 'tabs: new page opens with auto-context',
     smoke: true,
     async run(ctx) {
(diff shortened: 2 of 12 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #2561, opened 2026-09-05
chore(app): remove sync data popup

shadowfax92:
Remove the “Sync your data” popup from the home page, including its sign-in trigger, dismissal storage definition, and component. The remaining import prompt now uses an import-only hook with the same cooldown, delay, and cleanup.

Validation: root `bun run check`, full `bun run test`, production build, and independent review. Browser smoke confirms that a signed-out profile with import dismissed shows no sync popup, including after reload with an old sync-dismissal value. The eligible import prompt still appears, and home text input works.


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #2559, opened 2026-09-05
fix(onboarding): wait for native setup with recoverable retry

shadowfax92:
Both browser-owned onboarding WebUIs now show **Finishing setup** immediately after Skip or a final action. Native setup failures show **Setup didn’t finish** and a primary **Retry** button, using the existing orange BrowserOS and blue BrowserOS neo themes.

The API v1 bridge gains optional `setupState` (`idle`, `preparing`, `failed`, `ready`) independently of importer status. `browserosOnboardingComplete` stays unchanged and becomes idempotent; `browserosOnboardingRetrySetup` is available only after native setup failure. Receiver registration and optimistic state updates precede native messages, preserving synchronous replies and restored state after reload. BrowserClaw’s renderer navigation is removed so Chromium owns the readiness handoff. Native installation/readiness work is a separate coordinated PR.

Validation:
- Both onboarding apps: focused tests, typechecks, and `bun run build:chromium` resource verification passed.
- Headless Chrome against both built bundles: all five completion exits, duplicate actions, failure/Retry, synchronous native replies, and restored preparing/failed/ready states passed with no renderer navigation.
- Root `bun run check` passed. The full repository test run passed every suite except the extension app’s existing Bun 1.3.6 inherited `@/*` alias resolution issue; all 64 extension test files on the final branch passed in isolated processes using an explicit scratch tsconfig with the same app path mappings. No extension app source/config changes are included.
- Independent adversarial review: no findings. Native worker confirmed matching protocol names.

PR #2563, opened 2026-09-05
fix: wait for bundled extensions before first-run handoff

shadowfax92:
A fast first-run Skip could open the product extension before Chromium finished loading it. Bundle discovery also waited on an unbounded network request when local metadata was missing, and three overlapping installation paths could race.

This change makes discovery local-only, gives bundled file installs transient USER_VISIBLE priority, and uses one coordinator with Chromium's external provider as the install executor. Native onboarding waits for the target profile's primary extension registry READY event; completion and Retry callers join the same operation. Installed same/newer versions are retained, valid newer bundles install first, and bounded recovery/maintenance uses the provider's update interface. Periodic cleanup, safe re-enabling, and health reporting remain independent of network success.

The optional API v1 `setupState` field and `browserosOnboardingRetrySetup` message match the independently released onboarding UI in #2559. Import status remains separate, and setup state survives WebUI reloads. A native navigation guard keeps legacy BrowserClaw resources on onboarding; native opens its MCP destination after READY.

Validation:
- Chromium chrome build through the shared sfmux build slot.
- Context-free code review, including pending URL retries, secondary bundle failures, reloads, and shutdown.
- Extracted patches checked against the task branch and applied to the Chromium base.
- Chromium unit/browser tests omitted per repository instructions.

### DIFF
--- a/packages/browseros/chromium_patches/.features.yaml
+++ b/packages/browseros/chromium_patches/.features.yaml
@@ -481,3 +481,14 @@ features:
     description: "feat: macOS passkey entitlements"
     files:
       - chrome/app/app-entitlements-browseros.plist
+
+  extension-startup-readiness:
+    description: "feat: bundled extension startup readiness"
+    files:
+      - chrome/browser/chrome_content_browser_client_navigation_throttles.cc
+      - chrome/browser/extensions/external_provider_impl.h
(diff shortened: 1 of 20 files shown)


shadowfax92:
@greptileai review

shadowfax92:
@claude review

PR #2582, opened 2026-09-07
fix(claw-app): stop showing a guessed harness next to the agent name

DaniAkash:
## The bug

A running Codex session renders as **`CODEX  CLAUDE CODE`** in the Running now card. Two fields sit side by side and only the first one is real:

```tsx
<span className="text-white">{session.label}</span>       // CODEX        announced
<span className="text-white/45">{session.harness}</span>  // CLAUDE CODE  invented
```

## Why it says Claude Code

```ts
/** Coerces contract strings into the UI harness union with an honest fallback. */
export function harnessForRow(value: string | undefined): Harness {
  if (!value) return 'Claude Code'
  return HARNESSES.includes(value) ? value : 'Claude Code'
}
```

`harness` is only ever populated from a matched agent profile:

```rust
harness: profile.map(|profile| profile.harness.to_string()),   // cockpit/query.rs
```

An agent with no configured profile resolves to `ClientIdentity::Ephemeral`, so `matched_profile` returns `None`, the field never reaches the client, and the fallback fires. There is a second path to the same wrong answer: `HARNESSES` requires an exact match against seven strings, so an announced name like `codex-mcp-client` misses and also lands on `'Claude Code'`.

The doc comment claims an honest fallback. Hard-coding one vendor's product name to mean "unknown" is the opposite, and a unit test pinned exactly that: `expect(harnessForRow('Atlas-9000')).toBe('Claude Code')`.

## The replay subtitle, same shape, different cause

```ts
harness: task.profileId ?? 'unknown',
```

That renders a raw profile id, or the literal word `unknown`, under the label of a harness. Also removed.

## What changed

Both display sites are gone, along with the now-unused coercion, type fields, and fixtures. Nothing computes or renders an inferred harness any more.

**Kept deliberately:** the `/mcp` connect screen and the cockpit onboarding chips still list harnesses. That list is a real user choice about which harness configs are connected, not an inference from session identity, so it stays exactly as it is.

## Verification

`tsc --noEmit` clean, 387 tests pass, biome clean on the changed files. The only removed assertions were the ones pinning the invented value.

## Note

This is the display-side sibling of the identity work in #2581. Same underlying habit: guessing an agent's identity from a small closed set instead of using what the agent actually announced. The two changes are independent and touch different files.

### DIFF
--- a/packages/browseros-agent/apps/claw-app/components/cockpit/AgentRunningCard.tsx
+++ b/packages/browseros-agent/apps/claw-app/components/cockpit/AgentRunningCard.tsx
@@ -94,7 +94,6 @@ export function AgentRunningCard({
               style={{ background: session.color }}
             />
             <span className="truncate text-white">{session.label}</span>
-            <span className="shrink-0 text-white/45">{session.harness}</span>
           </span>
           {showingLive ? (
             <span className="inline-flex shrink-0 items-center gap-1.5 text-[#8fb4ff]">
--- a/packages/browseros-agent/apps/claw-app/components/cockpit/RunningGrid.test.tsx
+++ b/packages/browseros-agent/apps/claw-app/components/cockpit/RunningGrid.test.tsx
@@ -86,7 +86,6 @@ function session(
     slug: 'codex',
     label: 'Codex',
     name: 'Research BrowserClaw',
-    harness: 'Codex',
     color: '#0254ec',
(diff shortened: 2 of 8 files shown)


DaniAkash:
Folded into #2581 as a commit on that branch, so the identity work and this display fix land together. Closing here.

PR #2583, opened 2026-09-07
fix(mcp): enforce the evaluate timeout instead of only forwarding it to CDP

### LINKED ISSUE #2039: [Bug] evaluate timeout parameter is accepted but not enforced
## Issue Type

Agent Issue

## Operating System

Both / server-side behavior

## Description of the bug

The `evaluate` MCP tool accepts a `timeout` argument, but the current real-browser contract test explicitly records that renderer evaluation is not preempted by it:

https://github.com/browseros-ai/BrowserOS/blob/5ac300052b7b890a7c09a2800456fbcd252a0dd8/packages/browseros-agent/contracts/claw-mcp/tests/cases-read-eval.ts#L314-L329

The Rust browser core sends `Runtime.evaluate` with `awaitPromise: true` but does not pass or otherwise enforce a timeout:

https://github.com/browseros-ai/BrowserOS/blob/5ac300052b7b890a7c09a2800456fbcd252a0dd8/packages/browseros-agent/crates/browseros-core/src/browser.rs#L114-L143

As a result, a page-context promise can run beyond the caller-provided limit. This behavior was also identified by the merged contract-suite work in #1899. I have not rerun the gated `BROWSEROS_BINARY` suite locally; the report is based on the current checked-in contract and implementation at `5ac300052b7b890a7c09a2800456fbcd252a0dd8`.

## Steps to Reproduce

1. Call `evaluate` with code that waits for about eight seconds before returning.
2. Set `timeout` to `1500` milliseconds.
3. Observe that the call succeeds after the page-context wait instead of timing out near 1.5 seconds.

The checked-in contract contains this exact case.

## Expected behavior

- A fast evaluation below the limit succeeds normally.
- An evaluation exceeding the requested timeout returns a recognizable timeout error near the configured deadline.
- The MCP session remains usable after the timeout.
- The real-browser contract is updated to assert timeout enforcement rather than the current limitation.

## BrowserOS Version

Current `main` at `5ac300052b7b890a7c09a2800456fbcd252a0dd8`.

## Additional Context

Searches across open and closed issues/PRs found the limitation documented in #1899 but no dedicated issue or active PR owning the fix. If the expected timeout semantics are confirmed, I would be happy to prepare a focused PR.

avinashgola:
Fixes #2039

## The bug

`evaluate` accepts a `timeout` and the tool does forward it to CDP. But CDP's `Runtime.evaluate` timeout only terminates **synchronous** execution — with `awaitPromise: true`, a promise parked on a macrotask leaves an empty JS stack, so there is nothing for CDP to terminate and the call runs to completion.

The checked-in contract records exactly this today:

```ts
// Renderer evaluation is not preempted by this request timeout today.
```

So `timeout: 1500` against `await new Promise(r => setTimeout(r, 8000))` still returns after the full eight seconds.

### One correction to the issue

#2039 points at `browseros-core/src/browser.rs::evaluate` as the code path. That function has no callers in the tree — `run`'s `browser.evaluate()` dispatches to `tool:evaluate` (`run.rs:164`), the same MCP tool — so the fix belongs in `crates/browseros-mcp/src/tools/evaluate.rs`, where it covers both entry points. I left `browser.rs` alone rather than change an unused public signature; happy to include it if you would rather it not drift.

## The fix

Bound the request as well as the renderer:

- the CDP `timeout` stays, so a synchronous overrun is still terminated and reported by CDP itself;
- `tokio::time::timeout` wraps the request at `timeout + 250ms`, so this guard only catches what CDP cannot preempt;
- cancellation is honoured through the same `tokio::select!` idiom `response.rs` already uses;
- on elapse the tool returns `evaluate: timed out after {timeout}ms`, matching `wait`'s existing phrasing.

The session stays usable afterwards: dropping the request leaves its entry in the CDP client's pending map, and `client.rs:405` removes it and discards the send when the late reply arrives.

## Tests

- New harness test `evaluate_times_out_instead_of_waiting_out_the_page`: asserts the timeout error and that the next evaluation on the same session still succeeds. `HarnessConnection` gains a `stall_evaluate` knob so `Runtime.evaluate` can withhold its reply the way an awaited promise does.
- Unit test for the deadline arithmetic.
- The real-browser contract case now asserts enforcement (error, near the deadline, session survives) instead of recording the limitation.

I verified the new test fails without the fix — it waits out the full stall and returns a **success** result — and passes with it.

```
cargo test -p browseros-mcp                              79 passed, 0 failed
cargo clippy --workspace --all-targets -- -D warnings    clean
cargo fmt --all --check                                  clean
bunx tsc --noEmit                                        clean
bunx @biomejs/biome check <contract file>                clean
```

**Not run:** the gated `BROWSEROS_BINARY` contract suite. I do not have a BrowserOS build locally, so the updated contract case is unverified against a real renderer and would be worth a CI run before merge.

### DIFF
--- a/packages/browseros-agent/contracts/claw-mcp/tests/cases-read-eval.ts
+++ b/packages/browseros-agent/contracts/claw-mcp/tests/cases-read-eval.ts
@@ -311,7 +311,7 @@ export const readEvalCases: ContractCase[] = [
     },
   },
   {
-    name: 'evaluate: accepts a timeout parameter',
+    name: 'evaluate: enforces the timeout parameter',
     async run(ctx) {
       const page = await ctx.openPage(ctx.fixture('/form.html'))
       // A fast evaluation under the accepted timeout returns normally.
@@ -319,13 +319,28 @@ export const readEvalCases: ContractCase[] = [
       if (!fast.includes('2')) {
         throw new Error(`evaluate with a timeout did not return: ${fast}`)
       }
-      // Renderer evaluation is not preempted by this request timeout today.
+      // A page-context wait past the deadline is cut short, not awaited.
+      const started = Date.now()
       const slow = await ctx.mcp.callTool('evaluate', {
         page,
         code: 'await new Promise(r => setTimeout(r, 8000)); return "slow"',
         timeout: 1_500,
       })
-      expectOk(slow, 'evaluate page-context wait')
+      const elapsed = Date.now() - started
+      const text = payload(textOf(slow))
+      if (!slow.isError || !text.includes('timed out after 1500ms')) {
+        throw new Error(
+          `evaluate did not time out: isError=${slow.isError} text=${text}`,
+        )
+      }
+      if (elapsed > 5_000) {
+        throw new Error(`evaluate waited ${elapsed}ms for a 1500ms timeout`)
+      }
+      // The session stays usable once the abandoned wait finally resolves.
+      const after = await evalIn(ctx, page, 'return 6 + 1', 5_000)
+      if (!after.includes('7')) {
+        throw new Error(`session unusable after a timeout: ${after}`)
+      }
     },
   },
   {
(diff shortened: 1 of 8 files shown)


avinashgola:

DaniAkash:
@greptileai

PR #2584, opened 2026-09-07
docs: correct the contributor setup instructions

avinashgola:
A new contributor following `CONTRIBUTING.md` cannot reach a working build today — every command in the agent setup path is wrong for the current repo.

## CONTRIBUTING.md

| The doc says | What the repo does |
| --- | --- |
| `yarn install` | `packages/browseros-agent/package.json` pins `bun@1.3.6` and sets `engines` to `"yarn": "please-use-bun"` (same for npm and pnpm). There is no `yarn.lock`, only `bun.lock`. |
| `cp .env.example .env` | No `.env.example` exists. The tracked templates are `.env.development.example` and `.env.production.example`; the old per-app env files were merged into root ones. |
| `yarn build:dev` | There is no `build:dev` script at the workspace root. Setup is `bun run dev:setup` then `bun run dev:watch`. |
| Load unpacked from `packages/browseros-agent/dist/` | `apps/app/wxt.config.ts` sets `outDir: 'dist'` and WXT suffixes the mode, so a dev build lands in `apps/app/dist/chrome-mv3-dev`. |
| Links to `packages/browseros-agent/CONTRIBUTING.md` | That file does not exist. Now points at the package `README.md` and `CLAUDE.md`. |
| "Node.js 18+, ~500MB disk, 10 minutes" | Bun, plus **Go** — `tools/dev/run.sh` exits with an install prompt when Go is missing, so `dev:setup` and `dev:watch` both fail without it — plus Rust for `crates/`. |
| "Testing: Vitest (not Jest)" | Tests run under `bun test` (`apps/app` test script shells out to `scripts/run-bun-test.ts`, and the agent `CLAUDE.md` says `bun test`). |
| Structure tree with `src/lib`, `src/sidepanel`, `src/newtab`, `src/background` | The agent workspace is `apps/`, `crates/` and `packages/`. Also `packages/browseros/build/` is really `bos_build/`. |

Two further changes:

- **Code standards now link to the per-package `CLAUDE.md` files** instead of restating them. The hand-copied copy had drifted (Vitest, and Zod/alias rules that aren't in the ground rules) while the real rules — extensionless imports, kebab-case, `@browseros/shared` constants — were missing entirely. Restating them here is what let them rot; the highlights that remain are corrected against the source.
- **Browser setup** gained the `uv sync` and `cp .env.example .env` steps that `bos_build/README.md` documents, plus `--show-plan`. The existing `browseros build` invocations were correct and are unchanged.

## docs/contributing.mdx

The published page carries the same class of errors:

- `apps/controller-ext` does not exist — `apps/` is app, app-onboard, claw-app, claw-onboard, claw-server-rust, cli, server. Listed `apps/cli` instead.
- Same per-app `.env.example` copies, same bare `dist/` load-unpacked path.
- `bun run build:ext` is not a script; row removed.
- Vitest → `bun test`.

## Verification

I ran the documented paths rather than only reading them:

- `bun install` → `bun run codegen:agent` → `bun run build:agent:dev` produces `apps/app/dist/chrome-mv3-dev`. Worth noting: `build:agent:dev` alone **fails** on a fresh clone (`Could not load apps/app/generated/graphql/gql`), so the doc names the codegen step explicitly.
- `uv sync` then `uv run browseros build --list` and `--help` both work, confirming the browser commands and the `--chromium-src` / `--provision` / `--show-plan` flags.
- Every relative link and every repo path referenced in the rewritten file resolves.
- `git diff --check` is clean.

No code changes; docs only.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -3,8 +3,9 @@
 Hey there! Thanks for your interest in BrowserOS. Whether you're fixing bugs, adding features, improving docs, or just poking around the code, we're glad you're here.
 
 BrowserOS is a monorepo with two main parts:
-- **Agent** - The Chrome extension with AI features (TypeScript/React)
-- **Browser** - The custom Chromium build (C++/Python)
+
+- **Agent** — The agent platform: extension UI (TypeScript/React), server, MCP tools (Rust), and a CLI (Go)
+- **Browser** — The custom Chromium build (C++/Python)
 
 Most folks start with the agent since it's way easier to set up and iterate on.
 
@@ -23,11 +24,11 @@ Most folks start with the agent since it's way easier to set up and iterate on.
 - Testing & docs
 
 **What you need:**
-- Node.js 18+
-- ~500MB disk space
-- 10 minutes to set up
+- [Bun](https://bun.sh) — the version pinned in `packages/browseros-agent/package.json`
+- [Go](https://go.dev/dl/) — required by the `dev:*` scripts
+- [Rust](https://rustup.rs/) — only if you touch `crates/`
 
-**Skills:** TypeScript, React, Chrome APIs
+**Skills:** TypeScript, React, Rust, Go, Chrome APIs
 
 **[→ Agent Setup](#agent-development)**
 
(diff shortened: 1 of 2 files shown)


DaniAkash:
already fixed in https://github.com/browseros-ai/BrowserOS/pull/2666

PR #2585, opened 2026-09-08
fix(mcp): stop emitting null inside nullable enum tool schemas

### LINKED ISSUE #2479: MCP Tool Schema Pydantic Validation Failure (enum.3)
### Issue Type

Both / Not sure

### Operating System

macOS

### Description of the bug

Bug Report: The MCP client fails during tool schema decoding when encountering a null (None) value inside a string enumeration array (enum.3) provided by the BrowserOS neo MCP server. Pydantic v2 strict typing expects all enum elements to be valid strings, resulting in a ValidationError.

dcode version : v0.1.65

langchain deepagent dcode 
"
Error: Agent error: ValidationError: 1 validation error for Schema
enum.3
  Input should be a valid string [type=string_type, input_value=None, input_type=NoneType]
    For further information visit https://errors.pydantic.dev/2.13/v/string_type
"

### Steps to Reproduce

1- set up browseros-neo.
"
    "browseros-neo": {
      "type": "http",
      "url": "http://127.0.0.1:9010/mcp"
    },
"
2- run  'dcode '. langraph deepagent dcode version : v0.1.65
3- dcode tui result error on any prompt executed:
"
Error: Agent error: ValidationError: 1 validation error for Schema
enum.3
  Input should be a valid string [type=string_type, input_value=None, input_type=NoneType]
    For further information visit https://errors.pydantic.dev/2.13/v/string_type
"

### Screenshots / Videos

Image

### BrowserOS Version

BrowserOS - 0.49.5.0

### Additional Context

works on most  other harnesses; opencode, vscode, antigravity, custom harness.

avinashgola:
Fixes #2479

## The bug

schemars renders an `Option<SomeEnum>` tool argument as:

```json
{ "type": ["string", "null"], "enum": ["left", "middle", "right", null] }
```

A Pydantic-backed MCP client decodes every `enum` entry against the declared type, so the trailing `null` fails the whole tool list before any prompt runs — exactly the reported error, where `enum.3` is that `null`:

```
ValidationError: 1 validation error for Schema
enum.3
  Input should be a valid string [type=string_type, input_value=None, input_type=NoneType]
```

Three arguments in the catalog carry one today. I found them by walking every generated schema:

```
tab_groups/properties/color/enum = ["grey","blue","red","yellow","green","pink","purple","cyan","orange",null]
act/properties/button/enum       = ["left","middle","right",null]
act/properties/direction/enum    = ["up","down","left","right",null]
```

`act.button` is the reporter's `enum.3`. This also explains why the report says other harnesses work — a client that does not strictly type-check `enum` entries never notices.

## The fix

`framework.rs` already normalizes generated schemas into "the object-only form expected by strict MCP clients" — rewriting boolean schemas, stripping boolean defaults. The nullable enum belongs in that same pass, so `collapse_nullable_enum` drops the `null` from `enum` and collapses `["string","null"]` back to `"string"`.

Two reasons this is safe rather than a semantic loss:

- **Optionality is already carried by `required`.** `act.button` is absent from `required` both before and after; the added test asserts it.
- **The server never validates against this schema.** `parse_args` deserializes with serde, and there is no JSON-Schema validator in the crate. `Option<Button>` still accepts an explicit `null` at runtime. The change alters what clients are *told*, not what the server accepts.

An enum consisting only of `null` is left alone rather than emptied.

## Tests

- `tool_schemas_never_put_null_inside_an_enum` walks every input and output schema in the catalog, so a future `Option<Enum>` argument fails CI instead of shipping.
- `optional_enum_arguments_keep_their_variants_and_stay_optional` pins `act.button` to its three variants, a plain `"string"` type, and absence from `required`.

Both fail on `main` and pass with the fix:

```
test tests::optional_enum_arguments_keep_their_variants_and_stay_optional ... FAILED
  left: Some(Array [String("left"), String("middle"), String("right"), Null])
 right: Some(Array [String("left"), String("middle"), String("right")])
test tests::tool_schemas_never_put_null_inside_an_enum ... FAILED
```

```
cargo test -p browseros-mcp                              79 passed, 0 failed
cargo clippy --workspace --all-targets -- -D warnings    clean
cargo fmt --all --check                                  clean
```

**Not run:** the gated `BROWSEROS_BINARY` contract suite — no local BrowserOS build. The change is confined to schema generation and is covered by the unit tests above.

### DIFF
--- a/packages/browseros-agent/crates/browseros-mcp/src/framework.rs
+++ b/packages/browseros-agent/crates/browseros-mcp/src/framework.rs
@@ -509,6 +509,13 @@ fn collapse_nullable_enum(object: &mut JsonObject) {
     if !values.iter().any(Value::is_null) {
         return;
     }
+    let Some(Value::Array(types)) = object.get("type") else {
+        return;
+    };
+    if !types.iter().any(|entry| entry.as_str() == Some("null")) {
+        return;
+    }
+
     let kept_values: Vec<Value> = values
         .iter()
         .filter(|value| !value.is_null())
@@ -518,11 +525,7 @@ fn collapse_nullable_enum(object: &mut JsonObject) {
     if kept_values.is_empty() {
         return;
     }
-    object.insert("enum".to_string(), Value::Array(kept_values));
 
-    let Some(Value::Array(types)) = object.get("type") else {
-        return;
-    };
     let mut kept_types: Vec<Value> = types
         .iter()
         .filter(|entry| entry.as_str() != Some("null"))
@@ -533,6 +536,7 @@ fn collapse_nullable_enum(object: &mut JsonObject) {
         1 => kept_types.remove(0),
         _ => Value::Array(kept_types),
     };
+    object.insert("enum".to_string(), Value::Array(kept_values));
     object.insert("type".to_string(), replacement);
 }
 
--- a/packages/browseros-agent/crates/browseros-mcp/src/framework.rs
+++ b/packages/browseros-agent/crates/browseros-mcp/src/framework.rs
@@ -656,3 +656,61 @@ pub fn page_json(page: &browseros_core::pages::PageInfo) -> Value {
     }
     value
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn normalized(schema: Value) -> Value {
+        let Value::Object(object) = schema else {
+            panic!("test schema should be an object");
+        };
+        Value::Object((*normalize_schema_object(object)).clone())
+    }
+
+    #[test]
+    fn nullable_enum_normalization_collapses_schemars_option_enum_shape() {
+        let schema = normalized(json!({
+            "type": "object",
+            "properties": {
+                "button": {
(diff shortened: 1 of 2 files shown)


DaniAkash:
@greptileai

avinashgola:
Pushed 62fc00f9b to retrigger CI after narrowing the nullable enum catalog guard to only reject null entries. Local verification is green: `bun run --filter @browseros/browseros-mcp-rust test:rust` and `bun run --filter @browseros/browseros-mcp-rust lint:rust` both pass.\n\nThe new Tests and Code Quality workflow runs are currently marked `action_required` with no jobs, so they need maintainer approval to run on this fork commit.

DaniAkash:
@avinashgola thanks for the atomic enum/type rewrite and the unit tests — I kept both, but reverted the narrowing that only stripped `null` when `type` was the `["…","null"]` array: a strict Pydantic client rejects a `null` enum entry regardless of what `type` says, so dropping `null` from any enum (matching the catalog guard) is the safer contract.

Issue #2408, opened 2026-08-21
Windows 10: BrowserOS neo fails to start — stub exits with code 3 unless launched with --no-sandbox


2217173240:
### Issue Type

Browser Issue

### Operating System

Windows 10 Pro for Workstations 22H2 (build 19045), physical machine (Intel LAPKC71E)

### Description of the bug

BrowserOS neo fails to start on Windows 10 — the browser process exits immediately without a window. The stub (`chrome.exe`) exits with **code 3** even in minimal mode (`chrome.exe --version` returns exit code 3 with no output). The only workaround is launching with `--no-sandbox`, after which everything works: window opens, CDP (port 9110) and the MCP endpoint (port 9010) all function normally.

Two failure signatures observed without `--no-sandbox`:

1. **Clean early exit with code 3** (most launches): the stub exits within ~1 second, no crash event is written to the Windows Application event log. This happens before any profile/argument handling — `--version` alone also fails with code 3.
2. **Crash with 0xC0000409 (STATUS_STACK_BUFFER_OVERRUN / fail-fast) in chrome.dll** (seen during first-run install): exception code `0xc0000409`, faulting module `chrome.dll` (148.0.7988.97), consistent fault offset `0x105054be` across two separate crash dumps — deterministic, not random memory corruption.

### Steps to Reproduce

1. Install BrowserOS neo on Windows 10 (build 19045)
2. Launch via the Start Menu shortcut (or `chrome.exe --remote-debugging-port=9110 --user-data-dir=...`)
3. Observe: process exits immediately, no window appears; event log may show `0xc0000409` in chrome.dll, or no event at all
4. Launch the same binary with `--no-sandbox` → browser starts and works correctly

### What I already tried (did not help)

- Closing the profile (no stale Singleton locks; `--version` fails identically, ruling out profile issues)
- Disabling Control Flow Guard for chrome.exe via `Set-ProcessMitigation -Name chrome.exe -Disable CFG` (as admin) — still exits with code 3
- Checked virtualization: physical machine, no VM

### Expected behavior

BrowserOS neo should start with the sandbox enabled (default launch path), without requiring `--no-sandbox`.

### Additional Context

- BrowserOS version: 0.49.5.0 (BrowserClaw server 0.0.44)
- Chromium version: 148.0.7988.97
- The `--no-sandbox` workaround works reliably but disables the sandbox entirely, which is not acceptable as a long-term solution for a browser used to access authenticated sites.
[...]

PR #2634, opened 2026-09-09
fix(build): disable Chromium test field trials in Windows releases

2217173240:
## Problem

Related to #2408. cc @shadowfax92

The Windows release preset omits `disable_fieldtrial_testing_config = true`, although the Windows debug, macOS release, and Linux release presets already set it. For non-Google-Chrome-branded builds, `is_official_build = true` does not prevent Chromium's field-trial testing configuration from being applied.

At the pinned Chromium base `8f5d36bc16f57115aeeff34baf4ad6aa964d509c`, that testing configuration enables `BrowserDynamicCodeDisabled` on Windows. The browser enables ACG before TSF activation. On the affected Windows 10 machine, Sogou's in-process initialization uses Xbyak, throws `Xbyak::Error` 14 (`ERR_CANT_PROTECT`), and calls `ExitProcess(3)`. The live policy at the exception was `BlockDynamicCode=ON`, `AllowThreadsToOptOut=ON`. This explains why reinstalling did not resolve the observed failure and why `--no-sandbox` was an unnecessarily broad workaround.

## Change

Disable test-only field trials in Windows release builds, matching the other release presets. Add a regression test that runs release presets through the GN-argument generation path and checks this invariant in the generated `args.gn`. Explicit custom GN overrides remain supported.

This restores production feature defaults; it does not disable Chromium's sandbox or override enterprise dynamic-code policy. A separate experimental TSF thread-opt-out patch is not included.

## Validation

- The new test failed for Windows before the preset change; all 13 `bos_build.steps.setup.configure_test` tests pass afterward.
- On the affected machine, official BrowserOS 0.50.3 launched and restarted successfully with the equivalent `--disable-field-trial-config` switch. Both `SogouTSF.ime` and `SogouPY.ime` were loaded, Sogou remained enabled in Windows, and no `--no-sandbox` switch was used.
- A separate native TSF/ACG probe reproduced exit 3 with the installed Sogou provider, supporting the diagnosis.
- `git diff --check` passes.

A full Chromium rebuild, a newly packaged neo binary, and Chinese composition/candidate-window regression tests have not been run. The runtime-switch result validates the configuration diagnosis, not a newly built installer. Please validate the next Windows build against #2408 before considering the issue fully resolved.

## CLA

### DIFF
--- a/packages/browseros/bos_build/config/gn/flags.windows.release.gn
+++ b/packages/browseros/bos_build/config/gn/flags.windows.release.gn
@@ -4,6 +4,8 @@ use_siso=false
 is_debug = false
 is_component_build = false
 is_official_build = true
+# Non-Chrome branded builds otherwise enable Chromium's test-only experiments.
+disable_fieldtrial_testing_config = true
 symbol_level = 0
 is_clang=true
 chrome_pgo_phase = 0
--- a/packages/browseros/bos_build/steps/setup/configure_test.py
+++ b/packages/browseros/bos_build/steps/setup/configure_test.py
@@ -53,6 +53,18 @@ class ConfigureValidateTest(unittest.TestCase):
 
 
 class ConfigureExecuteTest(unittest.TestCase):
+    def test_release_presets_exclude_chromium_testing_experiments(self):
+        presets = Path(configure.__file__).resolve().parents[2] / "config" / "gn"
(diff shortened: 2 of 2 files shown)


2217173240:

2217173240:

2217173240:
recheck

PR #2637, opened 2026-09-09
fix(server): bind reused provider credentials to the stored endpoint

DaniAkash:
## Problem

Testing a saved LLM provider without re-typing its (redacted) key backfills the key from the saved row. That backfill was gated on the provider type **and** the base URL, resource name, and region all matching the saved row exactly.

In practice the connection settings do not always match at test time: editing a provider can change the base URL, and the client can populate a default URL that differs from the stored one. When they diverge, the backfill was skipped and the test ran with a blank credential, so a provider that connects fine would fail its test.

## Fix

Drop the destination comparison. Keep the type match (an OpenRouter key is never reused as, say, an Azure provider), and when a blank credential is filled from the stored row, take the connection settings (base URL / resource name / region) from that same row. A reused secret can therefore only ever be sent to the endpoint it was saved against, never a caller-supplied URL. A caller that supplies its own credential keeps its own settings, since no stored secret is at risk.

This preserves the exfiltration guard (a saved key can't be paired with an arbitrary URL) while removing the fragile comparison that caused the false failures.

## Tests

`mergeStoredCredentials` unit tests updated to cover the new behavior:
- blank key filled from the stored row and pinned to the stored base URL
- a caller-supplied key keeps the caller's own settings
- a caller-supplied base URL with a blank key is pinned back to the stored URL (exfiltration guard)
- type change still blocks reuse
- Bedrock credentials filled and pinned to the stored region

`bun test` (7 pass), `tsc --noEmit`, and biome all clean.

### DIFF
--- a/packages/browseros-agent/apps/server/src/api/routes/provider.ts
+++ b/packages/browseros-agent/apps/server/src/api/routes/provider.ts
@@ -37,11 +37,13 @@ type StoredProvider = {
  * Fills blank credential fields from the saved provider row so testing an
  * existing provider works even though reads redact its secrets.
  *
- * The stored secret is bound to the stored destination: it is reused only when
- * the request targets the same provider type AND the same connection settings
- * (base URL / resource name / region). Otherwise a caller who knows a saved
- * provider id could pair its key with an arbitrary base URL and exfiltrate it
- * to another endpoint. A changed destination must supply its own credential.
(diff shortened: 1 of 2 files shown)


DaniAkash:
Superseded by #2639, which contains this same credential-merge fix plus the trusted-app-origin guard for /test-provider and /refine-prompt. Continuing there.

PR #2667, opened 2026-09-11
fix(app): restore local-only conversation from the local server

### LINKED ISSUE #2665: Chat history: clicking a past session flashes and reverts to previous view (regression in 0.50.3)
## Issue Type

Bug

## Operating System

Windows 11 (25H2, build 26200)

## BrowserOS Version

BrowserOS 0.50.3.0 (Chromium 151.0.8160.137, 64-bit)

## Description of the bug

Starting with the 0.50.3 update installed yesterday, past chat sessions can no longer be opened from the agent UI. Clicking a history entry causes a brief flash/hang and then the view snaps back to whatever was showing before the click - the selected conversation never loads. This is a regression: on the previous version, clicking a history entry opened it normally.

Details:

- The history list itself opens fine and shows all past sessions.
- Clicking a specific history entry does nothing useful - it flashes once and reverts to the prior view.
- On the home / new-tab page, clicking any history entry flashes but keeps you on the home page - it never navigates into the session.
- All past sessions are affected, not just some.
- Everything else in the agent UI works normally; only loading a historical session is broken.

## Steps to Reproduce

1. Open BrowserOS 0.50.3.0 and sign in.
2. Open the agent UI / assistant panel and open the chat history list.
3. Click on any past session entry.
4. Observe: it flashes briefly and reverts to the previous view instead of opening the session.
5. (Also reproducible from the home page - clicking any history entry flashes and stays on the home page.)

## Expected Behavior

Clicking a past session in the history list should open and display that conversation, as it did before 0.50.3.

## Additional Context

- Regression introduced by the 0.50.3 update (installed yesterday); the previous version was working.
- Possibly related to the chat-history durability changes shipped for #559 / PR #2044, since the symptom appeared right around that update.

L4XB:
Closes #2665.

## What was wrong

Since #2542 the local server owns chat history for signed-in users too, and the side panel history always lists those local conversations first. Opening one still went through the cloud query only, the path from before #2542: for a conversation without a cloud record the query resolved to nothing, the `conversationId` param was cleared and the panel snapped back to its previous view. Cloud-backed conversations kept working, which is exactly the split reported in #2665 (local-only fails, account-backed opens).

#2625 added the local restore for the new tab and for logged-out users (`restoreLocally`), but the signed-in side panel never reached it.

## Change

In the side panel restore effect, when the cloud has no record of the requested conversation, fall through to the same local server restore that logged-out and new-tab users already use (`fetchServerConversation` via `restoreServerConversation`). A conversation that does exist in the cloud is restored from there as before, so nothing changes for account-backed history. On a real miss the local path settles the UI the same way the cloud miss did (param cleared, no error banner in the side panel).

## Validation

- `bun run scripts/run-bun-test.ts --cwd=apps/app ./apps/app`: 433 pass, 0 fail across 61 files.
- `bunx wxt prepare && tsc --noEmit` in `apps/app`: clean (GraphQL types generated from `schema/schema.graphql`).
- `bunx @biomejs/biome check` on the changed file: clean.

Not done here: a browser check against a running local server. The change is confined to the branch that already runs for logged-out users, so the manual check that would help most is a signed-in user opening a local-only conversation from the side panel history.

### DIFF
--- a/packages/browseros-agent/apps/app/modules/chat/chat-session.hooks.ts
+++ b/packages/browseros-agent/apps/app/modules/chat/chat-session.hooks.ts
@@ -692,10 +692,14 @@ export const useChatSession = (options?: ChatSessionOptions) => {
           conversationIdParam as ReturnType<typeof crypto.randomUUID>,
         )
         setMessages(restoredMessages)
+        setRestoredConversationId(conversationIdParam)
+        setSearchParams({}, { replace: true })
+        return
       }
-      setRestoredConversationId(conversationIdParam)
-      setSearchParams({}, { replace: true })
-      return
+      // Not in the cloud. Since #2542 the local server owns a signed-in user's
+      // history too, so a conversation opened from the local history list has
+      // no cloud record: read it from the server instead of giving up, which
+      // left the side panel snapping back to its previous view (#2665).
     }
 
     if (isLoadingProviders || isLoadingAgentUrl) return


L4XB:

DaniAkash:
@greptileai

L4XB:
Thanks for the re-run. Greptile came back 4/5 with one substantive item, and I think it is right: there is no focused regression test for the path this PR actually adds — signed-in, cloud has no record, fall through to the local server restore.

That gap is real rather than cosmetic. `chat-session.hooks.ts` has no test file of its own; the nearest coverage is `chat-session-restore.test.ts`, which exercises `chat-session-restore.ts` rather than the side-panel effect that chooses between cloud and local. So the branch I changed is the one part of the flow nothing asserts, and a future refactor could drop the fallback without turning anything red.

I'd like to close it. What I'd add is a cell that drives the effect with a signed-in session and a conversation the cloud query resolves to nothing for, and asserts the local restore is called and the `conversationId` param is cleared once — plus the control, that a cloud-backed conversation still restores from the cloud and never reaches the local path.

Two questions before I write it:

1. Do you want it in this PR, or would you rather merge the fix and take the test separately? Greptile called it non-blocking and I don't want to hold the fix behind it if the regression is biting people.
2. Where would you prefer it — a new `chat-session.hooks.test.ts`, or folded into `chat-session-restore.test.ts` next to the existing restore coverage?

On the current state: `bun run scripts/run-bun-test.ts --cwd=apps/app ./apps/app` is 433 pass / 0 fail at `cceb386`, and `tsc --noEmit` and Biome are clean on the changed file. The one thing still not done is a browser check against a running local server — a signed-in user opening a local-only conversation from side-panel history. If that is something you can hit quickly, it is the check that would most directly confirm #2665 is gone.

DaniAkash:
@L4XB we can merge this PR, i'll handle the test coverage separately on a new PR

PR #2675, opened 2026-09-14
fix(server): launch newest sidecar and retain deferred updates

shadowfax92:
## summary
- launch the newest usable local sidecar before checking the feed. a newer downloaded copy starts directly; a newer or equal browser bundle wins.
- keep prepared updates across busy deferrals and browser restarts. retry activation independently of downloading, including when the feed is offline.
- confirm startup health before publishing the running version. reject failed downloaded candidates, reap their processes, and fall back to the bundle.

## design
this replaces the startup workaround in #2652. the old code launched the bundle before loading the downloaded version, then treated "already downloaded" as "already running". the workaround could still strand an update when the server was busy.

the version store owns bundled and ready installations plus their persisted selection. the updater downloads and prepares candidates. the manager owns launch selection, health-confirmed running identity, and deferred activation. bounded `--version` probes run off the ui thread; they validate executable identity, while startup health confirms the launched instance. the store implementation uses an existing compiled source file, so no gn changes are needed.

```text
bundle + persisted ready copy
             |
     async version validation
             |
     newest usable local copy
             |
       launch -> healthy -> running
             |
       failed download -> reject -> bundle

download -> verify -> persist ready
                          |
                  running server busy?
                     /          \
                   yes           no
                    |             |
             retain + retry    activate
```

## test plan
- [x] required ch1 chrome build passes; rerun before opening this pr reports no work needed.
- [x] all 14 isolated runtime scenarios pass with the built browser: local version selection, invalid and timed-out candidates, startup failure fallback, browserclaw selection, responsive ui during probing, shutdown cancellation, busy-to-idle activation with an offline feed, and restart while busy.
- [x] all 7 extracted patches byte-match the verified chromium commit and apply to the recorded base; no skipped patches.
- [x] whitespace check passes.

chromium source tip: `a0cd2891a9c64785e48e181104c056b30c454812`. no unit or browser test targets were built or run. runtime checks used fixture sidecars and an existing downloaded folder, not a fresh signed network download. signature verification is unchanged. native windows/linux builds and runtime checks weren't run.


shadowfax92:
@greptileai review

shadowfax92:
@claude review

